PR #81, opened 2018-01-07
[WIP] New submitter for FireWorks

fabiencastan:
FireWorks is a free, open-source code for defining, managing, and executing workflows.
See https://materialsproject.github.io/fireworks

### DIFF
--- a/meshroom/submitters/fireworksSubmitter.py
+++ b/meshroom/submitters/fireworksSubmitter.py
@@ -1,0 +1,73 @@
+#!/usr/bin/env python
+# coding:utf-8
+
+import os
+
+import fireworks
+from meshroom.core.desc import Level
+from meshroom.core.submitter import BaseSubmitter
+
+
+class FireworksSubmitter(BaseSubmitter):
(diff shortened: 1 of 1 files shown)


marcfielding1:
I can jump on this also, I would need to understand how you go about splitting our the jobs in meshroom to the different instances though?

marcfielding1:
Oh actually from reading it, it looks like this is most of the way there, it's just firing off compute requests....

marcfielding1:
So I emailed AliceVision also on this, do the workers just need a shared volume with the meshroom directory in it? So they get a message and all write to the same cache?

fabiencastan:
@marcfielding1 Sorry I missed your message.
Yes, it assumes that you have access to a shared storage.

nagilum99:
I'm poking the corpse: Why's this PR still open?
The autochecks are too old to see the problems via "Details".

I guess it's a "download the file and it may work" solution for now?

fabiencastan:
It's a sample, so people can start from there if they need to implement a new submitter.

PR #120, opened 2018-05-20
Install doc and workarounds

fabiencastan:


meshroom/ui/qml/main.qml
```diff
@@ -206,9 +206,9 @@ ApplicationWindow {
             dialog.text = args[1]
             dialog.detailedText = args[2]
         }
-        onInfo: createDialog(dialogsFactory.info, arguments)
-        onWarning: createDialog(dialogsFactory.warning, arguments)
-        onError: createDialog(dialogsFactory.error, arguments)
+        // onInfo: createDialog(dialogsFactory.info, arguments)
+        // onWarning: createDialog(dialogsFactory.warning, arguments)
+        // onError: createDialog(dialogsFactory.error, arguments)
```
fabiencastan: @yann-lty: I don't know how to do things differently to avoid the Qt 5.11 crash, so I just disabled it currently.

yann-lty: It seems to be a recent PySide bug when connecting signals with 2+ arguments. A temporary fix consists in wrapping those arguments in a simple structure with Properties.

PR #154, opened 2018-07-21
[core] multiple trials to replace the status file to avoid file conflicts on windows

fabiencastan:
On Windows, attempting to remove a file that is in use causes an exception to be raised. So we may need multiple trials.
I didn't get the problem anymore with this fix.


meshroom/core/node.py
```diff
@@ -197,6 +197,15 @@ def saveStatusFile(self):
         statusFilepathWriting = statusFilepath + '.writing.' + str(uuid.uuid4())
         with open(statusFilepathWriting, 'w') as jsonFile:
             json.dump(data, jsonFile, indent=4)
+        for i in range(20):
+            try:
+                os.remove(statusFilepath)
+                # if remove is successful, we can stop the iterations
+                break
+            except:
```
yann-lty: except WindowsError:

PR #157, opened 2018-07-27
Save nodes position + undoable nodes moves

yann-lty:
This PR adds nodes move operations to the undo stack and save nodes positions in the project file. This allows to restore nodes positions when loading a project or whenever an operation recreates a deleted node on "undo".
It also improves GraphEditor UX regarding node selection, visual feedbacks and attributes connections.

- [x] [core] introduce position property on Nodes
- [x] [core][io] add "position" field to serialized node data
- [x] [core] expose available features given a Meshroom file version
- [x] [ui] add commands to include Nodes movements in the undo stack
- [x] [ui] GraphLayout: move graph layout management to Python
- [x] [ui] use GraphLayout from Python after nodes duplication/sfm augmentation
- [x] [ui] GraphEditor UI improvements

### DIFF
--- a/meshroom/core/graph.py
+++ b/meshroom/core/graph.py
@@ -13,6 +13,7 @@ from enum import Enum
 import meshroom
 import meshroom.core
 from meshroom.common import BaseObject, DictModel, Slot, Signal, Property
+from meshroom.core import Version
 from meshroom.core.attribute import Attribute, ListAttribute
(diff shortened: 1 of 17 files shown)


fabiencastan:
Error when loading a previous scene:
```
Error while loading monstrock.mg - Meshroom 2018.1

An unexpected error has occurred
'unicode' object has no attribute 'name'
```

PR #163, opened 2018-08-03
Setup standalone release

yann-lty:
This PR contains features, fixes and adjustments for the upcoming prebuilt release of Meshroom. 
Related to #128.

### Setup
- [X] initialize executables environment to work with prebuilt package structure (see `setupEnvironment` in `meshroom.__init__`)
- [X] use official PySide2 release from PyPI (excludes Python 2.7 on Windows)
- [X] add dependencies in COPYING file
- [X] make meshroom_photogrammetry an Executable for cx_Freeze
- [X] use Meshroom logo .ico file as application icon on Windows

### UI
- [X] AboutDialog with Open Source Licences from COPYING files info and useful links
- [X] main window footer with cache folder info
- [X] prompt warning when computing an unsaved project
- [X] store main window size in persistent settings

PR #166, opened 2018-08-06
[setup] add PlatformExecutable to handle variations between platforms + dependencies lookup

yann-lty:
Introduce `PlatformExecutable` class to manage variations between platforms for executable freezing via cx_Freeze. This PR also fixes the issue reported (meshroom_photogrammetry without targetName) after #163 was merged.
After trying packaged application on different Linux distributions, it turns out that we also need to pack some system libraries (e.g: ssl) for things to work. We took this [setup script](https://github.com/Ultimaker/cura-build/blob/master/packaging/setup_linux.py.in) as reference to tweak dependencies lookup.


yann-lty:
@bartoszek could you confirm this solves the error you had after we merged #163? Also, I would be interested if you have any feedback regarding what has changed here with dependencies lookup.

bartoszek:
@yann-lty build fine on my end.

PR #167, opened 2018-08-06
GraphLayout: fix bouding box computation

yann-lty:
Fix incorrectly initialized width and height for graph bounding box.


jonbragado:
Once the control points are calculated from the images, not everything in there needs to be reconstructed as a mesh. 
The user should be able to choose a bounding box size to compute or even with lasso type selection tools in the viewport, remove points to prevent calculation. 

A very integral feature currently shown in photogrammetry software.

PR #184, opened 2018-08-11
Travis CI: Add flake8 to the testing

cclauss:
Add http://flake8.pycqa.org tests to look for Python syntax errors and undefined names

E901,E999,F821,F822,F823 are the "showstopper" flake8 issues that can halt the runtime with a SyntaxError, NameError, etc. Most [other flake8 issues](https://travis-ci.org/alicevision/meshroom/jobs/414794287#L479) are merely "style violations" -- useful for readability but they do not effect runtime safety.

    F821: undefined name name
    F822: undefined name name in __all__
    F823: local variable name referenced before assignment
    E901: SyntaxError or IndentationError
    E999: SyntaxError -- failed to compile a file into an Abstract Syntax Tree

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -5,6 +5,13 @@ python:
 
 install:
   - "pip install -r requirements.txt -r dev_requirements.txt --timeout 45"
+  - pip install flake8
+
+before_script:
+  # stop the build if there are Python syntax errors or undefined names
+  - flake8 . --count --select=E901,E999,F821,F822,F823 --show-source --statistics
+  # exit-zero treats all errors as warnings.  The GitHub editor is 127 chars wide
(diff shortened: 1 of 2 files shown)


fabiencastan:
Thanks for the contribution! That's a good idea to have flake8 checks.
It would be great to also launch it to appveyor.

cclauss:
To avoid duplication, we do not need identical tests in both places.  We should pick one or the other.  I am happy either way.

fabiencastan:
My intention was to keep things symmetric but yes, I agree with your argument.
And travis is more stable than appveyor, so it's better to have it in travis as you have done.

yann-lty:
Thanks for your contribution!

PR #213, opened 2018-08-27
Bugfix for using MESHROOM_SUBMITTERS_PATH env variable.

mike1158:
Setting MESHROOM_SUBMITTERS_PATH caused meshroom to try to load the bundled simpleFarmSubmitter plugin twice, ignoring the path in the variable. The problem lies with the way loadPlugins() works in meshroom.core--it's trying to import the package name "submitters" more than once. The second time, import_module() just returns a reference to the same "submitters" package that's part of the meshroom distribution.

This patch just makes MESHROOM_SUBMITTERS_PATH a replacement for the bundled submitters instead.  It's not the best fix, but it looks like a more proper fix might affect other, non-submitter, plugins as well, so I tread lightly.

### DIFF
--- a/.github/stale.yml
+++ b/.github/stale.yml
@@ -1,0 +1,17 @@
+# Number of days of inactivity before an issue becomes stale
+daysUntilStale: 120
+# Number of days of inactivity before a stale issue is closed
+daysUntilClose: 7
+# Issues with these labels will never be considered stale
+exemptLabels: false
+# Label to use when marking an issue as stale
+staleLabel: stale
+# Comment to post when marking an issue as stale. Set to `false` to disable
+markComment: >
+  This issue has been automatically marked as stale because it has not had
+  recent activity. It will be closed if no further activity occurs. Thank you
+  for your contributions.
+# Comment to post when closing a stale issue. Set to `false` to disable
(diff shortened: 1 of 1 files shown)


meshroom/core/__init__.py
```diff
@@ -255,7 +255,9 @@ def loadSubmitters(folder, packageName):
 subs = loadSubmitters(meshroomFolder, 'submitters')
 # -  additional 3rd party submitters
 if "MESHROOM_SUBMITTERS_PATH" in os.environ:
-    subs += loadSubmitters(os.environ["MESHROOM_SUBMITTERS_PATH"], 'submitters')
+    subs = loadSubmitters(os.environ["MESHROOM_SUBMITTERS_PATH"], 'submitters')
+else:
+    subs = loadSubmitters(meshroomFolder, 'submitters')
```
fabiencastan: In that case, it could be:
`subs = loadSubmitters(os.environ.get("MESHROOM_SUBMITTERS_PATH", meshroomFolder), 'submitters')`

mike1158: Agreed, that is a bit nicer.  Shall I amend the PR?

PR #248, opened 2018-09-16
More intuitive add node menu

ChemicalXandco:
At the moment, when you add a node this menu appears: 

It:

- is not very helpful
- doesn't make use of space efficiently, so takes up lots of vertical space and barely any horizontal
- when more nodes are added it will just take up more space, potentially filling the entire height of the window

This fixes it by adding categories which then list the nodes, this makes it easier to navigate and makes more sense.

Sometime in the future, when the documentation is resolved, there should be a way added to get information on each node from this menu.


ChemicalXandco:
@fabiencastan 
preview
Is it good enough to merge or need more work or close PR?

fabiencastan:
Hi @ChemicalXandco ,
Thanks for the PR. We don't have the time to look at it right now. But we will come back to you asap.

ChemicalXandco:
Ok I understand.

natowi:
@ChemicalXandco your new Add node menu, including the preview images and descriptions looks really nice. However I would propose to keep a variant of the old "Add node" menu: 
right click -> new add node menu 
left click -> display the search bar (only!) from the old "Add node" menu, but with node name auto-complete feature. So users familiar with the node-names can quickly add new nodes to the graph.

ChemicalXandco:
@natowi that might be useful but maybe I should just add a search bar, I myself don't feel the need to add like nodes very quickly, as the processing time far outweighs it any time spent sorting out the graph. I think it is better to use this all the time, as it helps new users a lot. Let me know what you think is best, but I think it is unnecessary to have 2 at the same time.

fabiencastan:
Hi @ChemicalXandco,
Really sorry for the huge delay to answer and thanks for the contribution.

We agree that the current menu is too dense and difficult to read. On the other hand, the current proposition doesn't look like a right click menu and will not scale when we will continue to add new nodes.
The main usage of the graph editor is for advanced users, who want to navigate quickly.

So a first step would be to:
 - add sub-menus with categories, as in many nodal softwares (e.g. "Sparse Reconstruction", "Dense Reconstruction", "Utils", "Mesh Post-Processing")
 - add an "Help" tab on each node

Then, to make it easier for new users (as you suggested the need), we could consider an help entry in the menus which provides node description with help, with a way to create a new node from there. In that case, a modal window would make sense.

What do you think?

Would you be interested in contributing to these changes? If yes, we could discuss about it in more technical aspects. For instance, we try to limit as much as possible the knowledge of the nodes in the QML part. So if we add documentation on each node in should be on the node description on the Python side.

Best,
Fabien

ChemicalXandco:
@fabiencastan 
I am interested in contributing to these changes as in its current form, I agree that it does not work as a right-click menu. Can I just clarify some things:

- The right click menu made to be like the one in blender with sub-menus
untitled3576457
- all information about nodes moved to python side

Now what I don't quite understand is where you suggest the extended dialog should be accessed from, but I think there are possible places:

- right click menu: right click on the name of the node and another menu appears and has 'Add' and 'Info' with info opening the extended dialog and selecting it to get the description
- From the 'Help' menu at the top
- From node itself as in the right click menu on it
- All of above

I hope that made sense, is it good for me to go ahead with this?

ChemicalXandco:
@fabiencastan 
I still need to squash a bug or two, but it should be mostly done now: 
gif-1-61d9348df2ec
The info for each node needs to be done, but I don't think I know enough to write them so maybe a dev should do it? If you have any more feedback, I am happy to receive it.

ChemicalXandco:
Hello, this should be good to merge, is there anything I need to do?

meshroom/nodes/aliceVision/CameraCalibration.py
```diff
@@ -6,6 +6,9 @@
 class CameraCalibration(desc.CommandLineNode):
     commandLine = 'aliceVision_cameraCalibration {allParams}'
 
+    category = 3
```
fabiencastan: Use string keyword for category (instead of arbitrary number), like "Utils".

fabiencastan:
The graph should not have any knowledge of the nodes content, so the menu in the GraphEditor should be dynamically built based on the category names.

fabiencastan:
We will not merge an empty info widget. We don't merge intermediate steps into develop. The develop branch should be able to become a release if needed.
It will take a long time before we have full presentation for each node. To ease the merge, it would be better to split the problems into multiple PRs. So I would highly recommend to make a first PR only with the categories.

natowi:
@ChemicalXandco nice work. I can help you out with the missing node descriptions.
Have a look at my [WIP documentation](https://docs.google.com/document/d/17HYtYS1tvx053k3_nO6Z2GnP2R3cXMlGMN-1WIe3kJE). In the "complete node list" you can find descriptions and references for the missing nodes.

natowi:
new node https://github.com/alicevision/meshroom/pull/639

ChemicalXandco:
Would be good to get this merged for the next release

natowi:
@fabiencastan This would be really useful for the next release, as we have so many new nodes that do not fit all on screen with the old add node menu.

natowi:
@ChemicalXandco fyi, we now have node documentation  https://github.com/alicevision/meshroom/pull/808/commits/0e606eef4e03f00efa7a0abcbffa73a9d6f88c8c and https://github.com/alicevision/meshroom/pull/808/commits/64a4c9426ad204eaf261758ab0c1cf4fed886886

natowi:
@fabiencastan with the increasing number of nodes it would be great to get this PR included in the next release...

fabiencastan:
The PR needs to be updated. I have not tested it yet.

ChemicalXandco:
I can update it if we need this feature

fabiencastan:
Yes, it would be great to have categories in this menu, but we need to find a way to keep the search bar.

ChemicalXandco:
> we need to find a way to keep the search bar

Last I checked this doesn't remove the search bar.

I shall make a new branch based on upstream develop because this one is too old.

fabiencastan:
Yes, I don't remember exactly how this version was working... :s
Thanks for doing it again!

PR #321, opened 2018-12-16
Typo Side->Size

natowi:
name='textureSide',
label='Texture Side',
description='''Output texture size''',


fabiencastan:
If we change the name here, we have to do the same in AliceVision:
https://github.com/alicevision/AliceVision/blob/develop/src/software/pipeline/main_texturing.cpp#L100

natowi:
I see. This typo can also be found in src/aliceVision/mesh/UVAtlas.cpp src/aliceVision/mesh/UVAtlas.hpp src/software/pipeline/main_texturing.cpp src/aliceVision/mesh/Texturing.hpp

PR #329, opened 2018-12-19
[CLI] Add possibility to run pipeline from a .sfm file

simogasp:
* Allows to run the pipeline from a .sfm file, optionally containing the camera intrinsics
* exposes the refineIntrinsics parameter in sfm node to lock the camera intrinsics (eg the cameras have been already calibrated

PR #330, opened 2018-12-20
New option to SfMTransform node + new options to bin/meshroom_photogrammetry

FabienDanieau:
--- SfMTransform
The option "from_single_camera" is available in the transformation method. It allows to use a view as the origin of the coordinate system. The name of the view as to be specified in the transformation text field.

--- meshroom_photogrammetry
The new option --graph allows to use a ".mg" file to specified a pipe previously configured for the reconstruction.


yann-lty:
Thanks for this PR! 
Would it be possible to split the 2 features in separate PRs ?
Modifications in `SfMTransform` can be merged easily but we have the pending PR #329 that might conflict with the changes in meshroom_photogrammetry.

FabienDanieau:
Done. It has been split in PR #353 and #354 
This PR can be closed.

PR #354, opened 2019-01-14
New options to bin/meshroom_photogrammetry

FabienDanieau:
The new option --graph allows to use a ".mg" file to specified a pipe previously configured for the reconstruction.


yann-lty:
Turns out some internal refactoring was actually needed to make everything work with #329 and initialize graphs properly in all cases. 
Could you take a look at #377 (`--graph` renamed to `--pipeline` to be more explicit) and check if it still correctly fits your needs?
Thanks!

Issue #311, opened 2018-12-10
Broken UI on 4k display (possible an issue with windows screen scaling)


Uber616:
### Problem

UI Is broken on 4k screen. Labels do not fit in their boxes. Things are misaligned. 

**Expected behavior:**  
I expected the UI to be readable on any screen resolution.

**Actual behavior:**  
A mess appeared before my eyes. 

### Steps to Reproduce

1. Get a computer with 4k screen.
2. Launch the meshroom.
3. It is mostly visible on the graph editor.

### Versions

* AliceVision branch/version: Meshroom 2018.1.0 (2018/08/09)
* OS: Win 10

fabiencastan:
Could you provide a screenshot to better understand what are the problems?

Uber616:
After some more investigation this might be more of a screen scaling issue then resolution problem. I have 200% scaling in windows because of the 4k display resolution. When I changed it to 150% it looks better. But still minor labeling issues. Below screenshot from 200%.

![screenshot_1](https://user-images.githubusercontent.com/33427330/49744989-4ce66500-fc9e-11e8-8faa-4f8446155af2.png)

ftc:
I am having a similar issue.  The UI appears to not be scaling correctly on my 4k Dell Xps 9550 running Windows 10. The biggest issue is that clicking directly on the buttons does not work, I can click below the start button to hit it.  A screenshot is shown below.

![meshroom_ui_scaling_bug](https://user-images.githubusercontent.com/762317/50500612-cc4e9500-0a0f-11e9-8e8a-6c16c506f432.PNG)

Unfortunately the workaround of changing the DPI or dropping my resolution to 1080p does not seem to help.

Thanks!

PR #378, opened 2019-01-29
Minor UI fixes

yann-lty:
- [X] filter non silenced, inoffensive warnings from QML + log Qt messages via Python logging
- [X] make sure Meshroom files are saved with `.mg` extension
- [X] use TextMetrics to handle AttributePins expanding instead of having 2 Labels
- [X] IntrinsicsIndicator: avoid warnings when trying to store `undefined` in string properties

### DIFF
--- a/meshroom/ui/app.py
+++ b/meshroom/ui/app.py
@@ -1,7 +1,7 @@
 import logging
 import os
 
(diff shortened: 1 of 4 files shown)


fabiencastan:
Should partially improve screen scaling issues: https://github.com/alicevision/meshroom/issues/311

PR #413, opened 2019-03-07
[CLI] Added new options

simogasp:
This PR introduces two main options to the command line scripts.
* `meshroom_photogrammetry`, a `--overrides` parameter is added that enables the user to specify a json file containing parameters values to use for each node, thus overriding the default settings. The json file follows a similar structure of the `.mg` file containing the pipeline, ie:
```json
{
        "<node1Name>": {
                "<param1Name>": "<param1Value>"
                "<param2Name>": "<param2Value>"
        },
        "<node2Name>": {
                 "<param1Name>": "<param1Value>"
        },
...
}
```
For example one file setting file `settings.json` could be
```json
{
        "FeatureExtraction_1": {
                "describerPreset": "high"
        },
        "FeatureMatching_1": {
                "guidedMatching": "0"
        },
        "StructureFromMotion_1": {
                "useLocalBA": "0",
                "refineIntrinsics": "0"
        }
}
```

* `ui` can now take a parameter ~~`--pipeline`~~ `--project` that enables the user to specify a project file to load, i.e.
```bash
python meshroom/ui --project myProject.mg
```

Moreover, it add the possibility for the node `PrepareDenseScene` to export the list of undistorted images so they can, eg, published with the `Publish` node. This is because there was no other way of export the images since the suppression of the node `ExportUndistortedImages` (490ebfe63089026495bcc8cecd9834f023e686b1)

### DIFF
--- a/meshroom/ui/app.py
+++ b/meshroom/ui/app.py
@@ -1,7 +1,8 @@
 import logging
 import os
+import argparse
 
-from PySide2.QtCore import Qt, Slot, QJsonValue, Property, qInstallMessageHandler, QtMsgType
+from PySide2.QtCore import Qt, QUrl, Slot, QJsonValue, Property, qInstallMessageHandler, QtMsgType
 from PySide2.QtGui import QIcon
 from PySide2.QtWidgets import QApplication
 
@@ -53,8 +54,8 @@ class MessageHandler(object):
 class MeshroomApp(QApplication):
     """ Meshroom UI Application. """
     def __init__(self, args):
-        args = [args[0], '-style', 'fusion'] + args[1:]  # force Fusion style by default
-        super(MeshroomApp, self).__init__(args)
+        QtArgs = [args[0], '-style', 'fusion'] + args[1:]  # force Fusion style by default
+        super(MeshroomApp, self).__init__(QtArgs)
 
         self.setOrganizationName('AliceVision')
         self.setApplicationName('Meshroom')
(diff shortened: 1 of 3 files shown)


meshroom/ui/app.py
```diff
@@ -38,7 +39,7 @@ class MessageHandler(object):
     qmlWarningsBlacklist = (
         'Failed to download scene at QUrl("")',
         'QVariant(Invalid) Please check your QParameters',
-        'Texture will be invalid for this frame',
+        'Texture wiLuisa Cornall be invalid for this frame',
```
yann-lty: typo: will

meshroom/ui/app.py
```diff
@@ -95,6 +96,13 @@ def __init__(self, args):
         # Request any potential computation to stop on exit
         self.aboutToQuit.connect(r.stopExecution)
 
+        parser = argparse.ArgumentParser(prog=args[0], description='Launch Meshroom UI.')
+        parser.add_argument('--pipeline', metavar='MESHROOM_FILE', type=str, required=False,
```
yann-lty: Maybe `project` would be more relevant here.

meshroom/ui/app.py
simogasp: oopsi... a wild cut&paste!

meshroom/ui/app.py
simogasp: ok, i just thought to be consistent with the other binary and effectively project is more self-explanatory.

yann-lty: Yes I understand :) In the other binary the `pipeline` argument is to load a custom template pipeline (uninitialized project with custom parameters/nodes) and run it on input images, while in this case I would say it is mainly to load an initiliazed project file.

PR #445, opened 2019-04-18
Avoid issue with use ==/!= to compare str, bytes, and int literals

cclauss:
[flake8](http://flake8.pycqa.org) testing of https://github.com/alicevision/meshroom on Python 3.7.1

$ __flake8 . --count --select=E9,F63,F72,F82 --show-source --statistics__
```
./meshroom/core/node.py:461:34: F632 use ==/!= to compare str, bytes, and int literals
            if v is not None and v is not '':
                                 ^
./meshroom/core/node.py:480:34: F632 use ==/!= to compare str, bytes, and int literals
            if v is not None and v is not '':
                                 ^
2     F632 use ==/!= to compare str, bytes, and int literals
2
```

### DIFF
--- a/meshroom/core/node.py
+++ b/meshroom/core/node.py
@@ -458,7 +458,7 @@ class BaseNode(BaseObject):
             self._cmdVars[name] = '--{name} {value}'.format(name=name, value=v)
             self._cmdVars[name + 'Value'] = str(v)
 
-            if v is not None and v is not '':
+            if v not in (None, ''):
                 self._cmdVars[attr.attributeDesc.group] = self._cmdVars.get(attr.attributeDesc.group, '') + \
                                                           ' ' + self._cmdVars[name]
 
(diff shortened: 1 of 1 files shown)


meshroom/core/node.py
```diff
@@ -458,7 +458,7 @@ def _buildCmdVars(self):
             self._cmdVars[name] = '--{name} {value}'.format(name=name, value=v)
             self._cmdVars[name + 'Value'] = str(v)
 
-            if v is not None and v is not '':
+            if v not in (None, ''):
```
fabiencastan: Wouldn't it be better to just do: `if v:`?

yann-lty: Indeed, `v` is a string value at this point. Only testing `if v:` is more relevant here.
By the way, it would be also nice to add Python 3.7 variant in appveyor.yml and .travis.yml.

PR #518, opened 2019-06-18
Update Dockerfile

sky3:
The location of the reference docker image seems wrong.


Dockerfile
```diff
@@ -1,6 +1,6 @@
 ARG CUDA_TAG=7.0
 ARG OS_TAG=7
-FROM alicevision:centos${OS_TAG}-cuda${CUDA_TAG}
```
fabiencastan: It should be: `alicevision/alicevision:centos${OS_TAG}-cuda${CUDA_TAG}`

See https://hub.docker.com/r/alicevision/alicevision

fabiencastan:
It has been fixed in another PR.

PR #531, opened 2019-07-01
[PrepareDenseScene] option to correct images exposure value

ClaraVere:
Advanced parameter : boolean to choose whether or not to correct images exposure values (EV) during the PrepareDenseScene.


meshroom/nodes/aliceVision/PrepareDenseScene.py
```diff
@@ -55,6 +55,14 @@ class PrepareDenseScene(desc.CommandLineNode):
             uid=[0],
             advanced=True
         ),
+        desc.BoolParam(
```
fabiencastan: Need to update node __version__

PR #537, opened 2019-07-03
Fixing meshroom_photogrammetry --input cameras.sfm behaviour

pocketpixels:
When meshroom_photogrammetry is called with the --input parameter pointing to a SFM file, then the camera intrinsics and poses are loaded from that file. In this case we should not call CameraInit.buildIntrinsics()

### DIFF
--- a/bin/meshroom_photogrammetry
+++ b/bin/meshroom_photogrammetry
@@ -101,7 +101,8 @@ else:
     graph = multiview.photogrammetry(inputViewpoints=views, inputIntrinsics=intrinsics, output=args.output)
     cameraInit = getOnlyNodeOfType(graph, 'CameraInit')
 
-views, intrinsics = cameraInit.nodeDesc.buildIntrinsics(cameraInit, images)
+if not (views and intrinsics):
+    views, intrinsics = cameraInit.nodeDesc.buildIntrinsics(cameraInit, images)
 cameraInit.viewpoints.value = views
 cameraInit.intrinsics.value = intrinsics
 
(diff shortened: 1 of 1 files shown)


bin/meshroom_photogrammetry
```diff
@@ -101,7 +101,8 @@ else:
     graph = multiview.photogrammetry(inputViewpoints=views, inputIntrinsics=intrinsics, output=args.output)
     cameraInit = getOnlyNodeOfType(graph, 'CameraInit')
 
-views, intrinsics = cameraInit.nodeDesc.buildIntrinsics(cameraInit, images)
+if not (views and intrinsics):
```
yann-lty: Or rather:
```if images:```
We still want to go through buildIntrinsics when we have "unresolved" input images (from args.inputImages), even when retrieving views and intrinsics from an pre-calibrated sfm file. 
Is this okay for your use-case ?

Comment (simogasp) [reply bin/meshroom_photogrammetry]: well, now that #567 has been merged my suggestion is
```suggestion
if images:
    views, intrinsics = cameraInit.nodeDesc.buildIntrinsics(cameraInit, images)
    cameraInit.viewpoints.value = views
    cameraInit.intrinsics.value = intrinsics
```
because:
* whether a default or custom pipeline (`--pipeline`) is passed at this point `cameraInit` already has the `views` and `intrinsics`
* if `image` is passed at this point we can read the rest of unresolved images (`buildIntrinsics`) and update `cameraInit` (`cameraInit.*.value = ...`)
* otherwise nothing else need to be done

pocketpixels: It looks like #567 should address the issue I had. 
The issue being that the poses and intrinsics loaded from the SFM file would be discarded during initialization. 
I do not know (and am unfortunately unable to test) whether the additional changes discussed above are still necessary.

pocketpixels: But the proposed change looks reasonable to me.

PR #539, opened 2019-07-04
Features Viewer

yann-lty:
## Description
Based on QtAliceVision plugin, provide a viewer for extracted feature points within the 2D Image Viewer.

features

* Provides 3 visualization modes: Points, (scaled) Squares, Orientated Squares
* FeatureExtraction nodes can be double clicked to drive which node is used for the FeaturesViewer
* Component is loaded dynamically to evaluate QtAliceVision plugin availability - an error is displayed if the plugin fails to load:
image
* A picker with a predefined palette is available to select feature points colors
featcolors

### DIFF
--- a/meshroom/ui/qml/Controls/ColorChart.qml
+++ b/meshroom/ui/qml/Controls/ColorChart.qml
@@ -1,0 +1,66 @@
+import QtQuick 2.10
+import QtQuick.Controls 2.10
+
+import Utils 1.0
+
+/**
+ * ColorChart is a color picker based on a set of predefined colors.
+ * It takes the form of a ToolButton that pops-up its palette when pressed.
+ */
+ToolButton {
+    id: root
+
+    property var colors: ["red", "green", "blue"]
+    property int currentIndex: 0
+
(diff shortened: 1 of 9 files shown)


ChemicalXandco:
Is meshroom still supporting Qt 5.11? This addition makes error when using Qt 5.11: `meshroom/ui/qml/Viewer/FeaturesInfoOverlay.qml:2 module "QtQuick.Controls" version 2.11 is not installed`
This is very good feature btw!

yann-lty:
Thanks for the feedback!
Yes 5.11 is still supported for now, you can lower the version of imported QtQuick Controls (I was working with 5.13 and did not paid enough attention to modules versions here).
But the goal is to move on to 5.13 really soon to be able to benefit from the improvement made to the JS engine (ES7) and some of the new controls (like the new in QuickControls 2.x).
Were you able to build the QtAliceVision C++ plugin without issues ?

ChemicalXandco:
Sounds good! I am trying to compile QtAliceVision at the moment, I will report back if it is working not.

PR #544, opened 2019-07-07
Button to select the chunk being processed

ChemicalXandco:
Minor addition which adds a button in the list of chunks in the node log to select the chunk that is being processed, very useful when scrolling through lots of chunks to find it.

The button only shows when the node is running or has error and has multiple chunks.
chunk_button


fabiencastan:
IMHO, having a new button just for that takes space and complicates the GUI for nothing.
It could be in a right click menu if needed (and associated to a shortcut).
I have not checked the code, but just to be sure: multiple chunks could have the "RUNNING" status at the same time on renderfarm.

fabiencastan:
@ChemicalXandco Could you adjust this PR to use the right click menu?

PR #558, opened 2019-07-23
[Texturing] set texturing downscale default value to 1

ClaraVere:


meshroom/nodes/aliceVision/Texturing.py
```diff
@@ -91,14 +91,14 @@ class Texturing(desc.CommandLineNode):
...
         ),
-        desc.BoolParam(
-            name='correctEV',
-            label='Correct Exposure',
-            description='Uniformize images exposure values.',
-            value=False,
-            uid=[0],
-            advanced=True,
-        ),
+        # desc.BoolParam(
```
fabiencastan: Should not be removed, isn't it?

PR #564, opened 2019-07-26
Visual interface for node statistics

lgeertsen:
Visual charts to get computer information during node execution.

- CPU usage
- RAM usage
- GPU usage, memory & temperature

Automatically updates charts while nodes are computing

image

Possibility to view CPU usage for every core and the average usage

image

Highlighting curves when hovering over the checkbox of a cpu for better visibility

image


natowi:
@lgeertsen Just an idea on this: An overheat protection for local computations would be useful (pause computation until temp is good).

PR #567, opened 2019-07-31
[bin] Fix project creation from custom pipeline

simogasp:
This PR fixes a problem when using a custom pipeline in meshroom_photogrammetry: the views and the intrinsics that might have been read before from the a json/sfm file were not added to the graph, thus leading to failure when calling cameraInit in AliceVision.
```
[11:25:25.780397][error] Program need -i or --imageFolder option
```

### DIFF
--- a/bin/meshroom_photogrammetry
+++ b/bin/meshroom_photogrammetry
@@ -80,6 +80,8 @@ elif os.path.isfile(args.input) and os.path.splitext(args.input)[-1] in ('.json'
     # args.input is a sfmData file: setup pre-calibrated views and intrinsics
     from meshroom.nodes.aliceVision.CameraInit import readSfMData
     views, intrinsics = readSfMData(args.input)
+else:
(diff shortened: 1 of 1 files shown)


bin/meshroom_photogrammetry
```diff
@@ -80,6 +80,8 @@ elif os.path.isfile(args.input) and os.path.splitext(args.input)[-1] in ('.json'
     # args.input is a sfmData file: setup pre-calibrated views and intrinsics
     from meshroom.nodes.aliceVision.CameraInit import readSfMData
     views, intrinsics = readSfMData(args.input)
+else:
```
fabiencastan: This is an error, as `args.inputImages` could have been used instead of `args.input`.

fabiencastan: New PR to fix it: https://github.com/alicevision/meshroom/pull/577

PR #578, opened 2019-08-08
Release 2019.2.0

fabiencastan:
AliceVision 2.2.0 release: https://github.com/alicevision/AliceVision/pull/670

### DIFF
--- a/requirements.txt
+++ b/requirements.txt
@@ -1,5 +1,5 @@
 # runtime
 psutil
 enum34;python_version<"3.4"
-PySide2==5.11.1
+PySide2==5.13.0
 markdown==2.6.11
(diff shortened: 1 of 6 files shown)


requirements.txt
```diff
@@ -1,5 +1,5 @@
 # runtime
 psutil
 enum34;python_version<"3.4"
-PySide2==5.11.1
+PySide2==5.13.0
```
simogasp: is it mandatory or 5.12 can work as well?

fabiencastan: Yes, 5.12 is compatible.
We use 5.13 here to match the version we will use to make the release binaries.

simogasp: ok, can we add a line in `INSTALL.md` with the minimum qt version supported and warning to have a matching version between qt and pyside?
E.g. here https://github.com/alicevision/meshroom/blob/develop/INSTALL.md#python-environment adding another bullet:
```markdown
*  Qt and PySide 5.X.Y.  To install the latest supported version of PySide, you can use `requirements.txt` as described below, or you can manually install the version that matches the Qt version installed in your system.
```

PR #590, opened 2019-08-13
Add Task Manager for computing nodes and ability to work on graph while computing

lgeertsen:
### Created a task manager that manages the computing of nodes.
Added a "Task Manager" tab in meshroom to have a clear overview of the submitted, running and pending nodes

image

When a node fails computing, the rest of the nodes of the branch will be taken out of the cue. And the task manager will start executing the next branch of nodes

image

### Meshroom isn't locked anymore while computing
Now you can still work in the meshroom while nodes are being computed
While computin you can now:
- Submit more nodes to the task manager
- Edit and remove nodes, if the node or one of the connected nodes hasn't been submitted to the task manager
- Duplicate nodes

### DIFF
--- a/meshroom/common/qt.py
+++ b/meshroom/common/qt.py
@@ -216,6 +216,7 @@ class QObjectListModel(QtCore.QAbstractListModel):
     def reset(self, objects):
         self.setObjectList(objects)
 
+    @QtCore.Slot(QtCore.QObject, result=bool)
     def contains(self, obj):
         """ Returns true if the list contains an occurrence of object;
         otherwise returns false.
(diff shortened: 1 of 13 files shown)


fabiencastan:
Would be good to have access to the node selection from the Task Manager.

fabiencastan:
When we submit on renderfarm and compute another branch locally and then stop the local computation, it removes the submitted status on all nodes including the nodes computing on renderfarm (with status:SUBMITTED execMode:EXTERN).

fabiencastan:
Included in https://github.com/alicevision/meshroom/pull/612.

PR #598, opened 2019-08-16
Added automatic stale detection and closing for issues

simogasp:
## Description

Add automatic detection and marking of stale issues. 
These will be automatically closed after 7 days.

see https://github.com/apps/stale


fabiencastan:
That's great!
I would start with more conservative values, like: `daysUntilStale: 120`.

PR #610, opened 2019-08-26
Bugfix for MESHROOM_SUBMITTERS_PATH env variable

mike1158:
Setting MESHROOM_SUBMITTERS_PATH caused meshroom to try to load the bundled
simpleFarmSubmitter plugin twice, ignoring the path in the variable.

I believe the problem lies with the way loadPlugins() works in meshroom.core,
because it's trying to import the package name "submitters" more than once.  The
second time, import_module() just returns a reference to the

This is a revised PR based on feedback from the first one I submitted last year,
which went stale and then got erased as I was stumbling my way through git.


fabiencastan:
Sorry for not having merged that earlier.

PR #629, opened 2019-09-10
[GitHub] add template for questions / help only 

simogasp:
Add a template only for questions and help needed, not related to program crashes and failures.


simogasp:
also, we can now remove the automatic stale and use the one provided by github which apparently is able to distinguish between issues and prs.

PR #631, opened 2019-09-12
Log manager for Python based nodes

ChemicalXandco:
## Description

There is currently no easy way for nodes that are executed in python to use the log file, this system makes adding a message as simple as `chunk.logger.info('message')`, currently the Publish node uses `print()` but this is not very helpful because it is not stored in a file that can be viewed afterwards.

```
chunk.logManager.waitUntilCleared()
chunk.logger.setLevel(chunk.logManager.textToLevel(chunk.node.verboseLevel.value))

chunk.logger.info('info message')
chunk.logger.warning('warning message')
chunk.logger.error('error message')

chunk.logManager.makeProgressBar(100)
for i in range(101):
    chunk.logManager.updateProgressBar(i)
chunk.logManager.completeProgressBar()
```
becomes
```
[18:06:47.016][info] info message
[18:06:47.022][warning] warning message
[18:06:47.028][error] error message
0%   10   20   30   40   50   60   70   80   90   100%
|----|----|----|----|----|----|----|----|----|----|
***************************************************
```

## Features list

* LogManager class to manage the logs
* Simulation of `boost::progress_display` within LogManager
* Updated Publish node to use this system

## Implementation remarks

* Progress bar can be used while adding messages
* Log clears automatically every time the chunk is ran

I do welcome feedback because I am not sure if this is the best way to implement this.


meshroom/core/node.py
```diff
@@ -142,6 +237,7 @@ def __init__(self, node, range, parent=None):
         super(NodeChunk, self).__init__(parent)
         self.node = node
         self.range = range
+        self.log = LogManager(self)
```
```
    def __init__(self, node, range, parent=None):
        ...
        self.logManager = LogManager(self)
        ...

    @property
    def logging(self):
        return self.logManager.logger
```

meshroom/nodes/aliceVision/Publish.py
```diff
@@ -41,23 +51,26 @@ def resolvedPaths(self, inputFiles, outDir):
...
 
         if not os.path.exists(chunk.node.output.value):
             os.mkdir(chunk.node.output.value)
 
         for iFile, oFile in outFiles.items():
-            print('Publish file', iFile, 'into', oFile)
+            chunk.log.add('Publish file {} into {}'.format(iFile, oFile))
             shutil.copyfile(iFile, oFile)
-        print('Publish end')
+        chunk.log.add('Publish end')
```
fabiencastan:
```
chunk.logging.info('Publish end')
```

meshroom/core/node.py
ChemicalXandco: I have called the property `logger` instead of `logging` because it makes more sense imo.

fabiencastan: yes, I agree.

TigerVersusT:
@ChemicalXandco HI, a stupid question，is the log here means printing messages on the console? It seems that every CLI based nodes( for example the texturing node ) has a log file in the cache folder

TigerVersusT:
@ChemicalXandco a stupid question, it seems that your logManager class is used for python based nodes only, how about CLI based node? Is it suitable to use your class for log implementtation in my own CLI nodes? And could you tell me somthing about how do nodes like "Texturing" in meshroom handle the log message.Thanks.

ChemicalXandco:
@TigerVersusT The C++ side of the software (AliceVision framework) uses it's own [logging system](https://github.com/alicevision/AliceVision/blob/979d6409f7be269036da070173d557418e52723f/src/aliceVision/system/Logger.hpp). By 'CLI based node', I'm assuming you mean an executable file not from the AliceVision framework. In that case, the best approach imo would be to forward the messages from stdout of the process to the logger implemented here.

TigerVersusT:
@ChemicalXandco Thanks for your help!

PR #632, opened 2019-09-12
Improve command line arguments

fabiencastan:
## Description

This PR improves meshroom_photogrammetry and meshroom command lines.

## Features list

- meshroom_photogrammetry: "--save" does not disable the computation so you can save your batch reconstruction project and open it later interactively in meshroom. Additionnal --compute option if you need only to create a meshroom project file from this command line.
- "meshroom -h / --help": now provides command line documentation
- meshroom: new positional argument to set project file
- meshroom: new "--pipeline" option to override the default pipeline with your own project file. It also supports an environment variable MESHROOM_DEFAULT_PIPELINE.
- meshroom_photogrammetry: new "--paramOverrides NODE.param=value" allows to set a parameter directly from the command line.

## Implementation

- minor change in photogrammetry() to keep the whole graph creation in a single GraphModification.

### DIFF
--- a/bin/meshroom_photogrammetry
+++ b/bin/meshroom_photogrammetry
@@ -11,15 +11,15 @@ import meshroom.core.graph
 from meshroom import multiview
 
 parser = argparse.ArgumentParser(description='Launch the full photogrammetry pipeline.')
-parser.add_argument('-i', '--input', metavar='FOLDER_OR_SFM', type=str,
-                    default='',
-                    help='Input folder containing images or file (.sfm or .json) '
-                         'with images paths and optionally predefined camera intrinsics.')
-parser.add_argument('--inputImages', metavar='IMAGES', type=str, nargs='*',
+parser.add_argument('-i', '--input', metavar='SFM/FOLDERS/IMAGES', type=str, nargs='*',
                     default=[],
-                    help='Input images.')
+                    help='Input folder containing images or folders of images or file (.sfm or .json) '
+                         'with images paths and optionally predefined camera intrinsics.')
+parser.add_argument('-I', '--inputRecursive', metavar='FOLDERS/IMAGES', type=str, nargs='*',
+                    default=[],
+                    help='Input folders containing all images recursively.')
 
-parser.add_argument('--pipeline', metavar='MESHROOM_FILE', type=str, required=False,
+parser.add_argument('-p', '--pipeline', metavar='MESHROOM_FILE', type=str, required=False,
                     help='Meshroom file containing a pre-configured photogrammetry pipeline to run on input images. '
                          'If not set, the default photogrammetry pipeline will be used. '
                          'Requirements: the graph must contain one CameraInit node, '
@@ -72,24 +72,24 @@ def getOnlyNodeOfType(g, nodeType):
     return nodes[0]
 
 
-if not args.input and not args.inputImages:
-    print('Nothing to compute. You need to set --input or --inputImages.')
+if not args.input and not args.inputRecursive:
+    print('Nothing to compute. You need to set --input or --inputRecursive.')
     sys.exit(1)
 
 views, intrinsics = [], []
 # Build image files list from inputImages arguments
-images = [f for f in args.inputImages if multiview.isImageFile(f)]
+images = []
 
 if args.input:
-    if os.path.isdir(args.input):
-        # args.input is a folder: extend images list with images in that folder
-        images += multiview.findImageFiles(args.input)
-    elif os.path.isfile(args.input) and os.path.splitext(args.input)[-1] in ('.json', '.sfm'):
+    if len(args.input) == 1 and os.path.isfile(args.input[0]) and os.path.splitext(args.input[0])[-1] in ('.json', '.sfm'):
         # args.input is a sfmData file: setup pre-calibrated views and intrinsics
         from meshroom.nodes.aliceVision.CameraInit import readSfMData
-        views, intrinsics = readSfMData(args.input)
+        views, intrinsics = readSfMData(args.input[0])
     else:
-        raise RuntimeError(args.input + ': format not supported.')
+        images += multiview.findImageFiles(args.input, recursive=False)
+
+if args.inputRecursive:
+    images += multiview.findImageFiles(args.inputRecursive, recursive=True)
 
 # initialize photogrammetry pipeline
 if args.pipeline:
--- a/bin/meshroom_photogrammetry
+++ b/bin/meshroom_photogrammetry
@@ -38,7 +38,7 @@ parser.add_argument('--cache', metavar='FOLDER', type=str,
                          'If not set, the default cache folder will be used: ' + meshroom.core.defaultCacheFolder)
 
 parser.add_argument('--save', metavar='FILE', type=str, required=False,
-                    help='Save the configured Meshroom project to a file.')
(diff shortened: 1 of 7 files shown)


meshroom/core/graph.py
```diff
@@ -234,7 +234,15 @@ def fileFeatures(self):
...
 
     @Slot(str)
-    def load(self, filepath):
+    def load(self, filepath, fileLink=True):
+        """
+        Load a meshroom graph ".mg" file.
+
+        Args:
+            filepath: project filepath to load
+            fileLink: Setup link to the project file, like setup cacheDir, keep filepath for save, etc.
```
simogasp:
```suggestion
            fileLink: if false, it only loads the project file as a template (all the data is discarded) and only the bare structure of the graph is preserved.
```
I found `Setup link to the project file, like setup cacheDir, keep filepath for save, etc.` a bit obscure/confusing

meshroom/core/graph.py
```diff
@@ -234,7 +234,15 @@ def fileFeatures(self):
         return Graph.IO.getFeaturesForVersion(self.header.get(Graph.IO.Keys.FileVersion, "0.0"))
 
     @Slot(str)
-    def load(self, filepath):
+    def load(self, filepath, fileLink=True):
```
simogasp: `fileLink` is a bit obscure.
why not turn the table and use `asTemplate=False`. I know that then all the conditions should be negate but it is less obscure than fileLink IMO.

meshroom/ui/app.py
```diff
@@ -105,12 +117,24 @@ def __init__(self, args):
...
+            # consider the environment variable
+            defaultPipeline = os.environ.get("MESHROOM_DEFAULT_PIPELINE", "")
+            if defaultPipeline:
+                r.setDefaultPipeline(args.pipeline)
+
         if args.project:
-            r.loadUrl(QUrl.fromLocalFile(args.project))
+            r.load(args.project)
+
+        if args.input:
```
simogasp: shouldn't `input` and `project` be exclusive?

At least if `project` and `input` both contain a .mg. In this case, I think it would be better to throw an error to let the user know that there is something wrong. (unless it is able to load both projects?)

And if we give `project` and `input` with folders, does it add the images to the current project? Something in the same spirit of meshroom_photogrammetry https://github.com/alicevision/meshroom/pull/632/files#diff-4f70151d95f3b3a1c7ea377cbaef3f10R114
```python
# initialize photogrammetry pipeline
if args.pipeline:
    # custom pipeline
    graph = meshroom.core.graph.loadGraph(args.pipeline)
    cameraInit = getOnlyNodeOfType(graph, 'CameraInit')
    # reset graph inputs
    cameraInit.viewpoints.resetValue()
    cameraInit.intrinsics.resetValue()
    # add views and intrinsics (if any) read from args.input
    cameraInit.viewpoints.extend(views)
    cameraInit.intrinsics.extend(intrinsics)

    if not graph.canComputeLeaves:
        raise RuntimeError("Graph cannot be computed. Check for compatibility issues.")

    if args.output:
        publish = getOnlyNodeOfType(graph, 'Publish')
        publish.output.value = args.output
else:
    # default pipeline
    graph = multiview.photogrammetry(inputViewpoints=views, inputIntrinsics=intrinsics, output=args.output)
    cameraInit = getOnlyNodeOfType(graph, 'CameraInit')

if images:
    views, intrinsics = cameraInit.nodeDesc.buildIntrinsics(cameraInit, images)
    cameraInit.viewpoints.value = views
    cameraInit.intrinsics.value = intrinsics
```

fabiencastan: Yes, there is no error for now but that's the usage intention.
So you're right, we can enforce the fact that it should be exclusive.

meshroom/core/graph.py
fabiencastan: I would prefer to avoid inverted negation: `asTemplate=False` will do more actions (setup the cacheDir, etc.) and `asTemplate=True` to do nothing.
I pushed a proposition for another name that doesn't negate the option: `setupFileRef`.
Does that sound fine to you?

meshroom/ui/app.py
fabiencastan: This is changed and I have also added another argument "--import".

meshroom/ui/app.py
```diff
@@ -61,9 +61,13 @@ def __init__(self, args):
 
         parser.add_argument('input', metavar='INPUT', type=str, nargs='?',
                             help='Meshroom project file (e.g. myProject.mg) or folder with images to reconstruct.')
-        parser.add_argument('--project', metavar='MESHROOM_FILE', type=str, required=False,
+        parser.add_argument('-i', '--import', metavar='IMAGES/FOLDERS', type=str, nargs='*',
```
simogasp: Isn't now the arguments a little bit too redundant? 
We can get rid of `input` at this point and just use project and the imports. It makes life easier, it's more readable and less confusing, and we don't need to check for conflicts. And too bad if we break compatibility, we already broke it with photogrammetry.

meshroom/core/graph.py
```diff
@@ -234,14 +234,14 @@ def fileFeatures(self):
         return Graph.IO.getFeaturesForVersion(self.header.get(Graph.IO.Keys.FileVersion, "0.0"))
 
     @Slot(str)
-    def load(self, filepath, fileLink=True):
+    def load(self, filepath, setupFileRef=True):
```
simogasp: what about `overrideProjectFile`?

meshroom/core/graph.py
```diff
@@ -234,7 +234,15 @@ def fileFeatures(self):
...
 
     @Slot(str)
-    def load(self, filepath):
+    def load(self, filepath, setupFileRef=True):
+        """
+        Load a meshroom graph ".mg" file.
+
+        Args:
+            filepath: project filepath to load
+            setupFileRef: Setup reference to the project file, like setup cacheDir, keep filepath for save, etc.
```
simogasp:
```suggestion
            overrideProjectFile: Override the input project file (any changes will be overwritten on the file) and use its cache directories. If false, it only loads the graph of the project file as a template.
```

bin/meshroom_photogrammetry
```diff
@@ -125,6 +128,29 @@ if args.overrides:
...
+        if t == ':':
+            nodesByType = graph.nodesByType(node)
+            if not nodesByType:
+                raise ValueError('No node with the type "{}" in the scene.'.format(node))
+            for n in nodesByType:
+                print('Overrides {node}.{param}={value}'.format(node=node, param=param, value=value))
+                n.attribute(param).value = value
+        elif t == '.':
+            print('Overrides {node}.{param}={value}'.format(node=node, param=param, value=value))
+            graph.findNode(node).attribute(param).value = value
```
simogasp: just asking, `findNode(node)` can fail in the same way as `nodesByType(node)`? Shall we manage the error?
Also, very minor thing, I'd put the print after the setting is done so it is not print in case of failure (hopefully `findNode(node)` or `attribute(param)` will print or raise some kind of error?)

In case, the same for the previous `nodesByType(node)`

fabiencastan: findNode is looking for one specific node and fails with an explicit message if it doesn't find it.
nodesByType looks for all nodes of a specific type and can return an empty list without error (that's why there is a specific check to provide an error message).
I put the print before so it's more visible in the error message :s, the last NODE.PARAM is printed and then you have the error message saying that it doesn't find PARAM for instance.

PR #650, opened 2019-09-22
[WIP] Dialog to view depth maps

ChemicalXandco:
## Description

This PR makes it easier to view depth maps - currently, the only way to view them is to drag and drop from the cache folder

example

## Features list

- [X] List of depth map files
- [X] 2D viewer for the selected file
- [X] Ability to load depth map into 3D viewer within Meshroom
- [ ] Filter depth maps by type
- [ ] Make feature overlay work properly on the depth map file on 2D viewer

## Implementation remarks

* Dialog is opened by double-clicking either a DepthMap or DepthMapFilter node
* Refresh button allows new depth maps to be viewed while the node is running

PR #657, opened 2019-09-25
Ability to load depth maps into 2D and 3D Viewers within meshroom

ChemicalXandco:
## Description

A better approach to #650 

## Features list

* Combo box to select image type (image, depthMap, simMap)
* Button to view depth map of the current image in the 3D viewer
* Active node that can be either DepthMap or DepthMapFilter

## Implementation remarks

DepthMap and DepthMapFilter are treated the same to avoid extra complexity
I haven't included nModMap because it does not seem to be important


natowi:
Thank you for this feature, it works well.
Only setting the Display Trackball does not work as reliably as with the point cloud or mesh. You sometimes need to click multiple times in an area to set the new rotation point.

fabiencastan:
Merged in https://github.com/alicevision/meshroom/pull/769.

Issue #674, opened 2019-10-08
FeatureExtraction - SURF and BRISK


leviresende:
For a while, I am using the software Meshroom, and then I just decided to try using other algorithms for features extraction, such as SURF and BRISK, and by default, these 2 aren't available there. So, my question: is it possible to implement these two algorithms in some script and then attached it to the software? Whether yes, how can I do it?

fabiencastan:
Hi Levi,

It would be interesting to add support for more feature types into Meshroom.

You can implement new feature detectors here:
https://github.com/alicevision/AliceVision/tree/develop/src/aliceVision/feature
And declare the new describers here:
https://github.com/alicevision/AliceVision/blob/develop/src/aliceVision/feature/imageDescriberCommon.hpp

You have some examples of feature types from OpenCV here:
https://github.com/alicevision/AliceVision/tree/develop/src/aliceVision/feature/openCV
So if you plan to add SURF and BRISK from opencv implementation, that would be a very good starting point.

And finally add the arguments on the Meshroom nodes (FeatureExtraction, FeatureMatching, StructureFromMotion) declaration:
https://github.com/alicevision/meshroom/tree/develop/meshroom/nodes/aliceVision

We can setup a confcall if you want to discuss about the implementation details.

leviresende:
Thanks @fabiencastan !

I already have these algorithms in python, but I saw that the codes of the extractors are in c++  . First of all, I'm gonna convert them, and then I'll let you know. Or, if you would rather we can to have a confcall first.

fabiencastan:
Sounds good to me!
You can create a Pull Request with "[WIP]" in the title, so we can follow your progression and put remarks while you're working on it to ensure that it will be merged at the end.

leviresende:
@fabiencastan as feedback I'll summarize what I'm doing (please correct me if something is wrong.)
As I'll work with OpenCV, I need to create the files:
-> ImageDescriber_SURF_OCV.cpp (done)
-> ImageDescriber_SURF_OCV.hpp (done)
-> ImageDescriber_BRISK_OCV.cpp
-> ImageDescriber_BRISK_OCV.hpp

For SURF I'll use the  SIFT algorithms as a base, and for BRISK I'll use AKAZE algorithms.

Once I have that files, I should edit the folowing files:
-> regionsFactory.hpp
-> FeatureExtraction.py
-> imageDescriberCommom.cpp
-> imageDescriberCommom.hpp
-> ImageDescriber.cpp

[...]

PR #676, opened 2019-10-11
Describer types (new, sorted)

natowi:
## Description
Added brisk_ocv and surf_ocv describers for https://github.com/alicevision/meshroom/issues/674
to FeatureExtraction, FeatureMatching, StructureFromMotion nodes.

## Implementation remarks

I grouped the describers by native, ocv and cctag, sorted in alphabetical order:

values=['akaze', 'akaze_liop', 'akaze_mldb', 'sift', 'sift_float', 'sift_upright', 'akaze_ocv', 'brisk_ocv', 'sift_ocv', 'surf_ocv', 'cctag3', 'cctag4'],

Another logical order could be the following:

values=['sift', 'sift_ocv', 'sift_float', 'sift_upright', 'akaze', 'akaze_ocv', 'akaze_liop', 'akaze_mldb', 'brisk_ocv', 'surf_ocv', 'cctag3', 'cctag4'],


simogasp:
A good idea would be to have some sort of general setting/variable that contains the list in order to avoid code replication in each concerned node (ie declare the list once and use it in each node). That is more maintainable than having the the list replicated in several parts of the code.

Also, I don't know if it is feasible but the list should only contain the supported features for the given alicevision binaries: I don't know if we can use the env variable to go see the config.hpp and check e.g. is opencv is supported and drop all relevant features if not. The same for cctag etc...

fabiencastan:
Yes, good idea, we can declare a list in FeatureExtraction.py and import that from FeatureMatching and StructureFromMotion.
To keep it simple, we will assume that opencv is available and this will be the case in the releases.

vic4key:
Please merge this feature @fabiencastan. 
P.S Sorry for pulling this thread up after a long time. But these algorithms might better for several cases. It should be merged.

fabiencastan:
The only pbl afaik is that it needs to be finalized.

PR #680, opened 2019-10-17
Depth map refactoring

fabiencastan:
## Description

See https://github.com/alicevision/AliceVision/pull/619 for more details.

## Features list

- [X] Expose more internal parameters


meshroom/nodes/aliceVision/DepthMap.py
```diff
@@ -52,6 +52,38 @@ class DepthMap(desc.CommandLineNode):
...
+            label='SGM: Step Z',
+            description='Semi Global Matching: Step used to compute the similarity volume on Z axis.',
+            value=-1,
+            range=(-1, 10, 1),
+            uid=[0],
+        ),
+        desc.IntParam(
+            name='sgmMaxSideXY',
+            label='SGM: Max Side',
+            description='Semi Global Matching: Max side in pixels used to automatically decide for sgmScale/sgmStep if not defined.',
```
natowi, line 82: can we use Max Width instead of Max Side ?

gregoire-dl (approved):

Issue #690, opened 2019-11-02
Add option to export dense point cloud to other formats (Workaround here)


miquelrosell99:
**Describe the bug**
PrepareDenseScene node fails for ConvertSfMFormat input in .ply, so i can only export .ply of low density point clouds
![Captura](https://user-images.githubusercontent.com/38435442/68072516-018de700-fd87-11e9-9563-e03de96ddeb4.JPG)

natowi:
You can simply change the Inter File Extension from abc to ply in the StructureFromMotion node (Advanced settings enabled) so you don´t need the ConvertSfMFormat node in this case.

miquelrosell99:
> You can simply change the Inter File Extension from abc to ply in the StructureFromMotion node (Advanced settings enabled) so you don´t need the ConvertSfMFormat node in this case.

But what if I need the dense point cloud? Meshing node only outputs .abc file, and I can't manage to use ConverSfM node there
Your suggestion was very useful for SfM node. thank you very much. I would really appreciate any help in the issue mentioned above. It's for a final degree project

natowi:
You can export the Raw Dense Point Cloud by enabling the advanced settings in the Meshing node. You might also want to enable "Colourize Output" to get a coloured point cloud.

miquelrosell99:
> You can export the Raw Dense Point Cloud by enabling the advanced settings in the Meshing node. You might also want to enable "Colourize Output" to get a coloured point cloud.

But the raw file is in .abc too. I need .ply

natowi:
Let me check a few things

natowi:
- [ConvertSfMFormat ](https://meshroom-manual.readthedocs.io/en/latest/node-reference/nodes/ConvertSfMFormat.html) at the moment only supports the internal sfm filetype

- Meshing node output filetype for the pointcloud can not be changed at the moment, although it should be possible as of this [PR](https://github.com/alicevision/AliceVision/pull/597)

- You can import the abc file in Blender (I am using v2.8) and then export the point cloudas obj (ply did not work for me...), then import the Point Cloud to Meshlab and File->ExportMeshAs->PLY

miquelrosell99:
> * [ConvertSfMFormat ](https://meshroom-manual.readthedocs.io/en/latest/node-reference/nodes/ConvertSfMFormat.html) at the moment only supports the internal sfm filetype
> * Meshing node output filetype for the pointcloud can not be changed at the moment - in the GUI!
> 
[...]

PR #691, opened 2019-11-02
[WIP] Add file type selection for dense point cloud

natowi:
## Description

This will (hopefully) add file type selection for dense point cloud

https://github.com/alicevision/meshroom/issues/690

https://github.com/alicevision/AliceVision/pull/597

## Implementation remarks

ToDo: 

[ ] find a way to exclude the file type drop down from the class
[ ] include rawPointCloud file type

### DIFF
--- a/meshroom/nodes/aliceVision/Meshing.py
+++ b/meshroom/nodes/aliceVision/Meshing.py
@@ -256,7 +256,7 @@ class Meshing(desc.CommandLineNode):
             name="output",
             label="Output Dense Point Cloud",
             description="Output dense point cloud with visibilities.",
-			value=desc.Node.internalFolder + 'densePointCloud.{outputSfMDataFilenameValue}',
+            value=desc.Node.internalFolder + 'densePointCloud.{outputSfMDataFilenameValue}',
             uid=[],
         ),
     ]
--- a/meshroom/nodes/aliceVision/Meshing.py
+++ b/meshroom/nodes/aliceVision/Meshing.py
@@ -225,7 +225,7 @@ class Meshing(desc.CommandLineNode):
             advanced=True,
         ),
         desc.ChoiceParam(
-            name='outputSfMDataFilename',
+            name='outputDensePointCloudFileType',
             label='Dense Point Cloud File Type',
             description='Dense Point Cloud File Type',
             value='abc',
@@ -255,8 +255,8 @@ class Meshing(desc.CommandLineNode):
         desc.File(
             name="output",
             label="Output Dense Point Cloud",
-            description="Output dense point cloud with visibilities (SfMData file format).",
-            value="{cache}/{nodeType}/{uid0}/'densePointCloud*.{outputSfMDataFilename}'",
+            description="Output dense point cloud with visibilities.",
+			value=desc.Node.internalFolder + 'densePointCloud.{outputDensePointCloudFileTypeValue}',
             uid=[],
         ),
     ]
--- a/meshroom/nodes/aliceVision/Meshing.py
+++ b/meshroom/nodes/aliceVision/Meshing.py
@@ -4,7 +4,7 @@ from meshroom.core import desc
 
 
 class Meshing(desc.CommandLineNode):
-    commandLine = 'aliceVision_meshing {allParams}'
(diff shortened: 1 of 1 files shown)


natowi:
Is there a reason for using "{cache}/{nodeType}/{uid0}/mesh.obj" in outputMesh instead of desc.Node.internalFolder ?

natowi:
The PR note by @gregoire-dl says "This PR adds the possibility to export a dense point cloud as a SfMData file (.abc, .json, ...)", but the CLI does not provide this option. (Attaching the desired file type does work, but is not documented. The note `"Output SfMData filename (.json, .bin, .xml, .ply, .baf"` was replaced by `"Output point cloud with visibilities as SfMData file."` with the last PR).

For me it looks like the main_convertSfMFormat.cpp needs to be modified for my MR PR to work.

I think there is no option for the file type defined yet - or am I missing something?
(c++ is new for me)

At the moment there is only
`std::string outputSfMDataFilename;`

Do we have to add something like:

`std::string outputSfMDataFilename = sfmDataIO::Save(sfmDataIO::FileType::ABC);`

 ```
// user optional parameters

("outputSfMDataFilename", po::value<std::string>(&outSfMDataFilename)->default_value(outSfMDataFilename),
          sfmDataIO::Save(sfmData, outputSfMDataFilename, sfmDataIO::ESfMData(flags)),
      "Output SfMData filename (.abc, .json, .bin, .xml, .ply, .baf)"
```

/ I also noted exportPointCloud((outDirectory/"densePointCloud_raw.abc") is hardcoded

**Comment by fabiencastan** [comment]
> Is there a reason for using "{cache}/{nodeType}/{uid0}/mesh.obj" in outputMesh instead of desc.Node.internalFolder ?
No you could do:
`            value=desc.Node.internalFolder + 'mesh.obj',`

**Comment by fabiencastan** [comment]
> I think there is no option for the file type defined yet - or am I missing something?

The file type is just defined by the file extension, so there is no need for an extra argument.

**Comment by natowi** [comment]
> extension

I tried my PR running MR from CLI, but get the following error when computing the Meshing node:

```
[2019-11-04 10:42:06,376][ERROR] Error on node computation: Error on node "Meshing_1":
Log:
ERROR: unrecognised option '--outputSfMDataFilename'
```

**Comment by fabiencastan** [comment]
See the existing parameters here:
https://github.com/alicevision/AliceVision/blob/develop/src/software/pipeline/main_meshing.cpp#L157

**Comment by natowi** [comment]
Ah, so it is outputDensePointCloud and not outputSfMDataFilename

**Comment by natowi** [comment]
At the moment the drop down selector is in
```
class Meshing(desc.CommandLineNode):
    commandLine = 'aliceVision_meshing {allParams}'
```
This results in the unrecognized option error...

**Comment by natowi** [comment]
This is a workaround for those who want to change the file type in the gui:

add 
```
desc.File(
            name="output",
            label="Output Dense Point Cloud",
            description="Output dense point cloud('abc', 'ply', 'json', 'xml', 'baf', 'bin').",
			value='',
            uid=[],
        ),
```
as last entry within `inputs = [       ]`
This allows for a custom path to be defined for the dense point cloud.

@fabiencastan Is it possible to allow editing the paths in `outputs = [       ]` as it is possible for inputs or is there another solution for changing the file type in nodes who do not have a cli expression for the file type? I did not find a way for the drop down menu to not be recognized as cli option. Any hints?

**Comment by natowi** [comment]
From what I learned from https://github.com/alicevision/meshroom/issues/697 I guess I could exclude the select file type option from the cli commands, but now I need the -- .

What is the correct syntax to replace _{allParams}_ with a list of the supported parameters (excluding outputDenseFileType)?

```
class Meshing(desc.CommandLineNode):
    commandLine = 'aliceVision_meshing {allParams}'
```

**Comment by natowi** [comment]
Ok I figured it out:

```
class Meshing(desc.CommandLineNode):
    commandLine = 'aliceVision_meshing --input {inputValue}
``` 

This way it should be possible to list only the supported options.

natowi:
Remaining problem: Texturing node expects .abc input format

fabiencastan:
Don't modify the full pipeline to use another file format, just use ConvertSfM where you need to convert your Alembic file to something else.

fabiencastan:
Most of the file formats are not able to store all the information we need: cameras, visibilities, etc.
For instance, PLY is only able to deal with point clouds without extra-attributes.

natowi:
I have tried convert sfm but it does not work with the densepointcloud.abc as input. The resulting ply is 1kb in size

fabiencastan:
Maybe some feature types are not enabled on the ConvertSfMFormat?
In all case, it's supposed to work: either it is a parameter problem or a bug.

natowi:
StructureFromMotion sfm.abc -> ConvertSfMFormat will convert the whole pointcloud

Meshing densePointCloud.abc -> ConvertSfMFormat will only convert the cameras and dismiss the pointcloud

fabiencastan:
You should check the "unknown" descriptor type as the points have been created by the depth maps without the notion of feature/descriptor.

natowi:
There is no "unknown" descriptor option in the cli or gui. I unselected sift but this does not work.

PR #711, opened 2019-11-24
Fix typo, add reference

natowi:
add title of citation to comments to better find the publication https://dl.acm.org/citation.cfm?id=2818068


meshroom/nodes/aliceVision/MeshDenoising.py
```diff
@@ -69,7 +69,8 @@ class MeshDenoising(desc.CommandLineNode):
             label='Mesh Update Method',
             description='Mesh Update Method\n'
                         ' * ITERATIVE_UPDATE (default): ShapeUp styled iterative solver \n'
-                        ' * POISSON_UPDATE: Poisson-based update from [Want et al. 2015]',
+                        ' * POISSON_UPDATE: Poisson-based update from [Wang et al. 2015]',
```
fabiencastan: Could be added in the description itself, isn't it?
`* POISSON_UPDATE: Poisson-based update from [Wang et al. 2015] "Rolling guidance normal filter for geometric processing"`

PR #712, opened 2019-11-25
Sketchfab upload node

ChemicalXandco:
## Description

Sketchfab is a popular website to share and view 3D files, this provides a node to allow direct upload to Sketchfab from meshroom using the data api
This PR is based off the same branch in my fork as #631 in order to use the logging functions here.
The code here is based on [the data api with python](https://sketchfab.com/developers/data-api/v3/python).

## Features list

- a node that allows files to be uploaded to Sketchfab
- API key is provided by the user in the node settings
- progress bar to track the progress of the upload

## Implementation remarks

I have tried using urllib to avoid having to add requests dependency, but in order for this to work it would need other dependencies anyway so requests is the best option here. I also tried using pyside QNetwork and did get it working, but in order for this to work in a node it would need to run on a separate process because of the QApplications conflicting, which is not ideal.

Currently, the API key needs to be put in for every new node, but I will look into a solution to overwrite the default node settings so that it can be automatically added when a new node is created. I will create a separate PR for this if I get round to it.


meshroom/nodes/aliceVision/SketchfabUpload.py
```diff
@@ -0,0 +1,200 @@
...
+            uid=[0],
+        ),
+        desc.StringParam(
+            name='description',
+            label='Description',
+            description='Description cannot be longer than 1024 characters.',
+            value='',
+            uid=[0],
+        ),
+        desc.StringParam(
```
fabiencastan: Maybe good to put a Choice with predefined list of possible licenses? Or at least put standard values in the tooltip.

meshroom/nodes/aliceVision/SketchfabUpload.py
```diff
@@ -0,0 +1,200 @@
...
+            'license': chunk.node.license.value,
+            'isPublished': chunk.node.isPublished.value,
+            'isInspectable': chunk.node.isInspectable.value
+        }
+        # pack files into .zip to reduce file size and simplify process
+        uploadFile = os.path.join(chunk.node.internalFolder, 'temp.zip')
+        files = self.resolvedPaths(chunk.node.inputFiles.value)
+        zf = zipfile.ZipFile(uploadFile, 'w')
+        for file in files:
+            zf.write(file, os.path.basename(file))
```
fabiencastan: Does this work when we connect Texture.outputTextures which contains the simplified regex: "texture_*.png"?

meshroom/nodes/aliceVision/SketchfabUpload.py
```diff
@@ -0,0 +1,200 @@
...
+        desc.StringParam(
+            name='license',
+            label='License',
+            description='License label.',
+            value='CC Attribution',
+            uid=[0],
+        ),
+        desc.BoolParam(
+            name='isPublished',
+            label='Publish',
```
natowi: I think it would be better to save the model in [Draft Mode](https://help.sketchfab.com/hc/en-us/articles/115000472906-Draft-Mode) by default. You likely want to fix the model orientation and lighting before publishing.

fabiencastan: The default value is already False.

meshroom/nodes/aliceVision/SketchfabUpload.py
```diff
@@ -0,0 +1,200 @@
...
+            value=False,
+            uid=[0],
+        ),
+        desc.BoolParam(
+            name='isInspectable',
+            label='Inspectable',
+            description='Allow 2D view in model inspector.',
+            value=True,
+            uid=[0],
+        ),
```
natowi: + Private Model: OFF (default) and Age-Restricted content: OFF (default) 
to complete the model properties

meshroom/nodes/aliceVision/SketchfabUpload.py
```diff
@@ -0,0 +1,200 @@
...
+            value='',
+            uid=[0],
+        ),
+        desc.StringParam(
+            name='description',
+            label='Description',
+            description='Description cannot be longer than 1024 characters.',
+            value='',
+            uid=[0],
+        ),
```
natowi: How about adding Categories (18) and Tags (comma separated)? This way the model Metadata would be complete.

meshroom/nodes/aliceVision/SketchfabUpload.py
natowi: I see, missed this.

fabiencastan:
Hi @ChemicalXandco,
Thanks for the contribution! That's a really interesting PR.
As a side note for potential futur contributions ;), it may be interesting to experiment with new nodes based on blender for the post-processing steps before uploading on sketchfab, as described here:
https://sketchfab.com/blogs/community/tutorial-low-poly-assets-from-3d-scans

meshroom/nodes/aliceVision/SketchfabUpload.py
ChemicalXandco: Yes because it goes through `resolvedPaths` which uses glob, like in the publish node. However, it does not work when Texture.output is connected because only the folder is copied, not the files in the folder. I will fix this

meshroom/nodes/aliceVision/SketchfabUpload.py
```diff
@@ -0,0 +1,282 @@
...
+            ),
+            name="tags",
+            label="Tags",
+            description="Maximum of 42 separate tags.",
+            group="",
+        ),
+        desc.ChoiceParam(
+            name='category',
+            label='Category',
+            description='Maximum of 2 categories.',
```
fabiencastan: Exclusive is set to True, so in all case, we can choose only one category.
So the description or the "exclusive" mode should be adjusted.

meshroom/nodes/aliceVision/SketchfabUpload.py
```diff
@@ -0,0 +1,282 @@
...
+            uploadFile = os.path.join(chunk.node.internalFolder, 'temp.zip')
+            files = self.resolvedPaths(chunk.node.inputFiles.value)
+            zf = zipfile.ZipFile(uploadFile, 'w')
+            for file in files:
+                zf.write(file, os.path.basename(file))
+            zf.close()
+            chunk.logger.debug('Files added to zip: {}'.format(str(files)))
+            chunk.logger.debug('Created {}'.format(uploadFile))
+            
+            fileSize = os.path.getsize(uploadFile)/1000000
```
fabiencastan: `fileSize = os.path.getsize(uploadFile)/(1024*1024)`

fabiencastan:
@ChemicalXandco Thank you for this nice contribution!

Issue #715, opened 2019-11-29
Stuck at meshing


umustbeloggedintododat:
Hey it says this do not call this duplicate cause there's a different thing in the log happened to me

This is my log 
Plugins loaded:  CameraCalibration, CameraInit, CameraLocalization, CameraRigCalibration, CameraRigLocalization, ConvertSfMFormat, DepthMap, DepthMapFilter, ExportAnimatedCamera, ExportColoredPointCloud, ExportMaya, FeatureExtraction, FeatureMatching, ImageMatching, ImageMatchingMultiSfM, KeyframeSelection, LDRToHDR, MeshDecimate, MeshDenoising, MeshFiltering, MeshResampling, Meshing, PrepareDenseScene, Publish, SfMAlignment, SfMTransform, StructureFromMotion, Texturing
Program called with the following parameters:
 * allowSingleView = 1
 * defaultCameraModel = "" (default)
 * defaultFieldOfView = 45
 * defaultFocalLengthPix = -1 (default)
 * defaultIntrinsic = "" (default)
 * groupCameraFallback =  Unknown Type "enum EGroupCameraFallback"
 * imageFolder = "" (default)
 * input = "C:\Users\Ina rosario\AppData\Local\Temp\tmpjnvfmoma/CameraInit/c448939571d5c70b05c9ae4ad416ada37d6273ca//viewpoints.sfm"
 * output = "C:/Users/Ina rosario/AppData/Local/Temp/tmpjnvfmoma/CameraInit/c448939571d5c70b05c9ae4ad416ada37d6273ca/cameraInit.sfm"
 * sensorDatabase = "C:\Users\Ina rosario\Desktop\Keegan's stuff\test\aliceVision\share\aliceVision\cameraSensors.db"
 * verboseLevel = "info"

[14:11:34.499019][warning] Some image(s) have no serial number to identify the camera/lens device.
This makes it impossible to correctly group the images by device if you have used multiple identical (same model) camera devices.
The reconstruction will assume that only one device has been used, so if 2 images share the same focal length approximation they will share the same internal camera parameters.
38 image(s) are concerned.
[14:11:34.559253][info] CameraInit report:
        - # views listed: 38
           - # views with an initialized intrinsic listed: 38
           - # views without metadata (with a default intrinsic): 0
        - # intrinsics listed: 1
[...]

PR #750, opened 2020-01-02
WIP Masking interface

ChemicalXandco:
## Description

The purpose of this is to make masks user guided within meshroom with minimal user interaction, which when combined with #708 and [alicevision/AliceVision #715](https://github.com/alicevision/AliceVision/pull/715) will get another option for masking workflow, rather than relying on thresholding which will only work for some setups.

## Features list

- Window overlay accessible from 'Edit' menu
- Masks generated using the GrabCut algorithm 
- The algorithm is sped up by setting iterations to 1 and lowering image resolution

screenshot

## Implementation remarks

Given the structure of meshroom, it would probably make more sense to have the mask calculated in c++ and then the exe ran from python to avoid adding opencv dependency when it is already on AliceVision, so I will look into this possibility

### DIFF
--- a/meshroom/ui/app.py
+++ b/meshroom/ui/app.py
@@ -12,6 +12,7 @@ from meshroom.ui import components
 from meshroom.ui.components.clipboard import ClipboardHelper
 from meshroom.ui.components.filepath import FilepathHelper
 from meshroom.ui.components.scene3D import Scene3DHelper
+from meshroom.ui.mask import Masking
 from meshroom.ui.palette import PaletteManager
 from meshroom.ui.reconstruction import Reconstruction
 from meshroom.ui.utils import QmlInstantEngine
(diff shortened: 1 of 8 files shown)


natowi:
Implementation remark: How about adding the Masking GUI to the image viewer as you did for the depth map viewer?  (Enable masks similar to feature viewer, display configuration options like the metadata field)  This would better integrate into the GUI.
- add editable output path to options

ChemicalXandco:
@natowi I avoided adding it to the image viewer because that would mean having the panel as an overlay which is not very practical because it would have to be hidden in order to draw on the canvas or drag the rectangle (otherwise it would be in the way), which is not practical really imo. Having it in another window allows the image to be displayed much bigger, which I think is quite important. Lmk if you disagree.

natowi:
@ChemicalXandco Ok, maybe this can be changed in the future, when the panes in the main gui are made detachable.

PR #778, opened 2020-02-05
[core] Option to make chunks optional

fabiencastan:
When computing on a single machine with very large ressources, the chunks can limit the ressources used.
A new environment variable `MESHROOM_USE_MULTI_CHUNKS` can be set to 0 to disable it.

Warning: If you mix usage with and without this option, the nodes status will be incoherent.

### DIFF
--- a/meshroom/__init__.py
+++ b/meshroom/__init__.py
@@ -1,10 +1,11 @@
 __version__ = "2019.2.0"
 __version_name__ = __version__
 
+from distutils import util
+from enum import Enum
+import logging
 import os
 import sys
-import logging
-from enum import Enum
 
 # sys.frozen is initialized by cx_Freeze and identifies a release package
 isFrozen = getattr(sys, "frozen", False)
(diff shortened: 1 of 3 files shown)


natowi:
Thank you for this feature.

>  A new environment variable MESHROOM_USE_MULTI_CHUNKS can be set to 0 to disable it.

Like this?:
```

set MESHROOM_USE_MULTI_CHUNKS=0
set MESHROOM_INSTALL_DIR=%CD%
set PYTHONPATH=%CD% && python meshroom/ui
```

Will there be a checkbox in the Meshroom main menu?

fabiencastan:
Yes, that's it.

It could become part of the settings and integrated in the menus. But for now, it's really an advanced feature for specific needs and we need to measure the impact before going further.

natowi:
> we need to measure the impact before going further.

Reduces overall computation time by ~20% (tested with Monstree full), tested on  i7-7820, GTX1070, 32GB
I will do a second test on a workstation.

natowi:
I tested this on a 48 core workstation. The speed-up is not huge, it reduces the overall computation time by ~10%. This reduces computation time in FE, SfM, DM and M but increases computation time slightly in DMF and TEX.

PR #780, opened 2020-02-07
[nodes] process must not be blocked by the log

fabiencastan:


fabiencastan:
@ChemicalXandco What is the purpose of logManager.waitUntilCleared?
It locks the computation when run on command line with meshroom_photogrammetry or meshroom_compute.

ChemicalXandco:
The purpose was to avoid clearing the log file after it had been written to, resulting in part of the log missing. I have not tested it with the command line, so I didn't consider this. I shall look into a solution for this because without waiting until the log file is cleared, the first few messages that are written will be deleted.

PR #783, opened 2020-02-07
fix python nodes being blocked by log

ChemicalXandco:
Fix for #780, the log is cleared directly from `processChunk` instead of relying on chunk.statusChanged

### DIFF
--- a/meshroom/nodes/aliceVision/Publish.py
+++ b/meshroom/nodes/aliceVision/Publish.py
@@ -50,27 +50,29 @@ class Publish(desc.Node):
         return paths
 
     def processChunk(self, chunk):
-        chunk.logManager.waitUntilCleared()
-        chunk.logger.setLevel(chunk.logManager.textToLevel(chunk.node.verboseLevel.value))
-        
-        if not chunk.node.inputFiles:
-            chunk.logger.warning('Nothing to publish')
-            return
-        if not chunk.node.output.value:
-            return
+        try:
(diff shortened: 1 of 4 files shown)


setup.py
```diff
@@ -41,6 +41,8 @@ def __init__(self, script, initScript=None, base=None, targetName=None, icons=No
 build_exe_options = {
     # include dynamically loaded plugins
     "packages": ["meshroom.nodes", "meshroom.submitters"],
+    # cx_Freeze does not copy idnadata by default, so it needs to be added here
```
fabiencastan:
```
"includes": [
    "idna.idnadata",  # Dependency needed by SketchfabUpload node, but not detected by cx_Freeze
    ],
```

meshroom/nodes/aliceVision/Publish.py
```diff
@@ -50,27 +50,29 @@ def resolvedPaths(self, inputFiles, outDir):
...
-        chunk.logManager.waitUntilCleared()
-        chunk.logger.setLevel(chunk.logManager.textToLevel(chunk.node.verboseLevel.value))
-        
-        if not chunk.node.inputFiles:
-            chunk.logger.warning('Nothing to publish')
-            return
-        if not chunk.node.output.value:
-            return
+        try:
+            chunk.logManager.start(chunk.node.verboseLevel.value)
```
fabiencastan: Would be nice to use with instead of start/end:
```
with chunk.logManager:
    ...
```

PR #842, opened 2020-03-27
[WIP] Estimate computation time

ChemicalXandco:
## Features list

- ETA label is displayed while computing
- ETA is calculated on a chunk by chunk basis
- Some formulas are based on a benchmark to base the time on the performance of the CPU
- Said benchmark only runs once and only takes around 0.1 seconds to complete, so the performance impact is minimal

## Implementation remarks

I'm not quite sure how to make this work for submitters, and I have no means to test it with submitters.

I shall work on factoring in node settings and improving accuracy in general

### DIFF
--- a/meshroom/core/desc.py
+++ b/meshroom/core/desc.py
@@ -423,6 +423,9 @@ class Node(object):
         """
         pass
 
+    def getEstimatedTime(self, chunk, reconstruction):
+        return 1 # each chunk is estimated to take 1 second by default
+
     def stopProcess(self, chunk):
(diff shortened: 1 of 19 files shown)


fabiencastan:
Hi @ChemicalXandco ,
Thanks for the contribution. I have not yet started to review it, but it sounds interesting. I would just recommend to start from simple to more complex tasks.
So regarding ressources usages, IMHO the first thing to do would be to have a display of;
- computation time per node for all chunks (node only and cumulative)
- disk usage per node (node only and cumulative)

PR #844, opened 2020-03-28
[docker] update to qt 5.14 and split build of dependencies

fabiencastan:
## Description

Split Dockerfile into 2 parts:
- dependencies installation; Qt, Python, all build tools, etc
- build Meshroom plugins and launch cx_freeze to create the Meshroom executable

### DIFF
--- a/Dockerfile_deps
+++ b/Dockerfile_deps
@@ -1,0 +1,62 @@
+ARG AV_VERSION
+ARG CUDA_VERSION=9.0
+ARG OS_VERSION=7
+FROM alicevision/alicevision:${AV_VERSION}-centos${OS_VERSION}-cuda${CUDA_VERSION}
+LABEL maintainer="AliceVision Team alicevision-team@googlegroups.com"
+
+# Execute with nvidia docker (https://github.com/nvidia/nvidia-docker/wiki/Installation-(version-2.0))
+# docker run -it --runtime=nvidia meshroom
+
+ENV MESHROOM_DEV=/opt/Meshroom \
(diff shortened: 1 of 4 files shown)


Dockerfile_deps
```diff
@@ -0,0 +1,62 @@
...
+WORKDIR "${MESHROOM_DEV}"
+RUN source scl_source enable rh-python36 && pip install -r dev_requirements.txt -r requirements.txt
+
+COPY ./docker/qt-installer-noninteractive.qs "${MESHROOM_DEV}/docker/"
+
+# Install Qt (to build plugins)
+ENV QT_VERSION_A=5.14 \
+    QT_VERSION_B=5.14.1
+WORKDIR /tmp/qt
+RUN wget https://download.qt.io/archive/qt/${QT_VERSION_A}/${QT_VERSION_B}/qt-opensource-linux-x64-${QT_VERSION_B}.run && \
```
simogasp: Have u tried? Because they changed the policy to install the libraries and u need an account now

fabiencastan: Yes, I changed from qt-unified-linux-x64-online to qt-opensource-linux-x64 for that.

natowi: Still accessible: https://download.qt.io/archive/qt/5.14/5.14.1/

EwoutH:
@fabiencastan Could you update the manual with these changes? https://meshroom-manual.readthedocs.io/en/latest/install/docker/docker.html

A short section on how to compile a release build would also be very nice!

PR #867, opened 2020-04-25
Feature fileselection

### LINKED ISSUE #461: [Improvement] Add file picker to add images
It appears the only way to add photos (when not doing "Live Reconstruction") is to drag and drop. This unfortunately is impossible for those of us without a file manager installed (there is nowhere to drag files from).

It should be pretty simple to add a button to choose files or make double (or single) clicking the drop area open a file chooser, and it would make it a lot easier to use. I currently use live reconstruction to pick a folder and then stop the live reconstruction, but that seems like a very roundabout way to do it.

### LINKED ISSUE #825: Image selection, without drag and drop
I'm on Linux and don't have a file manager that has drag and drop, because that's a silly feature.

I would like to be able to select the files from within the application in a dialog.

How can dragging and dropping be the only way to get images into meshroom.

skinkie:
## Description
https://github.com/alicevision/meshroom/wiki/Good-first-contributions point that _Add file picker to add images_ would be a good start for development. This would Fix #461 and Fix #825 

## Features list

- [X] Feature click on Drop Images Files (and icon). Fix #461, Fix #825 
- [X] Feature menu entries to in addition support augmenting.

## Implementation remarks

I would like to put down in this pull request has been my first contribution, and first QML work, which I found challenging. The first commit updates the Python interface, which splits the drag and drop slot to a drag and drop, and a more generic list of URLs.

The second commit adds the ability to be able to click on both the icon as the drop label. I choose this approach to prevent issues when clicking on any image later, this allowing the most obvious place a user would be able to click first. I added the new signal in because the different object type QObject vs  QList<QUrl>. I used FileDialog from QtQuick.Dialogs to allow a multifile selection. I wonder if the other Platform.FileDialog's should be updated in the future.

The third commit adds two menu entries, in order to add images dragless. As programmer I would have liked to reuse the same FileDialog, I don't know if this would be an anti-pattern in QML. I found it very hard to just reuse the FileDialog from one of the files. I would love to hear if it this is possible at all.

### DIFF
--- a/meshroom/ui/qml/ImageGallery/ImageGallery.qml
+++ b/meshroom/ui/qml/ImageGallery/ImageGallery.qml
@@ -1,6 +1,7 @@
 import QtQuick 2.7
 import QtQuick.Controls 2.3
 import QtQuick.Layouts 1.3
+import QtQuick.Dialogs 1.1
 import MaterialIcons 2.2
 import QtQml.Models 2.2
 
@@ -26,12 +27,25 @@ Panel {
 
     signal removeImageRequest(var attribute)
     signal filesDropped(var drop, var augmentSfm)
+    signal filesAdded(var fileUrls, var augmentSfm)
 
     title: "Images"
     implicitWidth: (root.defaultCellSize + 2) * 2
 
     function changeCurrentIndex(newIndex) {
         _reconstruction.cameraInitIndex = newIndex
+    }
+
+    FileDialog {
+        id: addFileDialog
+        title: "Add Files"
+        nameFilters: ["Images (*.jpg *.jpeg *.tif *.tiff *.png *.exr *.rw2 *.cr2 *.nef *.arw)"]
+        selectMultiple: true
+        onAccepted: {
+            root.filesAdded(fileUrls, false)
+            closed(Platform.Dialog.Accepted)
+        }
+        onRejected: closed(Platform.Dialog.Rejected)
     }
 
     QtObject {
@@ -229,9 +243,19 @@ Panel {
                     text: MaterialIcons.photo_library
                     font.pointSize: 24
                     font.family: MaterialIcons.fontFamily
+
+                    MouseArea {
+                        anchors.fill: parent
+                        onClicked: { addFileDialog.open() }
+                    }
                 }
                 Label {
                     text: "Drop Image Files / Folders"
+
+                    MouseArea {
+                        anchors.fill: parent
+                        onClicked: { addFileDialog.open() }
+                    }
                 }
             }
 
--- a/meshroom/ui/qml/WorkspaceView.qml
+++ b/meshroom/ui/qml/WorkspaceView.qml
@@ -69,6 +69,7 @@ Item {
                 currentIndex: reconstruction.cameraInitIndex
                 onRemoveImageRequest: reconstruction.removeAttribute(attribute)
                 onFilesDropped: reconstruction.handleFilesDrop(drop, augmentSfm ? null : cameraInit)
+                onFilesAdded: reconstruction.handleFilesUrls(fileUrls, augmentSfm ? null : cameraInit)
             }
             LiveSfmView {
                 visible: settings_UILayout.showLiveReconstruction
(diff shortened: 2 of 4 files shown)


natowi:
Thank you for the contribution. (I attached a screenshot for other users to illustrate your commits)

screen

skinkie:
> Thank you for the contribution. (I attached a screenshot for other users to illustrate your commits)

You missed to highlight the extra menu separator ;)

> I noted one thing that is _not directly_ related to your commit:
> 
> When using the new HDRI pipeline there is no SfM node, so "Augment Reconstruction" is without use. Maybe we could disable the Augment Reconstruction drop field/File Menu as long as there is no SfM node in the node graph to improve usability. @fabiencastan what do you think?

When I start the application I get a full Photogrammetry pipeline. The augment menu entry is explicitly disabled, it is not common to _hide_ menu entries in applications. What I did find out is that the augment droparea is enabled when there is a single image (images > 0), while it should actually be at least two (images >= 2).

natowi:
"hide menu entries in applications." No, not hide, disable when using File->New pipeline->HDRI or when no SfM node is being used in the graph

skinkie:
> > hide menu entries in applications.
> 
> No, not hide, disable when using File->New pipeline->HDRI or when no SfM node is being used in the graph

I see what you mean now, but it should also disable the drop area :-) Maybe open a new ticket? This sounds like a "good first issue" as well.

natowi:
I have added this to the Good-first-contributions list for now.

skinkie:
@natowi it was easier than expected. What do you think about it? :-)

natowi:
@skinkie It works :-)

natowi:
About file types: 
It is good to enable the image formats with best support by default, but it would be better to have the ability to add *.* and the other https://github.com/alicevision/meshroom/wiki/Supported-image-formats. I also remembered that Meshroom supports .mp4 video files, so it would be good to add this to the file type menu.

skinkie:
I took the image fileformat list from the Python code. https://github.com/alicevision/meshroom/blob/develop/meshroom/multiview.py#L9

I have tried MP4 before, but it does not yet work.

And you would be surprised that adding folders via dragging only works if the folder actually contains images, it does not work recursive. In addition QML does not allow both a multiple file selection and a folder selection. https://doc.qt.io/qt-5/qml-qtquick-dialogs-filedialog.html#selectFolder-prop

natowi:
> I have tried MP4 before, but it does not yet work.

When you drag and drop a mp4 file in Meshroom a new "KeyframeSelection" node is being created.
(At the moment you need to manually add the output folder as input, but that is another issue https://github.com/alicevision/meshroom/wiki/Good-first-contributions#better-handling-of-video-files-keyframeselection)

> In addition QML does not allow both a multiple file selection and a folder selection

I see...

skinkie:
What if we create a new menu "Images" with:
 - Add Images...
 - Add Image Folder...
 - Add Images (augment)...
 - Add Image Folder (augment)...

Surely I want this to work recursively as well. But I think a good UI design should be made.

natowi:
> But I think a good UI design should be made

I agree. Maybe we can take a look on how this is solved in other applications. But now it is enough for today I think :)

ChemicalXandco:
I believe this feature was already added in [this commit](https://github.com/alicevision/meshroom/pull/808/commits/f9477eadfc9424248e3febd79f4a29c1df4e4100) on #808

skinkie:
> I believe this feature was already added in [this commit](https://github.com/alicevision/meshroom/pull/808/commits/f9477eadfc9424248e3febd79f4a29c1df4e4100) on #808
> 
> (Though I don't think it allows for augmentation currently)

Would have been nice if that commit or pull request actually marked the issues.

natowi:
@ChemicalXandco thank you for the hint, I did not see this, too. 
But I think this PR is more advanced than this https://github.com/alicevision/meshroom/pull/808/commits/f9477eadfc9424248e3febd79f4a29c1df4e4100

Issue #553, opened 2019-07-14
[FeatureRequest] DistanceRatio between photos


Baasje85:
In #514 I requested to report in strong correlation versus features that have not been reconstructed reconstruction. Since Distance Ratio (hamburger menu, advanced attributes) in the Feature Matching node, influences what candidates progress, it might be good to report on the solution space. I envision a view where a per photo descending list is presented based on distance ratio. The user should be able to include and exclude pairs.

Issue #514, opened 2019-06-16
[FeatureRequest] Group photos that have feature correlation, but are not reconstructed


Baasje85:
At this moment photos are marked not reconstructed. But the user is not made aware what is the missing link between reconstructed photos and photos that could not be reconstructed. By showing what photos do strongly correlate with the edges between the reconstructed areas could be found. Preferably the application would hint about such correlation.

Islands which do strongly correlate, but can not be matched to a single scene, might recover to an independent result, hence again showing what could be matched, and what is missing overlap.

hargrovecompany:
out of curiosity, is there a way (currently) to see a view of an image with matched features visible/highlighted?

natowi:
@hargrovecompany no, but this is a good idea

natowi:
Not exactly what you are looking for, but in a similar direction: With the next release there will be an option to [preview the FeatureExtraction](https://github.com/alicevision/meshroom/pull/539) node and display the matches from different algorithms

Issue #512, opened 2019-06-16
[FeatureRequest] Contribution of single camera on SfM (and visa versa)


Baasje85:
It would be good to see the influence of a single camera on the point cloud. A rendering I could think of is using a selection colour on the point cloud. The inverse: what cameras have contributed to a single point would also be interesting.

From GUI perspective: 
 - a single click on a camera, would highlight the part of the point cloud influenced
 - a single click on a point, would highlight the cameras contributing to the location

laurelkeys:
Hi, I'm also really interested in this featured.

While there isn't currently a way to see which points are related to which photos on the GUI, may I get some advice on how to get this information from the logs (e.g. which node's info should I be looking into)?

I've just recently started using Meshroom so I'm still a bit unfamiliar with the code structure, and reading throught the docs I couldn't find an answer to this problem, so searching for it was what actually led me to this issue.

I tried looking into the `.log` files from the `FeatureExtraction` to the `StructureFromMotion` nodes in the default pipeline but couldn't really get a grasp of them for the most part. To better understand what each of them does, is there a good place to start reading into?

Thank you.

laurelkeys:
@natowi  I'm using Meshroom to reconstruct images captured by multiple drones (currently simulated on [AirSim](https://github.com/microsoft/AirSim)), with the aim of optimizing their trajectory for maximum scene coverage, so I'm trying to identify which places in the cloud are less dense to capture more pictures in those areas.

Therefore, I'd like to be able to query which points in the reconstructed point cloud came from a specific image, and also which camera viewpoints have contributed to a single point.

I was looking into the `StructureFromMotion` output (converted to JSON) and, from what I understood, it seems like I can get most of this info from there, since it has a `"viewId"` for each image and inside of `"structure"` I can get the coordinates of each point with `"X"`, and which pictures were used with `"observations"`.

Is this the information that would be used to implement the requested feature on the GUI?

fabiencastan:
@laurelkeys We could setup a confcall if you are interested to contribute to the implementation of such a feature.

natowi:
> a single click on a camera, would highlight the part of the point cloud influenced
   a single click on a point, would highlight the cameras contributing to the location



[UAV Mapper](https://github.com/cvBin/UAV-Mapper) has this feature, but sadly only binaries are provided so far. 

[...]

PR #873, opened 2020-04-29
Add SfM Statistics

theoleplomb:
## Description

Add widgets to display SfM statistics.

## Features list

- [X] Display stats per view
- [X] Display global SfM stats
- [X] Display landmarks with reprojection errors
- [X] Display features associated to a track but not reconstructed as a 3D landmark

## Bugs to fix

 - [X] disable display View/Global stats button should close the window
 - [X] update of tracks/landmarks are not updated when empty

Based on https://github.com/alicevision/QtAliceVision/pull/5
globalstats
meshroom_display_features
reprojectionserrors
viewstatswithoutscale

### DIFF
--- a/meshroom/ui/qml/Charts/InteractiveChartView.qml
+++ b/meshroom/ui/qml/Charts/InteractiveChartView.qml
@@ -1,0 +1,57 @@
+import QtQuick 2.9
+import QtQuick.Controls 2.3
+import QtQuick.Layouts 1.3
+import MaterialIcons 2.2
+import QtPositioning 5.8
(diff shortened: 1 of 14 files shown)


natowi:
@theoleplomb Thank you for these nice features.

Here are also some feature requests that are related to your contribution and may be of interest:
https://github.com/alicevision/meshroom/issues/553
https://github.com/alicevision/meshroom/issues/514
https://github.com/alicevision/meshroom/issues/512
Additionally I think it would be interesting to select features in one image and highlight them in the other images

PR #874, opened 2020-04-29
Travis CI: Update environment, remove Python 2.7 & add 3.8

EwoutH:
- Updates the build environment from Ubuntu 16.04 Xenial to Ubuntu 18.04 Bionic
 - Removes Python 2.7 from the build matrix (since support has ended)
 - Adds Python 3.8 to the build matrix

PR #875, opened 2020-04-29
AppVeyor: Update build environment and save artifacts

EwoutH:
- Updates the build environment to the Visual Studio 2019 image
 - Use Python 3.8 in AppVeyor builds
 - Saves the build folder as an artifact

### DIFF
--- a/appveyor.yml
+++ b/appveyor.yml
@@ -4,11 +4,10 @@ environment:
 
 install:
   - "set PATH=%PYTHON%;%PYTHON%\\Scripts;%PATH%"
-  - cmake .
   - "pip install -r requirements.txt -r dev_requirements.txt --timeout 45"
+  - "python meshroom/__init__.py"
 
-build:
-  project: meshroom.sln
+build: off
 
 test_script:
   - "pytest tests/"
--- a/appveyor.yml
+++ b/appveyor.yml
@@ -1,7 +1,7 @@
 
 environment:
   matrix:
-    - PYTHON: "C:\\Python36-x64"
+    - PYTHON: "C:\\Python38-x64"
 
 install:
   - "set PATH=%PYTHON%;%PYTHON%\\Scripts;%PATH%"
--- a/appveyor.yml
+++ b/appveyor.yml
@@ -14,6 +14,7 @@ test_script:
 
 after_test:
   - "python setup.py build"
-  
+  - 7z a meshroomWin64snapshot%APPVEYOR_PULL_REQUEST_HEAD_COMMIT% ./build/*
+
 artifacts:
- - path: build
+ - path: build/*.zip
--- a/appveyor.yml
+++ b/appveyor.yml
@@ -4,6 +4,7 @@ environment:
 
 install:
   - "set PATH=%PYTHON%;%PYTHON%\\Scripts;%PATH%"
+  - git submodule update --init --recursive
   - "pip install -r requirements.txt -r dev_requirements.txt --timeout 45"
 
 build: off
--- a/appveyor.yml
+++ b/appveyor.yml
@@ -4,7 +4,7 @@ environment:
 
 install:
   - "set PATH=%PYTHON%;%PYTHON%\\Scripts;%PATH%"
-  - git submodule update --init --recursive
+  - cmake .
   - "pip install -r requirements.txt -r dev_requirements.txt --timeout 45"
 
 build: off
--- a/appveyor.yml
+++ b/appveyor.yml
@@ -7,7 +7,8 @@ install:
   - cmake .
   - "pip install -r requirements.txt -r dev_requirements.txt --timeout 45"
 
-build: off
+build:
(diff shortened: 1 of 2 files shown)


EwoutH:
Blocked by #874.

natowi:
- rename _build_ to meshroomWin64snapshot%APPVEYOR_PULL_REQUEST_HEAD_COMMIT%.zip
so we can better distinguish the builds.

EwoutH:
Thanks for the quick reply. It's not that quick, since _build_ refers to the build folders path. I implemented it with an dedicated zip command though. [Example here](https://ci.appveyor.com/project/AliceVision/meshroom/builds/32549288/artifacts).

What do you think of #874?

appveyor.yml
```diff
@@ -1,7 +1,7 @@
-
+image: Visual Studio 2019
 environment:
   matrix:
-    - PYTHON: "C:\\Python36-x64"
+    - PYTHON: "C:\\Python38-x64"
```
simogasp, line 4: is it interesting to test other python versions?
otherwise, we can remove `matrix`.
If we test multiple python versions we should also add the version to the name of the file generated as artifact

EwoutH: Good call! Travis currently tests 3.6, 3.7 and 3.8 so I think only 3.8 is enough in AppVeyor.

Removed the `matrix:` line.

EwoutH:
Can someone give me a hand how to build a standalone / release version of Meshroom? (so including aliceVision and the qtPlugins)

fabiencastan:
Hi @EwoutH,

Thanks for the PRs.
I'm not sure what is the best way to create standalone binaires.
You can have a look at the dockerfiles:
 - https://github.com/alicevision/meshroom/blob/develop/Dockerfile_deps
 - https://github.com/alicevision/meshroom/blob/develop/Dockerfile

The docker files are relying on the dockerhub images of AliceVision.
The problem is that it needs binaries of AliceVision aligned with the current branch....

natowi:
It is possible to load builds from other Appveyor sessions. I looked into this when I was updating the AliceVision Appveyor file.
https://help.appveyor.com/discussions/questions/19858-accessing-build-artifacts-from-one-project-in-another
I think it should be possible to merge existing builds of the components or write an Appveyor  script to clone all dependencies into the environment and put together a complete build.

Before we look into this, we should implement auto builds for https://github.com/alicevision/QtOIIO and https://github.com/alicevision/qmlAlembic

EwoutH:
The PRs referenced above enable saving build artifacts for QtOIIO and qmlAlembic. From both projects, the whole build folder is saved as .zip, and the following artifacts are also save separately:

- alembicEntityQmlPlugin.dll
- alembicEntityQmlPlugin.lib
- QtOIIOPlugin.dll
- QtOIIOPlugin.lib
- depthMapEntityQmlPlugin.dll
- depthMapEntityQmlPlugin.lib

Is anything else needed from those two projects?

natowi:
@EwoutH  Oh, yes, I missed to mention
https://github.com/alicevision/QtAliceVision
We could need an appveyor build for that, too.


> # Meshroom standalone structure
>
> - Meshroom/
>    - aliceVision/
>        - bin/    # runtime bundled binaries (windows: exe + libs, unix: executables)
>        - lib/    # runtime bundled libraries (unix: libs)
>        - share/  # resource files
>            - aliceVision/
>                - COPYING.md         # AliceVision COPYING file
>                - cameraSensors.db   # sensor database
>                - vlfeat_K80L3.tree  # [voctree file](https://gitlab.com/alicevision/trainedVocabularyTreeData/raw/master/vlfeat_K80L3.SIFT.tree)
>    - lib/      # Python lib folder
>    - qtPlugins/
>          - imageformats
>              - QtOIIOPlugin dll/lib
>          - qml
>              - AlembicEntity
>                   - alembicEntityQmlPlugin dll/lib
>              - AliceVision
>                   - qtAliceVisionPlugin dll/lib
>              - DepthMapEntity
>                   - depthMapEntityQmlPlugin dll/lib
>    Meshroom    # main executable
>    COPYING.md  # Meshroom COPYING file

natowi:
@EwoutH Can you check two things? When I use the latest Alicevision snapshot, DepthMap terminates with "cudart64_93.dll missing". When using the Meshroom (this) snapshot, the GUI crashes while running FeatureExtracion.

EwoutH:
I would suggest merging this now. I will open an separate PR for release builds, probably using GitHub Actions.

jure:
This appears to be a very minimal and non-invasive PR that greatly simplifies testing the latest dev builds. 👍

jure:
The added comment invalidated the review although no code changes occurred. The original review by @simogasp is probably equally applicable to this latest code.

jure:
Yay, works! 🎉 E.g. https://ci.appveyor.com/project/AliceVision/meshroom/builds/34184086/artifacts

PR #877, opened 2020-05-03
InstantMeshes

natowi:
From https://github.com/natowi/meshroom_external_plugins/issues/2#issuecomment-595740223
Some comment cleanup. InstantMeshes 0.4 code by https://github.com/djoerg

### DIFF
--- a/meshroom/nodes/aliceVision/InstantMeshes.py
+++ b/meshroom/nodes/aliceVision/InstantMeshes.py
@@ -1,0 +1,377 @@
+# Instant Meshes node plugin for Meshroom
+#
+# Meshroom plugin v.0.1-0.4 by Jörg Dittmer (https://github.com/djoerg) 02-03.2020,
+# based on the plugin by natowi (https://github.com/natowi) 11.2019
+# Meshroom plugin license: Mozilla Public License Version 2.0
+#
+# Credits
+# -------
+#
+# Pre-compiled binaries for Instant Meshes can be downloaded from https://github.com/wjakob/instant-meshes
+# Instant Meshes, created by  Wenzel Jakob, Marco Tarini, Daniele Panozzo, Olga Sorkine-Hornung
+#
+# This code uses the PyMeshFix wrapper from the open source project PyVista. 
+# See: https://www.pyvista.org/
+#
+#    Sullivan et al., (2019). PyVista: 3D plotting and mesh analysis through a streamlined interface for the Visualization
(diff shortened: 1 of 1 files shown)


NexTechAR-Scott:
Awesome plug in.

Great work.

Validated Win 10 tonight.

My pipe is Linux so will look to provide feedback on that tomorrow.

ChemicalXandco:
I am planning to implement plug-in installation (similar to blender) as part of #964 which will allow this to be distributed separately from meshroom which is much easier because then the dependencies and the instant meshes executable are not tied to the meshroom version. What do you think?

natowi:
Having an easy option to add external executables to be accessible for new nodes within MR would be great. It would simplify things.

fabiencastan:
We should start simple.
Creating new nodes using InstantMesh, Blender, etc. and just require the path to the binaries in these nodes.
When we have enough nodes that have proved to be useful, we can setup a UI module to deal with addons.

natowi:
@fabiencastan I agree. We don´t need a sophisticated plugin menu, but a simple option to define a path to an executable and being able to use the path variable in nodes would be a huge help.

fabiencastan:
Could we remove PyMeshFix and numpy for a first basic integration of InstantMesh?

natowi:
I guess so, but it really helps to remove self-intersections and non-manifolds from the Instant Meshes output. So yes, sure, we can remove it for a first implementation. I´ll take a look at it within the next days. If you are faster, feel free to go ahead.

fabiencastan:
Maybe we can create a dedicated node for PyMeshFix and evaluate if we integrate it separately.

natowi:
Hi @fabiencastan, I removed PyMeshFix and tested the node. Here are the known issues:

- InstantMeshes produces (more) self intersecting faces with smoothing, setting it to 0 worked most of the time
- The "Basic" unwrap method of the Texturing node seems more robust than the other two, "LSCM" definitly crashes on self-intersecting faces with "OpenNL assertion failed"! (Tested this by removing self-intersections with MeshLab, after that LSCM worked.)
- The unmodified output from Instant Meshes isn't compatible with Meshroom and can cause crashes or unexpected behaviour if feed directly into a node.
- Quad Meshes are not supported by other Meshroom nodes like Texturing.
- Quad Meshes can be loaded into the 3D Viewer, but show up  with a triangle icon  next to it. We should add a Quad icon in the future.

PR #909, opened 2020-05-22
Augmenting instructions re setting up qml extensions

julianrendell:
I tried both `;` and `:` as path separators for `QML2_IMPORT_PATH` and had no luck. Change above was the first method that got everything enabled.

### DIFF
--- a/INSTALL.md
+++ b/INSTALL.md
@@ -46,7 +46,7 @@ Additional Qt plugins can be built to extend Meshroom UI features. They can be f
 though they might get better integration in the future.
 Note that they are optional but highly recommended.
 
-For Windows, there is a bug reported in QT 5.14.2 that `QML2_IMPORT_PATH` does not work for some users when it contains multiple paths.
+For Windows, there is a [bug](https://bugreports.qt.io/browse/QTBUG-84390) reported in QT 5.14.2 that `QML2_IMPORT_PATH` does not work for some users when it contains multiple paths.
 
 The quickest work around for this issue is to copy the QML plugins to the following directory structure:
 
--- a/INSTALL.md
+++ b/INSTALL.md
@@ -46,7 +46,9 @@ Additional Qt plugins can be built to extend Meshroom UI features. They can be f
 though they might get better integration in the future.
 Note that they are optional but highly recommended.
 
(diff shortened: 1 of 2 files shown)


INSTALL.md
```diff
@@ -46,6 +46,24 @@ Additional Qt plugins can be built to extend Meshroom UI features. They can be f
 though they might get better integration in the future.
 Note that they are optional but highly recommended.
 
+For Windows, there may be problems setting `QML2_IMPORT_PATH` to contain multiple paths. The easiest approach may be to duplicate what is done for the binary release. This has a directory structure like this:
```
fabiencastan, line 49: I would like to avoid "may be" in the wording of the documentation. Documentation is there to provide information that people can trust.

From Qt documentation QML2_IMPORT_PATH is supposed to accept multiple paths.
https://doc.qt.io/qt-5/qtqml-syntax-imports.html
Multiple path entries must be separated by ':' under Unix and ';' under Windows. 

Could you double check? If it doesn't work, it would be good to report an issue on the Qt project.

julianrendell: I just tried: 

```
set QML2_IMPORT_PATH="D:\Meshroom_Dev\qmlAlembic\qml;D:\Meshroom_Dev\QtOIIO\qml;D:\Meshroom_Dev\QtAliceVision\qml"

python meshroom/ui 
```

and none of the plugins are available.

I've created https://bugreports.qt.io/browse/QTBUG-84390

julianrendell: I've reworded the pull request. It's purposely worded more passively as I'm don't regularly build large projects any more, and even less so on Windows. There's a very large chance I've broken something when I used vcpkg built qt :-)

julianrendell: And added a link to the QT bug.

julianrendell: After follow up on https://bugreports.qt.io/browse/QTBUG-84390, turns out quotes don't work in path specifications.

The things you learn.

Update to remove most of the changes, add a comment re quotes, and tweak the example path specs to add to, rather than replace.

natowi (approved):

natowi:
An alternative would be https://github.com/alicevision/meshroom/pull/1169

Issue #547, opened 2019-07-08
General enhancements ideas


jonbragado:
1. Streamline node interface: Nodes are fantastic but the interface on them is made for engineers. there are too many things visible there, almost every parameter is exposed with names that dont tell the user the correct usage about the effect of modifying that parameter. Tooltips in plain language would be very useful.
2. Navigation & Viewport: The viewport should behave like the rest of the 3d DCCA standard apps out there, with the same visual controls and keyboard keys.
3.- View through camera: Very useful to be albe to select a camera and see through it to see the geo matching that pictures lineup. We should also be albe to see the corresponding point overlaid on the image and the geo.

natowi:
1. This is why we have the "Advanced Settings" mode since 2019.1. (In the node attributes, (...) button)
With the next release, there will be an [improved Add Node](https://github.com/alicevision/meshroom/pull/248) menu. In the future it will provide details (and maybe an example image) for each node. In the meantime you can find details on the nodes in the manual (wip).
Here are some other improvements for the Graph view (nodes) [#269] 
2. From my experience each 3d software behaves slightly different. But it would be a good idea to 
- [ ] enable the user to define the navigation shortcuts in a preferences menu
3. Good idea.
- [ ] View through camera
- [ ] show images as 2d planes (small at camera origin, and proportional larger on the view path [#358]
4. 
- [ ] A General preferences menu would require some configurable preferences. This could be:
- [ ] Select path [#472](https://github.com/alicevision/meshroom/issues/472)
- [ ] configure keyboard shortcuts
- [ ] System resource usage
- [ ] render farm settings
Related:
- [ ] Project Manager (configure and store projects and settings) [#360](https://github.com/alicevision/meshroom/issues/360)
- [ ] Manage node graphs [#361]

MightyBOBcnc:
#2 (user-configurable navigation) would be especially welcome.  Trackball/gizmo navigation is, frankly, awful and is not how viewport navigation functions in professional modeling/texturing software. (e.g. Maya, 3DS Max, Mudbox, Substance Painter, Marmoset Toolbag, Quixel Suite, 3D-Coat, Blender, even Sketchfab viewer and so on.)

While, as you note, there are variations in the viewport navigation controls the general commonality between most 3D software is that:
[...]

PR #964, opened 2020-06-24
Preferences menu

ChemicalXandco:
## Description

The main feature here is the ability to override the default node attribute values, which is important because the defaults are not good enough in some use cases, and not having to change them every time can reduce the overall time taken to complete a reconstruction.
Some examples where overriding the default attribute value is useful:
- API token on Sketchfab upload
- Max points on meshing (to not run out of ram)

image

## Features list

- Preferences menu opened either from 'Edit' menu or F2 key
- Setting for default palette (dark/light)
- Setting to override the default node attribute values with user defined ones

## Implementation remarks

- The general section only has 1 setting currently, I'm sure more will be added in the future 
- `ListAttribute` and `GroupAttribute` cannot be overridden because supporting these would require some complexity, and I have no use for overriding the values of attributes that use these types


meshroom/ui/qml/GraphEditor/AttributeItemDelegate.qml
```diff
@@ -11,12 +11,14 @@ RowLayout {
...
 
     property variant attribute: null
+    property string nodeName: ""
+    property bool updatePreferences: false
     property bool readOnly: false // whether the attribute's value can be modified
 
     property alias label: parameterLabel  // accessor to the internal Label (attribute's name)
     property int labelWidth               // shortcut to set the fixed size of the Label
 
-    readonly property bool editable: !attribute.isOutput && !attribute.isLink && !readOnly
```
fabiencastan, line 19: "attribute" sounds easier to read than a generic "object".

meshroom/ui/qml/GraphEditor/AttributeItemDelegate.qml
```diff
@@ -118,32 +120,40 @@ RowLayout {
         }
     }
 
+    function setAttribute(value) {
```
fabiencastan, line 123: Would be better to emit a signal and react to the event outside depending on the context.

meshroom/core/graph.py
```diff
@@ -452,6 +453,10 @@ def addNewNode(self, nodeType, name=None, position=None, **kwargs):
         Returns:
              The newly created node.
         """
+        # Apply default attribute value overrides set in preferences 
+        for a in Preferences().getAttributeOverrides(nodeType):
+            kwargs[a._name] = a._value
```
fabiencastan, line 458: Would be better to deal with in the generic NodeDesc where the actual node default value is stored.
We should keep the original value in memory, but the override would be better in the same place.

meshroom/core/desc.py
```diff
@@ -168,6 +168,10 @@ def __init__(self, name, label, description, value, uid, group='allParams', adva
 
     def validateValue(self, value):
         try:
+            if value == 'true':
```
fabiencastan, line 171: Maybe `bool(distutils.util.strtobool(value))`?

meshroom/core/attribute.py
```diff
@@ -152,6 +153,8 @@ def uid(self, uidIndex=-1):
     def isLink(self):
         """ Whether the attribute is a link to another attribute. """
         # note: directly use self.node.graph._edges to avoid using the property that may become invalid at some point
+        if not self._node:
+            return False
```
fabiencastan, line 157: Maybe: `return self._node and self.node.graph and self.isInput and self in self.node.graph._edges.keys()`
I'm not sure about the mix between self._node and self.node.

meshroom/core/attribute.py
```diff
@@ -122,8 +122,9 @@ def resetValue(self):
         self._value = ""
 
     def requestGraphUpdate(self):
-        if self.node.graph:
-            self.node.graph.markNodesDirty(self.node)
+        if self._node:
```
fabiencastan, line 125: `if self._node and self.node.graph:`

fabiencastan: same I'm not sure why we would have both self._node and self.node

natowi:
Improvement ideas:

- make the preferences window movable and rescaleable.

- Add a Save&Exit / Cancel button to close the window

- the add node menu should be the same as in the graph editor, in the current version with a search bar, and in the future with https://github.com/alicevision/meshroom/pull/248

- Shorten the Tab name and add a small description field at the beginning of the tab content

- Add option to set "Default Cache Folder Path" in General settings

scs

This PR addresses some ideas from this issue https://github.com/alicevision/meshroom/issues/547

natowi:
Maybe in the long term, custom keyboard shortcuts could be set in the preferences,ttoo, as in some cases changes may be required for practical reasons https://github.com/alicevision/meshroom/pull/1227#issuecomment-821231764

Issue #937, opened 2020-06-11
Question about JPEG 2000 File Format


lal0cura:
Hi Everyone. I hope that you are OK
I'm having this issue with .JP2 files under Linux. I drag my files to the image pane but they are ignored. 

The reason that I use that format is because it gives visually lossless quality in very little space.
For example .PNG files uses 4.8 Mb for each file. 
In .JP2 files with the same quality uses only 1.6 Mb each. So it can save a lot of space.

I have 5 Sequences in PNG that uses about 18Gb, in JP2 it uses 4.6Gb with the same visual quality.

So I was wondering if Meshroom can in fact support .JP2 File Formats. (Jpeg2000).
According to the manual it should be  because JP2 are part of the OIIO Library.

So I'm not sure if it is a bug, or I'm the only one having this issue. I hope you could check it up to see if you can use that file format or also you cant use them.

I use Meshroom in Ubuntu 20.04.

Thanks.

PD. when I drag the .jp2 file to the image Pane I get this error in the console:

Program called with the following parameters:
 * allowSingleView = 1
 * defaultCameraModel = "" (default)
 * defaultFieldOfView = 45
 * defaultFocalLengthPix = -1 (default)
 * defaultIntrinsic = "" (default)
 * groupCameraFallback =  Unknown Type "20EGroupCameraFallback"
 * imageFolder = "" (default)
 * input = "" (default)
 * output = "/tmp/tmp2w3jkfgy/CameraInit/c448939571d5c70b05c9ae4ad416ada37d6273ca/cameraInit.sfm"
 * sensorDatabase = "/home/lal0cura/apps/Meshroom-2019.2.0/aliceVision/share/aliceVision/cameraSensors.db"
 * verboseLevel = "info"

[23:40:57.083444][error] Program need -i or --imageFolder option
No input images.
ERROR:root:Error while building intrinsics : Traceback (most recent call last):
  File "/opt/Meshroom/meshroom/ui/reconstruction.py", line 399, in buildIntrinsics
  File "/opt/Meshroom/meshroom/nodes/aliceVision/CameraInit.py", line 192, in buildIntrinsics
[...]

PR #965, opened 2020-06-24
[multiview] declare more recognized image file extensions

fabiencastan:


natowi:
Did you test the file extensions? One user recently reported, CameraInit would not load .jp2 files. https://github.com/alicevision/meshroom/issues/937 I tested a few other extensions and bmp or ico for example did not work, too.

dsmtE:
The extensions excepted are still hard-coded on the alicevision side.
I plan to integrate the various extensions available from openimageIO (https://github.com/alicevision/AliceVision/pull/823).

Issue #617, opened 2019-09-03
[request] automatic axis rotation (and origin) correction for final model


frank26080115:
I've been using Meshroom for a few outdoor statue/sculpture scans. Every time it finishes, the model needs to be rotated on many axis, it seems totally random. I always have to open up Blender to rotate and center the model. Most of the time I'm taking about 100-300 pictures, good ones, never had a can't-place failure.

**Describe the solution you'd like**
It would be very rare for me to not hold the camera in landscape or portrait orientation. I feel like it would be easy to calculate the "floor" plane out of all of the camera placements, and the direction vector can be used to calculate a rough center of my subject.

**Describe alternatives you've considered**
Using Blender is OK... but it is a pain. My computer is... lower mid tier in terms of power (6/12 cores and GTX 1660).

I've found other threads with people who might appreciate this feature https://groups.google.com/forum/?nomobile=true#!topic/alicevision/KbfmKP1whnM https://github.com/alicevision/meshroom/issues/256

paulparneix:
@frank26080115  you can use the SFMTransform node to apply a transform to the pointcloud & cameras.

The node isn't super user friendly and doesn't have a gizmo, but still workable. It also have some auto options based on cameras that kinda work.

Automatic orient & ground detection would be amazing for sure, but the problem is cameras placement depends a lot on the subject you are scanning & your shooting pattern.

There are some nice existing approaches to guess ground position in a point cloud, PCA is working quite well to estimate average plane, and then extract transform matrix from it.

It's also possible to guess subject world size using EXIF focus distance & focus point 2D position (if they exist). Surprisingly, for close to mid range subjects the result is not that off.

frank26080115:
@paulparneix 
It didn't work, something did rotate, the SFM preview don't match the textured mesh preview, but it's all upside-down still and tilted

Do I need to mess with the "landmarks describer types"? Should I enable all of them?

Cdddo:
Maybe it could be possible to choose a few "anchor" pics that we know have a good orientation and it can calculate the ground plane and rotations from there.

If not automatic at least it would be nice to have a simple gizmo where we can set the position and rotation of the pivot / origin.

natowi:
@Cdddo "Maybe it could be possible to choose a few "anchor" pics"
This feature is already available in the SfMTransform node.
Apply a given transformation camera as the origin of the coordinate system with the SfMTransform node. You can rescale the scene based on the bounding box of CCTAG markers.


[...]

Issue #616, opened 2019-09-02
[request] Key control for 3D viewer


daktylus:
**Is your feature request related to a problem? Please describe.**
When I scan larger scenes (e.g. rooms) the 3D Viewer get always unusable (e.g. pan movements get very tiny) or stuck at a certain zoom level (perhaps it is also correlated to the distance from the absolute zero point. ). It is really hard to navigate through the scene.

**Describe the solution you'd like**
Key controls for the 3D viewer. For example WASD for pan control. Ideally independent from zoom level and coupled to the view (for example pan step size could be a fixed ration of the gizmo diameter) 

**Describe alternatives you've considered**
No idea.

**Additional context**
Sometimes switching the view to orthographic would be nice to have.

PR #978, opened 2020-07-07
[Viewer3D] Input bounding box (Meshing) & manual transformation (SfMTransform) thanks to a new 3D Gizmo

Julien-Haudegond:
## Description

- Add a Transform Gizmo to control Qt3D entities inside the 3D Viewer (generic).
- Add an input Bounding Box inside the Meshing node to customize the area to reconstruct.
- Add a new transformation mode to SfMTransform using the Transform Gizmo.

Needs https://github.com/alicevision/AliceVision/pull/846

## Features list
- [X] 3D Gizmo
  - [X] Basic visual entity
  - [X] Picking interaction & object transformations
  - [X] Mouse handling orientation
  - [X] Display on top of everything
  - [X] Improve performances
  - [X] Attach the Gizmo directly inside an object (using a special container entity)
  - [X] Adjust manipulator sensitivity
- [X] Bounding Box
  - [X] New bounding box parameter in AliceVision
  - [X] New bounding box parameter in Meshroom's Meshing node
  - [X] Display BBox edges/faces
- [X] SfMTransform
  - [X] New transformation mode (manual) + new manualTransform parameter in AliceVision
  - [X] New transformation mode (manual) + new manualTransform parameter in Meshroom's SfMTransform node
- [X] Integrate Transform/BoundingBox input in the 3D Inspector

To do later: use BoundingBox struct instead of Hexah in AliceVision meshing process.

## Transform Gizmo

TransformGizmo_Complete_01

### Implementation remarks

The gizmo is entirely done with Qt3D in QML. To be able to get local transformations (and not global ones), we have to deal with matrix operations. Not all the methods/functions exist in QML, so we have to implement our own functions using PySide2 in Python.
The gizmo is made with meshes. Maybe it would be better to draw it on screen directly?

The controls are pretty similar to the Blender's gizmo ones. By the way, the visual is strongly inspired by the Blender 2.8+ gizmo.
It is quite user-friendly.

#### EntityWithGizmo

Inside Viewer3D.qml :

```js
EntityWithGizmo {
    id: entityWithGizmoID
    sceneCameraController: cameraController // The id of the current DefaultCameraController
    frontLayerComponent: drawOnFront // The id of the Layer we want to draw on top of everything
    window: root // The id of the current FocusScope Item (acting as a window)

    // The entity
    Entity { components: [entityWithGizmoID.objectTransform, ...] } // The Transform of the Entity needs to be this attribute
}
```

## Bounding Box

MeshingWithBBoxComparison

*On the left, Meshing with default option. On the right, Meshing with the custom Bounding Box.*

### Implementation remarks

The Bounding Box is a simple Qt3D entity representing a box in the style of Blender. The transformations of this box are stored in the Meshing node. When we use the gizmo inside the 3D Viewer to adjust the box, the changes are sent to the Meshing node when we release the mouse button. Like this, we ensure data is stored in Python and not in QML. We can activate the use of the custom Bounding Box in the node and then adjust the parameters. In the 3D Inspector, on the right, we can show/hide the Bounding Box.

## SfMTransform

To activate the gizmo, you must choose the manual mode inside *method* parameter. You can then adjust the parameters with the sliders and/or the gizmo.

### Implementation remarks

When *manual* mode is set, the point cloud you see in the 3D Viewer is actually the input with a real-time Transform applied to it. Even when the node is computed, this is the transformed input which is displayed.
Fix https://github.com/alicevision/meshroom/issues/994

### DIFF
--- a/meshroom/ui/qml/Viewer3D/MediaLibrary.qml
+++ b/meshroom/ui/qml/Viewer3D/MediaLibrary.qml
@@ -277,6 +277,10 @@ Entity {
 
                 onCurrentSourceChanged: {
                     updateCacheAndModel(false)
+
+                    // Avoid the bounding box to disappear when we move it after a mesh already computed
+                    if(instantiatedEntity.hasBoundingBox && !currentSource)
+                        model.visible = true
                 }
 
                 onFinalSourceChanged: {
--- a/meshroom/ui/qml/Viewer3D/MediaLibrary.qml
+++ b/meshroom/ui/qml/Viewer3D/MediaLibrary.qml
@@ -46,8 +46,8 @@ Entity {
             "valid": true,
             "label": "",
             "visible": true,
-            "hasTransform": false,
-            "displayTransform": true,
+            "hasBoundingBox": false, // for Meshing node only
+            "displayBoundingBox": true, // for Meshing node only
             "section": "",
             "attribute": null,
             "entity": null,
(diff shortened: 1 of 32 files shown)


natowi:
This will (partially) solve https://github.com/alicevision/meshroom/issues/616 & https://github.com/alicevision/meshroom/issues/617

meshroom/ui/qml/Viewer3D/MediaLibrary.qml
```diff
@@ -277,6 +277,10 @@ Entity {
 
                 onCurrentSourceChanged: {
                     updateCacheAndModel(false)
+
+                    // Avoid the bounding box to disappear when we move it after a mesh already computed
+                    if(instantiatedEntity.hasBoundingBox && !currentSource)
+                        model.visible = true
```
fabiencastan, line 283: Maybe easier to check if there is a bounding box in the only place where the model can be set to `visible=false`.

I have not checked in detail but could be there:
From:
```
                onFinalSourceChanged: {
                    // update media visibility
                    // (useful if media was explicitly unloaded or hidden but loaded back from cache)
                    model.visible = model.requested;
```
To
```
                    model.visible = model.requested || instantiatedEntity.hasBoundingBox;
```

meshroom/ui/qml/Viewer3D/MediaLibrary.qml
```diff
@@ -203,11 +203,14 @@ Entity {
...
-                    if(attribute && attribute.isLink)
-                        return attribute.linkParam.node.globalStatus === "SUCCESS";
-                    return true;
+                    if(!attribute)
+                        // if the node is removed, the attribute will be invalid
+                        return false
+                    if(attribute.isOutput)
+                        return attribute.node.globalStatus === "SUCCESS"
+                    if(attribute.isLink)
+                        return attribute.linkParam.node.globalStatus === "SUCCESS"
```
fabiencastan, line 212: linkParam is not recursive by default.
You should use: `getLinkParam(recursive=True)` (not yet exposed as a slot).
Maybe we should expose it as a property, like originLinkParam.

fabiencastan (approved):

PR #983, opened 2020-07-14
Import ABC from collections.abc

### LINKED ISSUE #982: [bug] Importing ABC directly from collections will be removed in Python 3.10
https://github.com/alicevision/meshroom/blob/0f2d8caf872a2b0214f6653217245540e3415a51/meshroom/core/desc.py#L233

https://github.com/alicevision/meshroom/blob/2381cde667d86dca13a539a3709865e8f68117d6/meshroom/core/attribute.py#L197

tirkarthi:
## Description

Fixes #982 . Import collections from collections.abc . Compatibility code is not added since Python 2 support has ended.


fabiencastan:
Thanks for the PR!

We still have to support Python-2 to be able to submit job on [Tractor renderfarm](https://renderman.pixar.com/tractor) which, unfortunately, still doesn't have a python-3 api.

tirkarthi:
I thought the repo was Python 3 only due to commit : https://github.com/alicevision/meshroom/commit/80d028eb388ae0f3d29a7238b3b0d83ce6e530d2 . I have added compatibility code to ensure the PR works with Python 2 and 3 .

fabiencastan:
Perfect, thanks!
Yeah, I was not in favor of removing python-2 from the CI. But, at some point, we will be able to finally drop python-2!

tirkarthi:
codacy reports that imports are not used but the file is for compatibility code. asserts are skipped but it was previous code and doesn't seem to be relevant to this PR.

tirkarthi:
Thanks @fabiencastan :)

fabiencastan:
Thank you for the fix!

PR #993, opened 2020-07-19
Infinite drag

ChemicalXandco:
## Description

This adds an 'infinite drag' feature to the image viewer, 3D viewer and graph editor similar to that of [Blender](https://www.blender.org/). This makes it much easier and quicker to drag the view around.

## Implementation remarks

When the 3D view is being rotated, the rate at which it rotates relative to the amount the mouse is moving slows and eventually stops the further it is dragged, this behaviour happened before but it was not as noticeable because the drags were limited by the size of the screen. The way the drags going off the widget are handled is the cursor is moved to the opposite side and an offset is applied in order to compensate, as there is no way to stop the drag AFAIK. An alternative implementation that would be easier to handle but maybe less visually responsive is to hide the cursor and keep it in the same place while it is being dragged.

### DIFF
--- a/meshroom/ui/qml/Viewer/HdrImageToolbar.qml
+++ b/meshroom/ui/qml/Viewer/HdrImageToolbar.qml
@@ -10,10 +10,11 @@ FloatingPane {
     padding: 5
     radius: 0
 
+    property real gainDefaultValue: 1
     property real gammaDefaultValue: 1
-    property real offsetDefaultValue: 0
-    property real gammaValue: gammaCtrl.value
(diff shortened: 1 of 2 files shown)


fabiencastan:
Hi @ChemicalXandco
I'm sorry I didn't test and check this PR. Do you plan to submit it differently or did you close it just because it was old (and without answer)?

ChemicalXandco:
@fabiencastan I plan to make another once #1227 is merged because I need to rework some things to make it compatible.

fabiencastan:
Ok, great!

PR #1002, opened 2020-07-26
Create start.sh

Mehehehehe82:
## Features list

- Added a start script

## Implementation remarks

Start scripts make your program (almost) idiot proof

### DIFF
--- a/start.sh
+++ b/start.sh
@@ -1,0 +1,3 @@
+#!/bin/sh
+export PYTHONPATH=$((dirname $0))
+python meshroom/ui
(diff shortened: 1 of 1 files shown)


start.sh
```diff
@@ -0,0 +1,3 @@
+#!/bin/sh
+export PYTHONPATH=$((dirname $0))
```
fabiencastan, line 2: Currently it gives me "syntax error: invalid arithmetic operator", the fix is:
```suggestion
export PYTHONPATH=$(dirname $0)
```

Another solution, which ensures absolute paths:
```suggestion
export PYTHONPATH="$(dirname "$(readlink -f "${BASH_SOURCE[0]}" )" )"
```

fabiencastan: @Mehehehehe82  Sorry for taking so much time to answer your simple PR.
What do you prefer between the 2 suggestions?

Mehehehehe82: Suggestion 2

PR #1007, opened 2020-07-29
[Nodes] Add a new 'enabled' parameter to node's attributes

Julien-Haudegond:
## Description

'enabled' is a new parameter for the node's attributes. It allows to enable/disable the attribute. If it is disabled, the attribute will not be displayed inside the UI and will not be sent to the command line.

This is quite useful when an attribute is used only if another is checked, for instance.

'enabled' can receive either a boolean or a lambda function.

⚠️ Disabled parameters will not be used anymore in the UID. So it will change the invalidation status on existing projects. The major versions of the modified nodes are therefore changed.

### DIFF
--- a/meshroom/core/attribute.py
+++ b/meshroom/core/attribute.py
@@ -98,7 +98,11 @@ class Attribute(BaseObject):
 
     def getEnabled(self):
         if isinstance(self.desc.enabled, types.FunctionType):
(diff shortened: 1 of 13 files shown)


fabiencastan:
As disabled params will not be used anymore in the UID, it will change all the invalidation, so you need to increment the major version of each modified node.

PR #1033, opened 2020-08-14
[ui] Fix graph axes naming for ram statistics

ChemicalXandco:
## Description

Before this change, the graph for the ram statistics used gigabytes as the unit, but [the actual value is in percent](https://github.com/alicevision/meshroom/blob/a24bfc14b0eb9a669c082b96950571c38cbdc128/meshroom/core/stats.py#L103).

## Features list

- ram graph now uses % instead of gigabytes
- refactoring of the id names of ValueAxis items to be more readable e.g. `valueAxisY3` -> `valueGpuY`

### DIFF
--- a/meshroom/ui/qml/GraphEditor/StatViewer.qml
+++ b/meshroom/ui/qml/GraphEditor/StatViewer.qml
@@ -164,7 +164,7 @@ Item {
         root.nbReads = categories[0].length-1
 
         for(var j = 0; j < nbCores; j++) {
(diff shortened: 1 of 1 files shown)


fabiencastan:
It gives me:
```
Qt Quick Layouts: Detected recursive rearrange. Aborting after two iterations.
```

fabiencastan:
Ram in percent is extracted here:
https://github.com/alicevision/meshroom/blob/058cb87d2a6e4361934bd15e1b2b09c610b18234/meshroom/core/stats.py#L103

fabiencastan (approved):

PR #1041, opened 2020-08-24
[ui] Camera Response Function display: fix instabilities on linux

fabiencastan:
## Description

Camera Response Function display crashes on linux with Qt 5.14.2.

## Features list

- [X] More checks on file content and catch exceptions.
- [X] Check nbColumns using a slot instead of using the property.
- [X] Use a workaround on the CSV loader to avoid a crash (do not set active
to False but use visibility instead).

### DIFF
--- a/meshroom/ui/components/csvData.py
+++ b/meshroom/ui/components/csvData.py
@@ -1,6 +1,6 @@
 from meshroom.common.qt import QObjectListModel
 
-from PySide2.QtCore import QObject, Slot, Signal, Property, Qt
+from PySide2.QtCore import QObject, Slot, Signal, Property
 from PySide2.QtCharts import QtCharts
 
 import csv
@@ -76,8 +76,8 @@ class CsvData(QObject):
             for elt in csvRows[1:]:
                 for idx, value in enumerate(elt):
                     dataList[idx].appendValue(value)
-        except:
-            logging.error("CsvData: Failed to load file: {}".format(self._filepath))
+        except Exception as e:
+            logging.error("CsvData: Failed to load file: {}\n{}".format(self._filepath, str(e)))
 
         return dataList
 
--- a/meshroom/ui/components/csvData.py
+++ b/meshroom/ui/components/csvData.py
@@ -1,10 +1,12 @@
 from meshroom.common.qt import QObjectListModel
 
-from PySide2.QtCore import QObject, Slot, Signal, Property
+from PySide2.QtCore import QObject, Slot, Signal, Property, Qt
 from PySide2.QtCharts import QtCharts
 
 import csv
 import os
+import logging
+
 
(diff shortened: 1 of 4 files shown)


meshroom/ui/qml/Viewer/CameraResponseGraph.qml
```diff
@@ -34,7 +35,8 @@ FloatingPane {
         onWheel: {}
     }
 
-    property bool crfReady: csvData.ready && csvData.nbColumns >= 4
+    // note: We need to use csvData.getNbColumns() slot instead of the csvData.nbColumns property to avoid a crash on linux.
+    property bool crfReady: csvData.ready && (csvData.getNbColumns() >= 4)
```
Julien-Haudegond, line 39: To avoid a binding warning (I got one during testing), you could check if *csvData* exists in first place, like this:
```suggestion
    property bool crfReady: csvData && csvData.ready && (csvData.getNbColumns() >= 4)
```

Everything else seems to work fine.

fabiencastan: csvData is a static qml item, I do not see how it can be invalid.
```
CsvData {
        id: csvData
```

Julien-Haudegond: I got this error:
```
TypeError: Cannot read property 'ready' of null
```

Julien-Haudegond (approved):

meshroom/ui/components/csvData.py
```diff
@@ -1,10 +1,12 @@
 from meshroom.common.qt import QObjectListModel
 
-from PySide2.QtCore import QObject, Slot, Signal, Property
+from PySide2.QtCore import QObject, Slot, Signal, Property, Qt
```
fabiencastan, line 3:
```suggestion
from PySide2.QtCore import QObject, Slot, Signal, Property
```

meshroom/ui/components/csvData.py
```diff
@@ -53,23 +61,23 @@ def read(self):
...
+            # Create the objects in dataList
+            # with the first line elements as objects' title
+            for elt in csvRows[0]:
+                dataList.append(CsvColumn(elt)) # , parent=self._data
+            # Populate the content attribute
+            for elt in csvRows[1:]:
+                for idx, value in enumerate(elt):
+                    dataList[idx].appendValue(value)
+        except:
+            logging.error("CsvData: Failed to load file: {}".format(self._filepath))
```
fabiencastan, line 80:
```suggestion
        except Exception as e:
            logging.error("CsvData: Failed to load file: {}\n{}".format(self._filepath, str(e)))
```

Julien-Haudegond (approved):

PR #1048, opened 2020-09-03
[nodes] Texturing: put downscale to 2 by default

fabiencastan:
Scale 1 could be better for objects, but is really too much for large environments.
It's better to have a default value which ensures that we can open the result.


natowi:
I would not mind this change, especially when we merge https://github.com/alicevision/meshroom/pull/964 that allows to set preferences.

servantf (approved):

PR #1120, opened 2020-10-25
[ui] Do not animate transformations between viewpoints

### LINKED ISSUE #1115: [bug] Image overlay + sync should transition the image as well
**Describe the bug**
When going from pose-to-pose the _pointcloud_ tweens between pose-to-pose. The next synced image is positioned statically so the visual appearance is rather strange, where the photo is placed first, while the pointcloud transitions on it.

https://www.youtube.com/watch?v=GoRZg6jZu6s

ChemicalXandco:
- perspective goes to the viewpoint instantly when 'Sync with image selection' is turned on and the selected image is changed
- fixes #1115

### DIFF
--- a/meshroom/ui/qml/Viewer3D/ViewpointCamera.qml
+++ b/meshroom/ui/qml/Viewer3D/ViewpointCamera.qml
@@ -28,16 +28,7 @@ Entity {
     }
 
     components: [
-        Transform {
-            id: transform
-
-            Behavior on rotation {
-                PropertyAnimation { duration: 200}
-            }
-            Behavior on translation {
-                Vector3dAnimation { duration: 200}
-            }
-        }
+        Transform { id: transform }
     ]
 
     StateGroup {


skinkie:
Thanks, better :)

HarikalarKutusu:
My comment is also related to this: https://github.com/alicevision/meshroom/issues/1439

The animation is very useful for the above-mentioned case. In a sequence of images when the camera makes large moves (understandable by animation), I can pinpoint a wrongly resolved image/camera and remove it from the set. As far as I can understand this will remove the animation as a whole. If this is the case, please consider making it optional via a toggle.

skinkie:
@HarikalarKutusu the problem was that the animation was kind of out-of-sync.

HarikalarKutusu:
> @HarikalarKutusu the problem was that the animation was kind of out-of-sync.

Sorry, @skinkie I watched the video and assumed from the title that the animations are going to paper-bin as a whole. So, this is just a fix for the out-of-sych problem...

skinkie:
> > @HarikalarKutusu the problem was that the animation was kind of out-of-sync.
> 
> Sorry, @skinkie I watched the video and assumed from the title that the animations are going to paper-bin as a whole. So, this is just a fix for the out-of-sych problem...

To be clear: if you look closely on the video you see that there is a "smooth transition" of the camera tracking the point cloud, that does not match the photo in the background already being replaced. That transition has been removed. In this pull request. You can still browse through the images. But the browsing just looks better.

PR #1159, opened 2020-11-20
Notify if new version is available

ChemicalXandco:
## Description

It is best practice to have the latest version of a software on one's device, this change will check for updates and let the user know if a newer version is available.

## Implementation remarks

Latest version is fetched by getting the github page and matching a regex, not the most efficient way but it works.

### DIFF
--- a/meshroom/ui/qml/main.qml
+++ b/meshroom/ui/qml/main.qml
@@ -49,6 +49,34 @@ ApplicationWindow {
     SystemPalette { id: activePalette }
     SystemPalette { id: disabledPalette; colorGroup: SystemPalette.Disabled }
 
+    MessageDialog {
+        id: updateDialog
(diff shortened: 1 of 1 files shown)


meshroom/ui/qml/main.qml
```diff
@@ -49,6 +49,34 @@ ApplicationWindow {
     SystemPalette { id: activePalette }
     SystemPalette { id: disabledPalette; colorGroup: SystemPalette.Disabled }
 
+    MessageDialog {
+        id: updateDialog
+        title: "Update Available"
```
fabiencastan, line 54: "Check Updates"

fabiencastan: Maybe better to put it in the "Help" entry, close to the "About Meshroom".

Issue #1179, opened 2020-12-11
[request] Integrate AprilTag library for detection of markers (tag16h5 family) 


elektrokokke:
Corresponding Meshroom issue for AliceVision feature [issue 949](https://github.com/alicevision/AliceVision/issues/949).

In the modules where it makes sense, add "tag16h5" to the available describer types, enabling the usage of AprilTag markers from the tag16h5 family.

natowi:
You don´t need to open an issue if you already opened a pull request

Issue #950, opened 2020-06-17
[question] Trouble using know camera positions


FSund:
**Describe the problem**
I have a set of images with accurately known camera poses. I'm trying to force these poses in Meshroom, but having some issues. I've followed the instructions from [this post](https://github.com/alicevision/meshroom/issues/788#issuecomment-587478786), but the StructureFromMotion node crashes when I try to run it using a modified ´cameras.sfm´.

In summary I do the following
 - Run the default pipeline to (including) the SfM node
 - Modify `cameras.sfm` by setting the folders to empty
```
"featuresFolders": [
        ""
																			 
    ],
    "matchesFolders": [
        ""
																		   
    ],
```
 - Set `"locked": "0"` to `"locked": "1"` for `data["intrinsics"]` and for all `data["poses"]`
 - Modify `"rotation"` and `"center"` for all poses
 - Remove the input connection for the SfM node and paste the path to the new sfm file
 - Check "Lock Scene Previously reconstructed" and "Force Lock of all Intrinsic Camera Parameters"

The dataset computes fine with default settings.

**Log**
When running the SfM node with default settings
```
Program called with the following parameters:
 * describerTypes = "sift"
 * extraInfoFolder = "C:/Users/filip.sund/Projects/photogrammetry/kvala/003-timelapse-bilder-forced-poses/MeshroomCache/StructureFromMotion/590bf9ab08f4064ff14bdcf696b5379b5559edf3"
 * featuresFolders =  = [C:/Users/filip.sund/Projects/photogrammetry/kvala/003-timelapse-bilder-forced-poses/MeshroomCache/FeatureExtraction/9731bf2ef8bb422a14a7d2e8aacb45889cbbaa9d]
 * initialPairA = ""
 * initialPairB = ""
 * input = "C:/Users/filip.sund/Projects/photogrammetry/kvala/003-timelapse-bilder-forced-poses/MeshroomCache/CameraInit/aa988f5ab6d71d2e2a7e8124b9e9b42b13a1ea5d/cameraInit.sfm"
 * interFileExtension = ".abc"
 * localBAGraphDistance = 1
 * localizerEstimator =  Unknown Type "enum aliceVision::robustEstimation::ERobustEstimator"
 * localizerEstimatorError = 0
 * localizerEstimatorMaxIterations = 4096
 * lockAllIntrinsics = 0
 * lockScenePreviouslyReconstructed = 0
[...]

PR #1180, opened 2020-12-11
Integration of AprilTag library according to issue #1179 and AliceVision pull request #950

elektrokokke:
According to Meshroom issue #1179 (https://github.com/alicevision/meshroom/issues/1179),

add the describer type "tag16h5" to the following modules:
- ConvertSfmFormat (e.g., to be able to export the 3D AprilTag positions in a human-readable format as .sfm,
or to see only the AprilTag marker positions in the 3D view via .abc)
- FeatureExtraction (to be able to detect AprilTag markers from the tag16h5 family)
- FeatureMatching (to be able to match AprilTag markers)
- SfmTransform (to be able to use AprilTag markers, e.g., for the auto_from_markers transform)
- StructureFromMotion (to be able to compute the 3D positions of AprilTag markers)

This goes together with the AliceVision pull request 950 (https://github.com/alicevision/AliceVision/pull/950).


elektrokokke:
Ouch, sorry, just pushed another commit to the fork, not realizing that this would go into this pull request. How do I undo this?

PR #1182, opened 2020-12-14
Graph Dependencies

fabiencastan:
Connection from an input attribute to another input attribute does not represent a real dependency between the nodes: the 2nd node does not need to have the 1st node computed.
This PR adds the option `dependenciesOnly` to many graph functions to choose if we want to manipulate the user graph or the dependency graph.

With this PR many actions like duplicate nodes, remove nodes, remove data, etc are applied only to the dependency graph.

### DIFF
--- a/meshroom/ui/qml/GraphEditor/Edge.qml
+++ b/meshroom/ui/qml/GraphEditor/Edge.qml
@@ -41,6 +41,8 @@ Shape {
         startY: root.startY
         fillColor: "transparent"
         strokeColor: "#3E3E3E"
+        strokeStyle: edge != undefined && ((edge.src != undefined && edge.src.isOutput) || edge.dst == undefined) ? ShapePath.SolidLine : ShapePath.DashLine
+        dashPattern: [4, 4]
         capStyle: ShapePath.RoundCap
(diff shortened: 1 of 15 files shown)


meshroom/ui/qml/GraphEditor/Edge.qml
```diff
@@ -41,6 +41,8 @@ Shape {
         startY: root.startY
         fillColor: "transparent"
         strokeColor: "#3E3E3E"
+        strokeStyle: edge != undefined && ((edge.src != undefined && edge.src.isOutput) || edge.dst == undefined) ? ShapePath.SolidLine : ShapePath.DashLine
+        dashPattern: [4, 4]
```
yann-lty, line 45: A dash pattern of [6,4] could more adapted to get a stronger feeling of a dashed line (VS a dotted line).
Also, avoid variation in dash pattern when the stroke is hovered by compensating strokeWidth:

```
ShapePath {
  ...
  // final visual width of this path (never below 1)
  readonly property real visualWidth: Math.max(strokeWidth, 1)
  ...
  dashPattern: [6/visualWidth, 4/visualWidth]
}
```

Issue #269, opened 2018-10-12
Improvements: Graph Editor


natowi:
For graph redesign:

- [ ] when hovering over a connector also highlight the start and endpoint on the nodes (and/or when hovering over the node option - at the moment the text is only highlighted blue)

![conne](https://user-images.githubusercontent.com/8138537/73079877-2943a580-3ec5-11ea-912d-88e058e592bb.jpg)

- [ ] click on connector: select connector; also highlight the start and endpoint on the nodes and make the connector detachable (to attach to other node) instead of needing to remove the connection first. Enable _Del_ button to remove a selected connector

- [ ] when creating a new connection, pan Graph Editor view when cursor is near the border. Otherwise you have to zoom out first. (similat to https://github.com/alicevision/meshroom/pull/993)

- [x] centre node option text for options with input and output. Keep Input only node options left aligned and Output only options right aligned.
This way it is easier to find the corresponding text to an output node.
(maybe also centre node name (StructureFromMotion) instead of left aligned)

![centre](https://user-images.githubusercontent.com/8138537/73080356-fbab2c00-3ec5-11ea-9417-a0aa8860a294.jpg)

- [ ] fix auto layout depth mode _minimum_ or remove it, as it is not helpful at the moment

In the future:

- [ ] colourize nodes based on categories used in https://github.com/alicevision/meshroom/pull/248
(could also be optional)
Sparse Reconstruction, Dense Reconstruction, Utils, Mesh post-processing
Similar to Blender:
![blend3](https://user-images.githubusercontent.com/8138537/73254684-2dbad780-41bf-11ea-8865-8274875257fa.JPG)

- [x] Insert: default pipeline / Draft Meshing pipeline

- [ ] add supported node: add "add supported node" to context menu when clicking on a node connector. Show the add node dialogue with node compatibility check enabled:
Distinct between supported _input_ nodes and supported _output_ nodes.

_Implemented:_

- [x] collapse and expand nodes to simplify the node graph and expand them if needed, auto connect supported inputs/outputs when collapsed

![col](https://user-images.githubusercontent.com/8138537/73084260-ef769d00-3ecc-11ea-878f-b7665d9834be.jpg)

![col1](https://user-images.githubusercontent.com/8138537/73254984-b174c400-41bf-11ea-9172-8f12eb08a4ac.JPG)

- [x] Improve node connector routing. Nodes should not overlay connectors
[...]

PR #1227, opened 2021-01-15
[ui] add support for selecting multiple nodes at once

ChemicalXandco:
## Description

The ability to select multiple nodes at once improves the efficiency of using the graph editor. I have tried to make the node selection behave similar to blender. The box select widget is used by dragging with the left mouse button. The main selected node is slightly brighter than the other selected nodes and has a blue border.
Screenshot_20210115_202417
Screenshot_20210115_193313

## Features list

- select multiple nodes at once
- use of control key modifier to select multiple nodes
- use of box select widget to select multiple nodes
- duplicate selected nodes
- remove selected nodes
- delete data of selected nodes
- move selected nodes together

(Resolves some points on #269)

### DIFF
--- a/meshroom/ui/commands.py
+++ b/meshroom/ui/commands.py
@@ -173,7 +173,18 @@ class RemoveNodeCommand(GraphCommand):
                                    self.graph.attribute(dstAttr))
 
 
-class DuplicateNodeCommand(GraphCommand):
+class _DuplicateNodes(GraphCommand):
+    def __init__(self, graph, parent=None):
+        super(_DuplicateNodes, self).__init__(graph, parent)
+        self.duplicates = []
+
+    def undoImpl(self):
+        # delete all the duplicated nodes
+        for nodeName in self.duplicates:
+            self.graph.removeNode(nodeName)
+
+
+class DuplicateNodeCommand(_DuplicateNodes):
     """
     Handle node duplication in a Graph.
     """
@@ -181,7 +192,6 @@ class DuplicateNodeCommand(GraphCommand):
         super(DuplicateNodeCommand, self).__init__(graph, parent)
         self.srcNodeName = srcNode.name
         self.duplicateFollowingNodes = duplicateFollowingNodes
-        self.duplicates = []
 
     def redoImpl(self):
         srcNode = self.graph.node(self.srcNodeName)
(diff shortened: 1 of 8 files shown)


fabiencastan:
I often get this error when manipulating nodes in the GraphEditor:
```
  File "C:\dev\meshroom\meshroom\common\qt.py", line 190, in remove
    raise ValueError("QObjectListModel.remove(obj) : obj not in list")
ValueError: QObjectListModel.remove(obj) : obj not in list
[2021-01-22 23:22:10,394][WARNING] file:///C:/dev/meshroom/meshroom/ui/qml/GraphEditor/GraphEditor.qml:446: Error: QObjectListModel.remove(obj) : obj not in list
```

meshroom/ui/commands.py
```diff
@@ -196,10 +206,21 @@ def redoImpl(self):
         self.duplicates = [n.name for n in duplicates]
         return duplicates
 
-    def undoImpl(self):
-        # delete all the duplicated nodes
-        for nodeName in self.duplicates:
-            self.graph.removeNode(nodeName)
+
+class DuplicateNodeListCommand(_DuplicateNodes):
```
fabiencastan, line 210: Don't need to have 2 classes to do it with one of multiple nodes and inheritance, etc.
Either we keep `DuplicateNodeListCommand` and always use it (even for a single node) or keep `DuplicateNodeCommand` and group the commands for multiple nodes.

fabiencastan (changes requested):
Thanks for the really cool PR!
I have added few notes to clean the code before integration.

fabiencastan:
Small behavior problem in this configuration (duplicate nodes and nodes behind):
image

It gives:
image

While the simple duplicate nodes:
image

Gives the correct result but with a strange selection:
image

fabiencastan:
It would be great the replace the "Alt+click" to select all the nodes after (instead of duplicating the nodes).
So if we want to move a node and all nodes connected behind it (a fairly common need), it would be simple.

fabiencastan:
Regarding the selection, I think we should clear the selection and select all the newly created nodes.

fabiencastan:
Really great PR, this is a life changer!
Looking forward to be able to merge it.

ChemicalXandco:
> Small behavior problem in this configuration (duplicate nodes and nodes behind)

I agree that this does not behave right. Should I remove the options '_ from here' or should I rework them to include the other selected nodes? I think that removing those options and relying on alt + click to select the following nodes is more intuitive than having these options that operate on some nodes that are **not** selected. What do you think @fabiencastan?

fabiencastan:
Good question. I agree that's redundant and makes two ways to do the same.
But, on the other hand, I think it's still useful to have the "from here" option as it's visual and do not require to know the keyboard shortcuts.
Also alt/shift/ctrl can be problematic in remote, or on some linux alt could be used by the os to move the window, etc.

natowi:
Love it :)

demo

Only a few minor layout issues https://github.com/alicevision/meshroom/issues/1389

fabiencastan:
@ChemicalXandco Is it ready for review? Or are you still working on it?

ChemicalXandco:
It is ready

fabiencastan (approved):

fabiencastan:
Great work!

PR #1271, opened 2021-02-12
Add mesh utilities nodes

gregoire-dl:
## Description

Add two nodes for mesh boolean operations and mesh conversion.

## Features list

- [x] Add MergeMeshes node
- [x] Add ConvertMesh node

See https://github.com/alicevision/AliceVision/pull/969

### DIFF
--- a/meshroom/nodes/aliceVision/ConvertMesh.py
+++ b/meshroom/nodes/aliceVision/ConvertMesh.py
@@ -1,0 +1,47 @@
+__version__ = "1.0"
+
+from meshroom.core import desc
+
+
+class ConvertMesh(desc.CommandLineNode):
+    commandLine = 'aliceVision_convertMesh {allParams}'
+    documentation = '''This node allows to convert a mesh to another format.'''
+
+    inputs = [
+        desc.File(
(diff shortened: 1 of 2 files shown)


natowi:
This part should be added to the MergeMeshes documentation: 

Operation types used to merge two meshes:
- boolean_union: Create a new mesh with the combined volume of the two input meshes.
- boolean_intersection: Create a new mesh from the intersected volumes of the two input meshes.
- boolean_difference: Create a new mesh from the volume of the first input mesh subtracted by the second input mesh.

fabiencastan (approved):

flobotics:
hi,
i tried it, but did not work ?  Also is there a way to output .obj also, now it seemed to output .stl only  ?

After which node should the mergeMeshes node be added,  after texturing?

The error in the log-view is like:

o-[DelaunyTetgen] Warning constraints have 4 edges on the border
Error: Encounterd a problem ,relaunch in diagnose mode
Error: Found 27 facets with intersections
o-[Booleans] Error could not tetrahedralize operand B

Then it saves an empty .stl file.

--I tried boolean-union option--

thx

Issue #1193, opened 2020-12-23
[request] Support HEIF and AVIF input


EwoutH:
Currently Meshroom 2020.1.1 doesn't supports importing HEIF and AVIF images. Since more and more cameras and RAW editors use these advanced file formats, support to import them directly into Meshroom would be great. The additional image convertion step to JPG currently reduces image quality, increases storage requirements and requires additional time.

[libheif](https://github.com/strukturag/libheif) could be used for importing both HEIF and AVIF images, while a separate [libavif](https://github.com/AOMediaCodec/libavif) is also available.

<img width="362" alt="Screenshot_308" src="https://user-images.githubusercontent.com/15776622/102980440-dc6ab380-4507-11eb-8bed-1e59b9ed72de.png"><img width="361" alt="Screenshot_309" src="https://user-images.githubusercontent.com/15776622/102980443-dd034a00-4507-11eb-953f-278562be1638.png">

fabiencastan:
Thanks for the feedback. It is already supported by oiio: https://github.com/OpenImageIO/oiio/blob/4dd4f2531141b7a365e9106350585d14405d55eb/src/heif.imageio/heifinput.cpp

So it is probably just a mater of build or just declaring the file extension as known.

fabiencastan:
Are you using Linux or Windows?
On Windows it is already built with HEIF support, see here: https://github.com/microsoft/vcpkg/blob/master/ports/openimageio/CONTROL#L6

So if you are on Windows, you could you try to add it on the Meshroom side here (and you can submit the change as a PR):
https://github.com/alicevision/meshroom/blob/develop/meshroom/multiview.py#L41
And it should work directly.

On Linux, we need to add it to the build:
https://github.com/alicevision/AliceVision/blob/develop/CMakeLists.txt#L569
PR on such topic would be more than welcome ;)

Best,

EwoutH:
I submitted a PR adding the format extensions to `multiview.py`: #1301

I looked into adding it in the the CMake setup for Linux, but unfortunately I didn't succeed. The above PR should at least enable HEIF and AVIF support on Windows.

thomas-fw:
Any news on this in the last four years? With most people capturing photography on an iphone it would be a very handy feature

PR #1301, opened 2021-02-25
meshroom/multiview.py: Add .heic, .heif and .avif as supported formats

EwoutH:
On most platforms Meshroom and Alicevision are already build with libheif enabled, this commit enables Meshroom to also accept the files as known image formats.

## Features list
- [x] Support HEIF and AVIF images on platforms that already build with libheif. Part of #1193.

## Implementation remarks
Could be a bit buggy on platforms that don't build with libheif by default.

### DIFF
--- a/meshroom/multiview.py
+++ b/meshroom/multiview.py
@@ -19,6 +19,8 @@ imageExtensions = (
     '.gif',
     # hdr:
     '.hdr', '.rgbe',
+    # heif
+    '.heic', '.heif', '.avif',
     # ico:
     '.ico',
     # iff:


meshroom/multiview.py
```diff
@@ -19,6 +19,8 @@
     '.gif',
     # hdr:
     '.hdr', '.rgbe',
+    # heif
+    '.heic', '.heif', '.avif'
```
fabiencastan, line 23: Missing comma:
```suggestion
    '.heic', '.heif', '.avif',
```

EwoutH: Thanks, reset, added the comma, and recommitted!

EwoutH:
Fixed the missing comma!

PR #1350, opened 2021-03-26
Context manager for python node logging

ChemicalXandco:
## Description

I have revised the implementation of node python logging to make it simpler and have built-in exception handling. These changes also contain other improvements to `SketchfabUpload` and `Publish` nodes.

Example of original:
```py
    try:
        chunk.logManager.start('debug')
        chunk.logManager.makeProgressBar(100, 'this is a progress bar')
        chunk.logManager.updateProgressBar(50)
        chunk.logger.debug('this is a debug log')
        chunk.logger.info('this is an info log')
        chunk.logger.warning('this is a warning log')
        chunk.logger.error('this is an error log')
        raise RuntimeError('something bad happened')
    except Exception as e:
        chunk.logger.critical(e)
        raise RuntimeError()
    finally:
        chunk.logManager.end()
```
Equivalent behaviour with context manager:
```py
    with desc.Logger(chunk) as logger:
        logger.makeProgressBar(100, 'this is a progress bar')
        logger.updateProgressBar(50)
        logger.debug('this is a debug log')
        logger.info('this is an info log')
        logger.warning('this is a warning log')
        logger.error('this is an error log')
        raise RuntimeError('something bad happened')
```
## Features list

- `SketchfabUpload`: use compression (file size can usually be at least halved from my testing)
- `Publish`: added option to enable/disable overwriting files (disabled by default)
- text file viewer highlights fatal and critical levels
- context manager used for python node logging

### DIFF
--- a/meshroom/nodes/aliceVision/SketchfabUpload.py
+++ b/meshroom/nodes/aliceVision/SketchfabUpload.py
@@ -1,6 +1,5 @@
 __version__ = "1.0"
 
-from meshroom.core import desc
 import glob
 import os
 import json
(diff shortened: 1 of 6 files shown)


meshroom/nodes/aliceVision/SketchfabUpload.py
```diff
@@ -184,27 +169,27 @@ class SketchfabUpload(desc.Node):
...
         try:
             r = requests.post(
                 modelEndpoint, **{'data': body, 'headers': headers})
-            chunk.logManager.completeProgressBar()
         except requests.exceptions.RequestException as e:
-            chunk.logger.error(u'An error occured: {}'.format(e))
-            raise RuntimeError() 
+            raise RuntimeError(u'An error occured: {}'.format(e))
+        finally:
+            os.remove(modelFile)
```
fabiencastan, line 207: The upoload function should not have the responsability of removing the input file.
The creation/removal of the temp file should be at the same level.

meshroom/nodes/aliceVision/SketchfabUpload.py
```diff
@@ -259,30 +234,27 @@ def processChunk(self, chunk):
...
-        except Exception as e:
-            chunk.logger.error(e)
-            raise RuntimeError()
-        finally:
-            if os.path.isfile(uploadFile):
-                os.remove(uploadFile)
-                chunk.logger.debug('Deleted {}'.format(uploadFile))
-
-            chunk.logManager.end()
+            self.upload(chunk.node.apiToken.value, uploadFile, data, logger)
```
fabiencastan, line 280: Here the temp file is created, so we should remove it at the end.

natowi:
@fabiencastan is this supposed to be closed?

PR #1355, opened 2021-03-28
Allow replacing edges

ChemicalXandco:
## Description

This makes using the graph editor more efficient because if I have node A connected to node C but then I want to connect node B to node C instead then I would have to first delete the edge from node A to node C before it was possible to drag the edge from node B to node C. With these changes it is possible to do this without having to explicitly delete the previous edge.

### DIFF
--- a/meshroom/ui/graph.py
+++ b/meshroom/ui/graph.py
@@ -535,8 +535,14 @@ class UIGraph(QObject):
         if isinstance(dst, ListAttribute) and not isinstance(src, ListAttribute):
             with self.groupedGraphModification("Insert and Add Edge on {}".format(dst.getFullName())):
                 self.appendAttribute(dst)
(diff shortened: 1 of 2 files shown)


fabiencastan:
Thanks @ChemicalXandco for the contribution. That's a good feature.

In term of implementation, I would like to avoid adding complexity to the low level commands.
The idea is to keep the basic commands as unitary as possible.

From the ui, just create a group of modifications: remove the edge and add a new one.
So you end up with 2 commands in one undo group.

What do you think?

fabiencastan (approved):

fabiencastan:
The PR looks good to me. I will merge it now, but it would be good to display in red the edge that will be removed, like here between MeshFiltering and Texturing:
image

ChemicalXandco:
> display in red the edge that will be removed

I agree, so I will make a PR.

PR #1357, opened 2021-03-29
Node highlight radius

ChemicalXandco:
This change increases the radius of the node highlight so that the inner radius of the highlight is the same as the node. This makes it so that there is no longer a gap between the highlight and the node at the corners. 

Before:
Screenshot_20210329_163149

After:
Screenshot_20210329_164455


fabiencastan (approved):

fabiencastan:
That was intentional to avoid too roundy stuff but that's fine too ;)

PR #1432, opened 2021-06-02
[nodes] New node "RenderAnimatedCamera" using blender API

DanielDelaporus:
## Description

This node uses Blender's API to Display the SFM Data at some specific points in the pipepline using the imported animated camera to show the result. 
Before going into details the user need to link the path to their blender.exe file. This is mandatory to use the software.

- First off, if you import your data as an alembic file. The node will be trying to display it as a Point Cloud using the particle system of Blender.
- Moreover, if you import your data as an obj file. The node will be trying to display it as a mesh. To be more exact, for visibility purposes, the edges of a mesh will be rendered.
- Importing the animated camera is required, it rebuilds the sequence in blender to see the render.
- Even if it's not mandatory, the node can display the background image corresponding to your data set. If you link the image folder of course.
- Finally the result of the render will be created at the path given in the node in any of the supported format of video.

## Features list

- The SFM Data extension changes drastically the way it's processed. It changes completely the render options so I added a system that displays the different render features dynamically using the SFM Data Path.
- If the Data is a Point Cloud, the user can change the size, the color and the density of the particle used to display in on blender.
- If the Data is a Mesh, the user can change the color of the edges of the display on blender.
- The feature to display the background image is not mandatory.
- Finally the sequence is render in 4 supported format. Gif are not included because that are not supported by blender. However we could imagine it being possible in the future using external converting tools.

- Something else that we could add it the possibility to add several render if the imported alembic containing the camera contains several of them.

## Implementation remarks

Most of the choices were made so the client could see and compare the result of meshroom to the original data set.
We tried to multiply features so that if the client has choices in their render method.

Every render was made using blender.


meshroom/nodes/blender/scripts/camera_support.py
```diff
@@ -0,0 +1,405 @@
+import bpy
```
fabiencastan, line 1: Rename the file to something like "renderAnimatedCameraInBlender.py"

meshroom/nodes/blender/RenderAnimatedCamera.py
```diff
@@ -0,0 +1,121 @@
+__version__ = "1.0"
+
+from meshroom.core import desc
+import os.path
+
+currentDir = os.path.dirname(os.path.abspath(__file__))
+
+class RenderAnimatedCamera(desc.CommandLineNode):
+    commandLine = '{blenderPathValue} -b --python {scriptPathValue} -- {allParams}'
+
```
fabiencastan, line 10: Add category and documentation variables.

meshroom/nodes/blender/scripts/camera_support.py
```diff
@@ -0,0 +1,405 @@
...
+
+    #import abc (Animated Camera)
+
+    try:
+
+        # In this part of the code we import the alembic in the cam_path to get the animated camera
+        # We use cam_location and cam_obj to store information about this camera
+        # We look for a camera (of type 'Persp') with the name 'anim' (not to confuse them with previously imported cams)
+
+        # Once the cam has been found we select is make it the main camera of the scene
```
fabiencastan, line 168:
```suggestion
        # Once the cam has been found we select the main camera of the scene.
```

meshroom/nodes/blender/scripts/camera_support.py
```diff
@@ -0,0 +1,405 @@
...
+        # The rest of the code is setting up the display of the background image,
+        # Since it's not a simple image but an image Sequence, we have to use the offset and the number of frame
+        # Information taken from the previous block of code.
+        # The frame method is the one that align with the Cloud of Point althought this may change
```
fabiencastan, line 172:
```suggestion
        # The frame method is the one that align with the Point Cloud althought this may change,
```

meshroom/nodes/blender/scripts/camera_support.py
```diff
@@ -0,0 +1,405 @@
...
+        # We are using a particle system later in the code to display the cloud point.
+        # To make it so, we need a model for the particle, a object that will be repeated a lot to make a shape.
+        # In order to do that we need a plane (one face only for optimisation purpose) that always face the camera.
+        # So we made a plane and made it a child (in the parenting system) of the camera. That way whenever the cam
+        # moves, the plane moves and turn accordingly.
+
+        # Bmesh creates the plane and put it into the mesh. We change the size of the plane according to
+        # the scale given in arguments. We need to adjust the plane's location because putting it at the
+        # exact location of the camera blocks the view. Then, the switcher gives a RGBA color according to
+        # the given argument... This is where it gets harder. We have to use a material that uses 'Emission'
```
fabiencastan, line 214:
```suggestion
        # the given argument. We have to use a material that uses 'Emission'
```

meshroom/nodes/blender/scripts/camera_support.py
```diff
@@ -0,0 +1,405 @@
...
+        # otherwise the particle is going to react to lights and we don't really need that (the color wouldn't be clear).
+        # To do that we have to use the shader 'node_tree' we clear all links between nodes, create the emission node
+        # and connects it to the 'Material Output' node (which is what we will see in render).
```
fabiencastan, line 217:
```suggestion
        # and connect it to the 'Material Output' node (which is what we will see in render).
```

meshroom/nodes/blender/scripts/camera_support.py
```diff
@@ -0,0 +1,405 @@
...
+        objectsPlane.active_material.node_tree.nodes.new(type='ShaderNodeEmission')
+        objectsPlane.active_material.node_tree.links.new(objectsPlane.active_material.node_tree.nodes['Emission'].outputs['Emission'], objectsPlane.active_material.node_tree.nodes['Material Output'].inputs['Surface'])
+        if (args.SFM_Data.endswith('.abc')):
+            objectsPlane.active_material.node_tree.nodes['Emission'].inputs[0].default_value = switcher.get(args.Particle_Color, 'Invalid Color')
+        
+    except RuntimeError:
+        print("Error: while setting up the particle model.")
+
+    if (args.SFM_Data.endswith('.abc')):
+        # This part is all about importing the Cloud Point and setting up the Particle System to make a good render
```
fabiencastan, line 247:
```suggestion
        # This part is all about importing the Point Cloud and setting up the Particle System.
```

meshroom/nodes/blender/scripts/camera_support.py
```diff
@@ -0,0 +1,405 @@
...
+        # After importing the alembic we look for a specific meshe in the file. Again the hard coded name would be a
```
fabiencastan, line 248:
```suggestion
        # After importing the alembic, we look for a specific mesh in the file. Again the hardcoded name would be a
```

meshroom/nodes/blender/scripts/camera_support.py
```diff
@@ -0,0 +1,405 @@
...
+        # problem if the previous nodes hadn't name it specificaly that (.001 because a meshe with the same name has
+        # been imported with the animated camera).
+
+        # Once the cloud point has been found. We make it the active object (important for the node_tree later).
+        # Then, we create a particle system on it. Render_type set to object and the said object is the plane,
+        # thanks to that the particle format is set to repeat the plane. Emit_from 'vert' so the points of the
+        # cloud of point are the one rendering the particle.
+        # The count is the number of particle repeated on the cloud of point. We use the rate given as arguments
+        # to give a number. Most of the following settings are just formalities but use_rotation and use_rotation_instance,
+        # those two make sure to use the same roation as the model which is vital to have the particle always facing the camera.
```
fabiencastan, line 258:
```suggestion
        # those two make sure to use the same rotation than the model (which is vital to have the particle always facing the camera).
```

PR #1479, opened 2021-07-17
Publish to basedir

gorenje:
## Description

This is a suggestion for a new publish node that creates a sub-directory for each compute of the Node. It also creates a `src.mg` file that is a copy of the current pipeline, i.e. the pipeline that generated the meshes and textures.

So for example, the file structure becomes:

```
/<basedir>/<project_name>/...
    src.mg
    <textures>
    <meshes>
/<basedir>/<project_name>_000/...
    src.mg
    <textures>
    <meshes>
/<basedir>/<project_name>_001/...
    src.mg
    <textures>
    <meshes>
```

This allows me to experiment with different settings without overwriting existing models.

## Implementation remarks

I made the assumption that this [line](https://github.com/gorenje/meshroom/blob/publish_to_basedir/meshroom/nodes/aliceVision/PublishToBaseDir.py#L75):

```
graph = chunk.node._attributes.parent().parent().parent()
```

will give me the top-level node, i.e. the project details. I don't know whether this is the correct
way to access the toplevel node ... 

A [second assumption](https://github.com/gorenje/meshroom/blob/publish_to_basedir/meshroom/nodes/aliceVision/PublishToBaseDir.py#L90-L96) I make in the node is the file format of `.mg` files. 
I found that over in the [Graph class](https://github.com/alicevision/meshroom/blob/18be350e6f7a410941d2d609722b9459469ea2a2/meshroom/core/graph.py#L1008-L1014).

### DIFF
--- a/meshroom/nodes/aliceVision/PublishToBaseDir.py
+++ b/meshroom/nodes/aliceVision/PublishToBaseDir.py
@@ -13,8 +13,11 @@ class PublishToBaseDir(desc.Node):
 
     category = 'Export'
     documentation = '''
-This node copies the results to the /mnt/models/<project name>_XXX directory. It also saves a copy
-of the current pipeline as src.mg.
+This node copies the results to a /<base dir>/<project name>_XXX directory. It copies all the input files.
+
+In addition, it saves a copy of the current pipeline as src.mg to that directory.
+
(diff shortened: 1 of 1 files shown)


fabiencastan:
@gorenje Thanks for the proposition. It could be confusing regarding the existing Publish node, this new name is not so explicit.
Maybe we could add various modes in the Publish node instead?

PR #1489, opened 2021-07-26
Draft Reconstruction pipeline

remmel:
## Description

In order to easier the use of Meshroom for user who don't have an NVIDIA CUDA card, I think that it is important to be able to start a project with the appropriate pipeline without customing it manually.
Related with wiki page https://github.com/alicevision/meshroom/wiki/Draft-Meshing

## Features list

Add File > New Pipeline > Photogrammetry (No CUDA)
image

## Remarks
Use the same parameters than others pipelines
An alternative would be to https://github.com/alicevision/AliceVision/issues/439 or to make it compatible with OpenMVS which doesn't need CUDA

### DIFF
--- a/meshroom/ui/qml/main.qml
+++ b/meshroom/ui/qml/main.qml
@@ -436,6 +436,10 @@ ApplicationWindow {
                     text: "Camera Tracking (experimental)"
                     onTriggered: ensureSaved(function() { _reconstruction.new("cameratracking") })
                 }
+                Action {
+                    text: "Photogrammetry Draft (No CUDA)"
+                    onTriggered: ensureSaved(function() { _reconstruction.new("photogrammetrydraft") })
+                }
             }
             Action {
                 id: openActionItem
--- a/meshroom/ui/reconstruction.py
+++ b/meshroom/ui/reconstruction.py
@@ -496,6 +496,9 @@ class Reconstruction(UIGraph):
         elif p.lower() == "cameratracking":
             # default camera tracking pipeline
             self.setGraph(multiview.cameraTracking())
+        elif p.lower() == "photogrammetrydraft":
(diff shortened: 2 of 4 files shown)


meshroom/ui/qml/main.qml
```diff
@@ -436,6 +436,10 @@ ApplicationWindow {
                     text: "Camera Tracking (experimental)"
                     onTriggered: ensureSaved(function() { _reconstruction.new("cameratracking") })
                 }
+                Action {
+                    text: "Photogrammetry (No CUDA)"
```
fabiencastan, line 440:
```suggestion
                    text: "Photogrammetry Draft (No CUDA)"
```

meshroom/ui/reconstruction.py
```diff
@@ -496,6 +496,9 @@ def new(self, pipeline=None):
         elif p.lower() == "cameratracking":
             # default camera tracking pipeline
             self.setGraph(multiview.cameraTracking())
+        elif p.lower() == "photogrammetrynocuda":
+            # default panorama fisheye hdr pipeline
```
fabiencastan, line 500:
```suggestion
            # draft mode for the photogrammetry pipeline that does not require cuda
```

meshroom/ui/reconstruction.py
```diff
@@ -496,6 +496,9 @@ def new(self, pipeline=None):
         elif p.lower() == "cameratracking":
             # default camera tracking pipeline
             self.setGraph(multiview.cameraTracking())
+        elif p.lower() == "photogrammetrynocuda":
```
fabiencastan, line 499: Would be better to name everything draft (instead of nocuda):
```suggestion
        elif p.lower() == "photogrammetrydraft":
```

fabiencastan (approved):

PR #1500, opened 2021-08-10
[ui] Add filters to image gallery

N0Ls:
## Description
New filter buttons at the bottom of the image gallery.
Allows to view filtered views of images.
Allows to view and edit intrinsics.

imagegallery

[python]
New python properties to get the name/label of an attribute with different level of precision/hierarchy.

## Features list

- [X] Filtered view for reconstructed cameras
- [X] Filtered view for non-reconstructed cameras
- [X] Placeholder when the view doesn't contain any filtered images
- [X] Intrinsics display as a tableview
- [X] Rename and expose new properties (label / name) inside the python

## Remarks
For now and because of version incompatibility, the headers for the intrinsics display are disabled and replaced with tooltips.


meshroom/ui/qml/ImageGallery/IntrinsicDisplayDelegate.qml
```diff
@@ -0,0 +1,198 @@
...
+        ToolTip.text: toolTipText
+
+        Rectangle {
+            width: parent.width
+            height: loaderComponent.height
+
+            color: rowIndex % 2 ? palette.window : Qt.darker(palette.window, 1.1)
+            border.width: 2
+            border.color: Qt.darker(palette.window, 1.2)
+
```
fabiencastan, line 40: TextInput seems to create out of its size...
```suggestion
clip: true
```

fabiencastan:
```
[WARNING] file:///C:/dev/meshroom/meshroom/ui/qml/ImageGallery/ImageGallery.qml:72: TypeError: Cannot read property 'count' of undefined
[WARNING] file:///C:/dev/meshroom/meshroom/ui/qml/ImageGallery/ImageGallery.qml:363:17: QML TableModel: appendRow(): expected a property named "Id" in row at index -1, but couldn't find one
[WARNING] file:///C:/dev/meshroom/meshroom/ui/qml/ImageGallery/IntrinsicDisplayDelegate.qml:119: TypeError: Cannot read property 'desc' of null
```

PR #1643, opened 2022-03-11
[nodes] calibration nodes

servantftechnicolor:
Add calibration nodes :

- 1 node for checkerboard detection
-  1 node for distortion estimation from checkerboards
- 1 node for intrinsics estimation from checkerboards

### DIFF
--- a/meshroom/nodes/aliceVision/RigCreate.py
+++ b/meshroom/nodes/aliceVision/RigCreate.py
@@ -1,0 +1,51 @@
+__version__ = "1.0"
+
+import json
+import os
+
+from meshroom.core import desc
+
+
+class RigCreate(desc.CommandLineNode):
+    commandLine = 'aliceVision_utils_rigCreate {allParams}'
+    category = 'Other'
+    documentation = '''
+Create a rig from multiple input sfm files.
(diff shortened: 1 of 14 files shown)


natowi:
It would be great to add support for the CameraInit node to use the calibration results as input :)

servantftechnicolor:
> Can the sfmData.sfm now be used as input for the CameraInit node? If not, that would be great :)

Currently we can use sfmtransfer to copy intrinsics to the output of camera init. However, it may be not user friendly?

natowi:
Using the input in the CameraInit node would be more intuitive I think: grafik

meshroom/nodes/aliceVision/RigCreate.py
```diff
@@ -0,0 +1,51 @@
...
+
+from meshroom.core import desc
+
+
+class RigCreate(desc.CommandLineNode):
+    commandLine = 'aliceVision_utils_rigCreate {allParams}'
+    category = 'Other'
+    documentation = '''
+Create a rig from multiple input sfm files.
+No modification is done on the data. Rig calibration is not performed here.
```
natowi, line 14: Better: Rig calibration can be performed with...

Issue #1647, opened 2022-03-16
Missing documentation and labeling on CameraInit Distortion Parameters


Phoenix-64:
I computed the brown distortion conditions k1 k2 k3 p1 p2 with an OpenCV script and want to use them in Meshroom but i dont know what parameter is corresponding to what.
If the Camera Type in the CameraInit is set to brown and I add 5 distortion parameters they are only labeled 0 trough 4 and not with the brown coefficients labels k1 k2 k3 p1 and p2 do I just have to enter them in this order or did I miss something completely?
Thanks for helping.

![grafik](https://user-images.githubusercontent.com/52720948/158689195-6af5bda1-2993-4e33-91fe-b76b54efef0f.png)

natowi:
Distortion Parameters:
0 - k1
1 - k2
2 - k3
3 - p1
4 - p2

Issue #1484, opened 2021-07-25
Transfering calibration data from OpenCV to Camera Init node


Phoenix-64:
I calibrated my camera with the OpenCV method [cv2.aruco.calibrateCameraCharucoExtended](https://docs.opencv.org/4.5.2/d9/d6a/group__aruco.html). This returns me the focal length and distortion parameters in the Brown-Conrady model. The specific parameters are k1, k2 and k3 for radial distortion and p1, p2 for tangential distortion.
The problem is that in the Camera Init node if the Camera Type is set to brown and I add 5 distortion parameters, they are not labelled. Wich are the tangential and which are the radial ones? And in what order are they? 
Transfering calibration data from OpenCV to Camera Init node
Here is a example of the parameters returned by OpenCV:
                 fx               fy                    cx                    cy                   k1  \
value   2983.609062  2983.609062  2008.786077  1129.857428  0.069759   
stddev     0.000000     2.959608     3.664702     4.215931  0.003851   

              k2           p1            p2             k3  
value   0.068102  0.001089  0.000459 -0.424215  
stddev  0.022468  0.000464  0.000493  0.040313 
 - OS: Windows 10
 - Meshroom version: 2921.1.0

natowi:
From https://github.com/alicevision/meshroom/issues/921 and by comparing to OpenMVG it should be as follows:
Distortion Parameters:
0 - k1
1 - k2
2 - k3
3 - p1
4 - p2

This information should be added to the tooltip

PR #1649, opened 2022-03-19
Distortion Parameter details

natowi:
## Description

Clarifying Distortion Parameter order

Related to
https://github.com/alicevision/meshroom/issues/1647
https://github.com/alicevision/meshroom/issues/1484

### DIFF
--- a/meshroom/nodes/aliceVision/CameraInit.py
+++ b/meshroom/nodes/aliceVision/CameraInit.py
@@ -80,7 +80,8 @@ Intrinsic = [
             description="Distortion Parameters",
         ),
     desc.BoolParam(name='locked', label='Locked',
-                   description='If the camera has been calibrated, the internal camera parameters (intrinsics) can be locked. It should improve robustness and speedup the reconstruction.',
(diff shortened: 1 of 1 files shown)


servantftechnicolor (changes requested):
This description depends on the camera model choosen. Definitly not a constant description.

natowi:
@servantftechnicolor In that case it would be great if you could provide more details

simogasp:
we need to put this info somewhere.
I had started to compile something like this
image
I need to complete it with the 3DE models but then we should put it in the Doxygen of each class and in the main doc

natowi:
The 3de parameters can be taken fom here: https://github.com/alicevision/AliceVision/blob/develop/src/aliceVision/camera/Pinhole3DE.hpp
https://github.com/alicevision/AliceVision/blob/develop/src/aliceVision/camera/Distortion3DE.hpp

I found documentation for the 3de models:
https://www.3dequalizer.com/user_daten/tech_docs/pdf/ldpk.pdf
https://www.3dequalizer.com/user_daten/tech_docs/pdf/distortion.pdf

PR #1728, opened 2022-07-14
[ui] Reduce confusion when qml loading fails

p12tic:
Currently we disable all logging out of qml by default. This is problematic in case qml loading fails for any reason (e.g. missing dependencies) as the user will be presented with a message that is not actionable: "QQmlApplicationEngine failed to load component". There is no way to understand what's happening unless the user knows about MESHROOM_OUTPUT_QML_WARNINGS.

This is improved by checking for presence of "QQmlApplicationEngine failed to load component" and warning if qml logging is disabled in that case.

There has been many issues with this log message and no further logs in the bug tracker, so I'm not the only one who hit this issue. Hopefully this improves debugging experience for everyone.

### DIFF
--- a/meshroom/ui/app.py
+++ b/meshroom/ui/app.py
@@ -51,9 +51,19 @@ class MessageHandler(object):
     @classmethod
     def handler(cls, messageType, context, message):
         """ Message handler remapping Qt logs to Python logging system. """
-        # discard blacklisted Qt messages related to QML when 'output qml warnings' is set to false
-        if not cls.outputQmlWarnings and any(w in message for w in cls.qmlWarningsBlacklist):
-            return
+
+        if not cls.outputQmlWarnings:
+            # If MESHROOM_OUTPUT_QML_WARNINGS is not set and an error in qml files happen we're
+            # left without any output except "QQmlApplicationEngine failed to load component".
+            # This is extremely hard to debug to someone who does not know about
+            # MESHROOM_OUTPUT_QML_WARNINGS beforehand because by default Qml will output errors to
+            # stdout.
+            if "QQmlApplicationEngine failed to load component" in message:
+                logging.warning("Set MESHROOM_OUTPUT_QML_WARNINGS=1 to get a detailed error message.")
+
+            # discard blacklisted Qt messages related to QML when 'output qml warnings' is not enabled
+            elif any(w in message for w in cls.qmlWarningsBlacklist):
+                return
         MessageHandler.logFunctions[messageType](message)
 
 


p12tic:
@fabiencastan Just a friendly ping :-)

meshroom/ui/app.py
```diff
@@ -50,6 +50,16 @@ class MessageHandler(object):
...
+        # extremely hard to debug to someone who does not know about MESHROOM_OUTPUT_QML_WARNINGS
+        # beforehand because by default Qml will output errors to stdout.
+        if "QQmlApplicationEngine failed to load component" in message and \
+                not cls.outputQmlWarnings:
+            logging.warning("Failed to load QML component, set MESHROOM_OUTPUT_QML_WARNINGS=1 "
+                            "to debug why")
+
         # discard blacklisted Qt messages related to QML when 'output qml warnings' is set to false
         if not cls.outputQmlWarnings and any(w in message for w in cls.qmlWarningsBlacklist):
             return
```
fabiencastan, line 65: Not tested:
```suggestion
        if not cls.outputQmlWarnings:
            # If MESHROOM_OUTPUT_QML_WARNINGS is not set and an error in qml files happen we're left
            # without any output except "QQmlApplicationEngine failed to load component". This is
            # extremely hard to debug to someone who does not know about MESHROOM_OUTPUT_QML_WARNINGS
            # beforehand because by default Qml will output errors to stdout.
            if "QQmlApplicationEngine failed to load component" in message:
                logging.warning("Set MESHROOM_OUTPUT_QML_WARNINGS=1 to get more detailed error message.")
            # discard blacklisted Qt messages related to QML when 'output qml warnings' is not enabled
            elif any(w in message for w in cls.qmlWarningsBlacklist):
                return
```

fabiencastan (changes requested):

meshroom/ui/app.py
fabiencastan: @p12tic Could you have a look to do this minor update before merge?

p12tic:
@fabiencastan friendly ping :-) I think this PR is good to go in.

meshroom/ui/app.py
```diff
@@ -51,9 +51,20 @@ class MessageHandler(object):
...
+            # If MESHROOM_OUTPUT_QML_WARNINGS is not set and an error in qml files happen we're
+            # left without any output except "QQmlApplicationEngine failed to load component".
+            # This is extremely hard to debug to someone who does not know about
+            # MESHROOM_OUTPUT_QML_WARNINGS beforehand because by default Qml will output errors to
+            # stdout.
+            if "QQmlApplicationEngine failed to load component" in message:
+                logging.warning("Failed to load QML component, set MESHROOM_OUTPUT_QML_WARNINGS=1 "
+                                "to debug why")
+
+            if any(w in message for w in cls.qmlWarningsBlacklist):
```
fabiencastan, line 66:
```suggestion
            # discard blacklisted Qt messages related to QML when 'output qml warnings' is not enabled
            elif any(w in message for w in cls.qmlWarningsBlacklist):
```

meshroom/ui/app.py
```diff
@@ -51,9 +51,20 @@ class MessageHandler(object):
...
-            return
+        if not cls.outputQmlWarnings:
+            # If MESHROOM_OUTPUT_QML_WARNINGS is not set and an error in qml files happen we're
+            # left without any output except "QQmlApplicationEngine failed to load component".
+            # This is extremely hard to debug to someone who does not know about
+            # MESHROOM_OUTPUT_QML_WARNINGS beforehand because by default Qml will output errors to
+            # stdout.
+            if "QQmlApplicationEngine failed to load component" in message:
+                logging.warning("Failed to load QML component, set MESHROOM_OUTPUT_QML_WARNINGS=1 "
+                                "to debug why")
```
fabiencastan, line 64:
```suggestion
                logging.warning("Set MESHROOM_OUTPUT_QML_WARNINGS=1 to get more detailed error message.")
```

meshroom/ui/app.py
fabiencastan: I have not tested it, but "elif" would make sense, isn't it?

p12tic: Agreed, this way full original message will be included.

meshroom/ui/app.py
p12tic: Addressed. Sorry for not noticing these changes in your original suggestion. I should just have applied it instead of rewriting manually.

fabiencastan (approved):

PR #1738, opened 2022-07-22
Fix node duplication/removal behaviour

cbentejac:
## Description

Fix corner cases that could lead to some errors when duplicating or removing nodes: 
- a node could be duplicated several times at once but only the first duplicate was stored, causing the "undo" operation to not remove any of the duplicates for that specific node beyond the first one; additionally, duplicated nodes could get stacked on top of other nodes, "hiding" them.
- there could be attempts to remove a single nodes several times at once, causing errors.

Change the ambiguous behaviour of the "duplicate node(s) from here" action that could lead to duplicating several times the same node unwillingly.

## Features list

- [X] Store correctly all the duplicated nodes: ensures that a duplication operation can be undone/redone properly, with all the duplicated nodes removed/re-duplicated.
- [X] Vertically align the duplicated nodes correctly: make sure there cannot be any duplicated node stacked on top of another.
- [X] Prevent the duplication/removal of a node more than once in the same action: ensures a node cannot be duplicated more. than once during a duplication operation, and fixes the cases where a same node was attempted to be removed several times during the same operation (thus raising errors).

### DIFF
--- a/meshroom/core/graph.py
+++ b/meshroom/core/graph.py
@@ -382,7 +382,7 @@ class Graph(BaseObject):
                 node, edges = self.copyNode(srcNode, withEdges=False)
                 duplicate = self.addNode(node)
                 duplicateEdges.update(edges)
-                duplicates[srcNode] = duplicate  # original node to duplicate map
+                duplicates.setdefault(srcNode, []).append(duplicate)
 
             # re-create edges taking into account what has been duplicated
             for attr, linkExpression in duplicateEdges.items():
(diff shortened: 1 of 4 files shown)


meshroom/core/graph.py
```diff
@@ -382,16 +382,21 @@ def duplicateNodes(self, srcNodes):
                 node, edges = self.copyNode(srcNode, withEdges=False)
                 duplicate = self.addNode(node)
                 duplicateEdges.update(edges)
-                duplicates[srcNode] = duplicate  # original node to duplicate map
+                if srcNode not in duplicates:
+                    duplicates[srcNode] = [duplicate]  # original node to duplicate map
+                else:
+                    duplicates[srcNode].append(duplicate)
```
fabiencastan, line 388: Not tested:
```suggestion
                duplicates.setdefault(srcNode, []).append(duplicate)
```

meshroom/core/graph.py
```diff
@@ -382,16 +382,21 @@ def duplicateNodes(self, srcNodes):
...
             for attr, linkExpression in duplicateEdges.items():
                 link = linkExpression[1:-1]  # remove starting '{' and trailing '}'
                 # get source node and attribute name
                 edgeSrcNodeName, edgeSrcAttrName = link.split(".", 1)
                 edgeSrcNode = self.node(edgeSrcNodeName)
-                # if the edge's source node has been duplicated, use the duplicate; otherwise use the original node
-                edgeSrcNode = duplicates.get(edgeSrcNode, edgeSrcNode)
+                # if the edge's source node has been duplicated (the key exists in the dictionary),
+                # use the duplicate; otherwise use the original node
+                if duplicates.get(edgeSrcNode) is not None:
```
fabiencastan, line 398:
```suggestion
                if edgeSrcNode in duplicates:
```

meshroom/core/graph.py
cbentejac: Tested (it works!) and pushed.

meshroom/core/graph.py
cbentejac: Done and pushed.

fabiencastan (approved):
Good job!

PR #1744, opened 2022-07-29
Add internal attributes in "Notes" tab

cbentejac:
## Description

This PR adds support for internal attributes, which are attributes that exist for all nodes independently from their type.
Internal attributes are displayed in the "Notes" tab of the graph editor (also added in this PR) and can be set for each node individually and independently. 

Four internal attributes are added with this PR: comments/invalid comments (to write some comments regarding a node), label (to rename a node), and color (to change a node's color).

Regarding the "comments" internal attribute, an icon is added to the node's header if a comment has been set, and hovering over that icon displays a tooltip containing the comment.

Internal attributes will be written to the template files if and only if they have non-default values (so they need to have been set). If no internal attribute has a non-default value, the "internalInputs" entry is not even written in the template file. The unit test checking on version/description conflicts in templates is updated accordingly. 

A new property, `uidIgnoreValue`, is added to the generic Attribute object. Its goal is to leave an attribute out of the node's UID computation if its value is set to the value it specifies. By default, it is set to `None` and any attribute that is considered for the UID computation will be taken into account. In the case of the internal attributes, `uidIgnoreValue=""` for the "invalidation" attribute: this prevents from invalidating all the existing graphs simply because this attribute has been added; instead, nodes will be invalidated because of that specific attribute if and only if its value is set to a non-empty string (when the invalidation is actually wanted). 

## Features list

- [X] Add support for internal attributes (attributes that will appear in the "Notes" tab)
- [X] Support multiline text field parameters and color selector text field parameters
- [X] Add a first set of internal parameters: "comments", "invalid comments", "label" and "color"
- [X] Add a "comment" icon in the header of nodes which have their "comment" attribute set, and a tooltip displaying that comment when the icon is hovered
- [X] Add a new attribute property, `uidIgnoreValue`, which allows to ignore an attribute during the node's UID computation depending on its value

## Implementation remarks

`uidIgnoreValue` is set to `None` by default for all types of attributes. It also is only exposed in the constructor of the `StringParam` attributes.

### DIFF
--- a/meshroom/core/attribute.py
+++ b/meshroom/core/attribute.py
@@ -143,6 +143,10 @@ class Attribute(BaseObject):
         self._enabled = v
         self.enabledChanged.emit()
 
+    def getUidIgnoreValue(self):
+        """ Value for which the attribute should be ignored during the UID computation. """
+        return self.attributeDesc.uidIgnoreValue
+
     def _get_value(self):
         if self.isLink:
             return self.getLinkParam().value
@@ -333,6 +337,7 @@ class Attribute(BaseObject):
     node = Property(BaseObject, node.fget, constant=True)
     enabledChanged = Signal()
     enabled = Property(bool, getEnabled, setEnabled, notify=enabledChanged)
+    uidIgnoreValue = Property(Variant, getUidIgnoreValue, constant=True)
 
 
 def raiseIfLink(func):
--- a/meshroom/core/attribute.py
+++ b/meshroom/core/attribute.py
@@ -168,6 +168,10 @@ class Attribute(BaseObject):
         # TODO: only update the graph if this attribute participates to a UID
         if self.isInput:
             self.requestGraphUpdate()
+            # TODO: only call update of the node if the attribute is internal
+            # Internal attributes are set as inputs
+            self.requestNodeUpdate()
+
         self.valueChanged.emit()
 
     def upgradeValue(self, exportedValue):
@@ -180,6 +184,12 @@ class Attribute(BaseObject):
         if self.node.graph:
             self.node.graph.markNodesDirty(self.node)
             self.node.graph.update()
+
+    def requestNodeUpdate(self):
+        # Update specific node information that do not affect the rest of the graph
+        # (like internal attributes)
+        if self.node:
+            self.node.updateInternalAttributes()
 
     @property
     def isOutput(self):
--- a/meshroom/core/desc.py
+++ b/meshroom/core/desc.py
@@ -14,7 +14,7 @@ class Attribute(BaseObject):
     """
     """
 
-    def __init__(self, name, label, description, value, advanced, semantic, uid, group, enabled):
+    def __init__(self, name, label, description, value, advanced, semantic, uid, group, enabled, uidIgnoreValue=None):
         super(Attribute, self).__init__()
         self._name = name
         self._label = label
@@ -25,6 +25,7 @@ class Attribute(BaseObject):
         self._advanced = advanced
         self._enabled = enabled
         self._semantic = semantic
+        self._uidIgnoreValue = uidIgnoreValue
 
     name = Property(str, lambda self: self._name, constant=True)
     label = Property(str, lambda self: self._label, constant=True)
@@ -35,6 +36,7 @@ class Attribute(BaseObject):
     advanced = Property(bool, lambda self: self._advanced, constant=True)
     enabled = Property(Variant, lambda self: self._enabled, constant=True)
     semantic = Property(str, lambda self: self._semantic, constant=True)
+    uidIgnoreValue = Property(Variant, lambda self: self._uidIgnoreValue, constant=True)
     type = Property(str, lambda self: self.__class__.__name__, constant=True)
 
     def validateValue(self, value):
@@ -201,8 +203,9 @@ class GroupAttribute(Attribute):
 class Param(Attribute):
     """
     """
-    def __init__(self, name, label, description, value, uid, group, advanced, semantic, enabled):
-        super(Param, self).__init__(name=name, label=label, description=description, value=value, uid=uid, group=group, advanced=advanced, semantic=semantic, enabled=enabled)
+    def __init__(self, name, label, description, value, uid, group, advanced, semantic, enabled, uidIgnoreValue=None):
+        super(Param, self).__init__(name=name, label=label, description=description, value=value, uid=uid, group=group, advanced=advanced, semantic=semantic, enabled=enabled,
+            uidIgnoreValue=uidIgnoreValue)
 
 
 class File(Attribute):
@@ -329,8 +332,9 @@ class ChoiceParam(Param):
 class StringParam(Param):
     """
     """
-    def __init__(self, name, label, description, value, uid, group='allParams', advanced=False, semantic='', enabled=True):
-        super(StringParam, self).__init__(name=name, label=label, description=description, value=value, uid=uid, group=group, advanced=advanced, semantic=semantic, enabled=enabled)
+    def __init__(self, name, label, description, value, uid, group='allParams', advanced=False, semantic='', enabled=True, uidIgnoreValue=None):
+        super(StringParam, self).__init__(name=name, label=label, description=description, value=value, uid=uid, group=group, advanced=advanced, semantic=semantic, enabled=enabled,
+            uidIgnoreValue=uidIgnoreValue)
 
     def validateValue(self, value):
         if not isinstance(value, str):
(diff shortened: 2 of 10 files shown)


meshroom/core/node.py
```diff
@@ -517,8 +518,13 @@ def getLabel(self):
         Returns:
             str: the high-level label of this node
         """
+        if self._internalAttributes.get("label").value.strip():
+            return self._internalAttributes.get("label").value.strip()
```
fabiencastan, line 522:
```suggestion
        label = self._internalAttributes.get("label").value.strip()
        if label:
            return label
```

meshroom/core/node.py
```diff
@@ -1085,19 +1110,24 @@ def __init__(self, nodeType, position=None, parent=None, **kwargs):
...
         self.setAttributeValues(kwargs)
+        self.setInternalAttributeValues(kwargs)
 
     def setAttributeValues(self, values):
         # initialize attribute values
         for k, v in values.items():
-            attr = self.attribute(k)
-            if attr.isInput:
-                attr.value = v
+            if self.hasAttribute(k):
```
fabiencastan, line 1127: I'm not sure if it's correct for setAttributeValues on a non-existing attribute to avoid the raise of an exception.

meshroom/ui/qml/GraphEditor/AttributeItemDelegate.qml
```diff
@@ -184,6 +191,121 @@ RowLayout {
...
+        }
+
+        Component {
+            id: color_textField_component
+            RowLayout {
+                CheckBox {
+                    id: color_textField_checkbox
+                    Layout.alignment: Qt.AlignLeft
+                    checked: node.color === "" ? false : true
+                    text: checked ? "Disable custom color" : "Enable custom color"
```
fabiencastan, line 252:
```suggestion
                    text: "Custom Color"
```
Avoid ambiguity in the name (the status or the action) and use Camel Case.

meshroom/core/node.py
cbentejac: Done.

meshroom/ui/qml/GraphEditor/AttributeItemDelegate.qml
cbentejac: Done.

meshroom/ui/qml/GraphEditor/AttributeItemDelegate.qml
```diff
@@ -152,6 +153,12 @@ RowLayout {
             case "BoolParam": return checkbox_component
             case "ListAttribute": return listAttribute_component
             case "GroupAttribute": return groupAttribute_component
+            case "StringParam":
+                if(attribute.desc.semantic === 'multiline')
+                    return textArea_component
+                if(attribute.desc.semantic === 'color/bool')
```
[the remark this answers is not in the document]
cbentejac: Done.

meshroom/core/node.py
cbentejac: Fixed.

fabiencastan (changes requested):
Need to deal with node upgrade of internal attributes.

cbentejac:
> Need to deal with node upgrade of internal attributes.

Done

meshroom/core/node.py
```diff
@@ -1334,7 +1352,7 @@ def attributeDescFromName(refAttributes, name, value, strict=True):
 
         return None
 
-    def _addAttribute(self, name, val, isOutput):
+    def _addAttribute(self, name, val, isOutput, internalAttr=False):
```
fabiencastan, line 1355: Need to declare the param in the documentation.

cbentejac: Done

fabiencastan:
- [ ] If we rename one internalAttribute, the upgrade is not working fine.
- [ ] Add an icon on nodes with comments (with the comment in tooltip)

mh0g:
On windows via the gui:
- colors tested, works
- invalidation tested, works
- renaming node tested, works
- 
Also tested with meshroom batch

Note: would it be possible to include the custom name (if any), in the status file created by each of the node?
As the folder name is not affected by the pr, this could be a convenient way to better differentiate two nodes of the same types.

mh0g (approved):

mh0g:
Does not break backward compatibility on projects from current develop .

fabiencastan (changes requested):

gregoire-dl (approved):
Rebased on develop and tested.
Works fine.

meshroom/core/node.py
```diff
@@ -1060,13 +1113,17 @@ def canBeCanceled(self):
 
     name = Property(str, getName, constant=True)
     label = Property(str, getLabel, constant=True)
+    color = Property(str, getColor, constant=True)
+    comment = Property(str, getComment, constant=True)
```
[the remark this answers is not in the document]
cbentejac: Done.

meshroom/ui/qml/GraphEditor/Node.qml
```diff
@@ -71,6 +71,13 @@ Item {
             root.x = root.node.x
             root.y = root.node.y
         }
+
+        onInternalAttributesChanged: {
+            nodeLabel.text = node ? node.label : ""
+            background.color = (node.color === "" ? Qt.lighter(activePalette.base, 1.4) : node.color)
+            nodeCommentTooltip.text = node ? node.comment : ""
```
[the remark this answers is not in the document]
cbentejac: Done.

meshroom/core/node.py
```diff
@@ -689,11 +689,16 @@ def toDict(self):
 
     def _computeUids(self):
         """ Compute node uids by combining associated attributes' uids. """
-        for uidIndex, associatedAttributes in self.attributesPerUid.items():
-            # uid is computed by hashing the sorted list of tuple (name, value) of all attributes impacting this uid
-            uidAttributes = [(a.getName(), a.uid(uidIndex)) for a in associatedAttributes if a.enabled]
-            uidAttributes.sort()
-            self._uids[uidIndex] = hashValue(uidAttributes)
+        conditionalUids = []  # uids that are taken into account in the final computation only if their attribute's value is not empty; their UID group is -1
```
fabiencastan, line 692: I don't see the reason to mix independant notions.
It's better to just add a property as it's done for the "enabled" property.

fabiencastan: Maybe add `uidIgnoreValue=None`.
So for a string we can set if to `uidIgnoreValue=""` to ignore an empty string.

cbentejac: Done. `uidIgnoreValue` is set to `None` by default and is only exposed in the StringParam constructor for now.

meshroom/core/attribute.py
```diff
@@ -143,6 +144,10 @@ def setEnabled(self, v):
         self._enabled = v
         self.enabledChanged.emit()
 
+    def getUidIgnoreValue(self):
+        """ Value for which the attribute should be ignored during the UID computation. """
+        return self._uidIgnoreValue
```
fabiencastan, line 149: `return self.attributeDesc.uidIgnoreValue`

No need to duplicate the value into the node. We can just use the value from the attribute description.

fabiencastan: Or maybe it's needed when we are in Compatibility mode?

meshroom/core/desc.py
```diff
@@ -485,6 +502,43 @@ class Node(object):
...
+            name="invalidation",
+            label="Invalidation Message",
+            description="A message that will invalidate the node's output folder.\n"
+                        "This is useful for development, we can invalidate the output of the node when we modify the code.\n"
+                        "It is displayed in bold font in the invalidation/comment messages tooltip.",
+            value="",
+            semantic="multiline",
+            uid=[0],
+            advanced=True,
+            uidIgnoreValue=""
```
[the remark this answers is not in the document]
cbentejac: Done.

meshroom/core/attribute.py
cbentejac: It doesn't seem to be needed in Compatibility Mode, so we can indeed avoid duplicating it in the node. Done!

meshroom/core/node.py
```diff
@@ -568,13 +600,40 @@ def attribute(self, name):
...
+        if "[" in name or "." in name:
+            p = self.attributeRE.findall(name)
+            return p[0][0] in self._attributes.keys() or p[0][1] in self._attributes.keys()
         return name in self._attributes.keys()
 
+    @Slot(str, result=bool)
+    def hasInternalAttribute(self, name):
+        if "[" in name or "." in name:
+            p = self.attributeRE.findall(name)
+            return p[0][0] in self._internalAttributes.keys() or p[0][1] in self._internalAttributes.keys()
```
fabiencastan, line 634:
```suggestion
```

fabiencastan (approved):

PR #1750, opened 2022-08-04
Add abstract InitNode

cbentejac:
## Description

This PR adds the abstract class "InitNode" which is meant to be inherited by all the initialization nodes (such as CameraInit), included those that might be created by the user. InitNode contains methods that can be reimplemented by the children classes if necessary.

This abstract class allows to keep on using scripts (such as meshroom_batch) without being limited to the use a specific initialization node (like it used to be with CameraInit in the case of meshroom_batch, for example), needing to have the knowledge of the used initialization node and/or having to modify the scripts as a workaround.

### DIFF
--- a/bin/meshroom_batch
+++ b/bin/meshroom_batch
@@ -9,7 +9,9 @@ meshroom.setupEnvironment()
 
 import meshroom.core.graph
 from meshroom import multiview
+from meshroom.core.desc import InitNode
 import logging
+
 
 parser = argparse.ArgumentParser(description='Launch the full photogrammetry or Panorama HDR pipeline.')
 parser.add_argument('-i', '--input', metavar='SFM/FOLDERS/IMAGES', type=str, nargs='*',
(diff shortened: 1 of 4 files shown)


bin/meshroom_batch
```diff
@@ -94,24 +96,30 @@ def getOnlyNodeOfType(g, nodeType):
...
 
+def getInitNode(g):
+    """
+    Helper function to get the Init node in the graph 'g' and raise an exception if there is no or
+    multiple candidates.
+    """
+    nodes = g.findInitNodes()
+    if len(nodes) != 1:
+        raise RuntimeError("meshroom_batch requires a pipeline graph with exactly one '{}' node, {} found."
+                           .format(InitNode, len(nodes)))
```
fabiencastan, line 107:
```suggestion
    if len(nodes) == 0:
        raise RuntimeError("meshroom_batch requires an Init node in the pipeline.")
    elif len(nodes) > 1:
        raise RuntimeError("meshroom_batch requires exactly one Init node in the pipeline, {} found: {}"
                           .format(len(nodes), str(nodes)))
```

cbentejac: Done

meshroom/nodes/aliceVision/CameraInit.py
```diff
@@ -250,6 +251,35 @@ class CameraInit(desc.CommandLineNode):
...
+        else:
+            filesByType.extend(findFilesByTypeInFolder(inputs, recursive=False))
+            searchedForImages = True
+
+        # If there was no input file, check that the directories do contain images
+        if searchedForImages and not filesByType.images:
+            raise ValueError("No valid input file or no image in the provided directories")
+        else:
+            views, intrinsics = self.buildIntrinsics(node, filesByType.images)
+            self.setAttributes(node, {"viewpoints": views, "intrinsics": intrinsics})
```
fabiencastan, line 281:
```suggestion

        views, intrinsics = self.buildIntrinsics(node, filesByType.images)
        self.setAttributes(node, {"viewpoints": views, "intrinsics": intrinsics})
```

cbentejac: Done

fabiencastan (approved):

PR #1754, opened 2022-08-17
Add dedicated "minimal" mode for templates

cbentejac:
## Description

Following #1727, project files can be used as templates, with the .mg files containing the entire description of the graph. This implies that if there is a single node whose description is modified (parameters added, parameters removed, etc.), description conflicts will arise when opening a template containing such a node. For these conflicts to be resolved, the template would need to be saved again with the upgraded version of the node every single time there is a change, which defeats the purpose of having templates.

This PR adds a "minimal" mode to save and load templates. Instead of saving the entire graph description, the minimal mode only saves the modified input/output parameters. When that template is opened, conflicts are detected only on the saved parameters (with non-default values) and automatically solved. This makes templates robust to changes in node descriptions.

## Features list

- [x] Support saving only modified input/output parameters of a graph in an .mg file
- [x] Support reading an .mg file containing only modified input/output parameters
- [x] Add a "Save As Template" option that saves a graph in minimal mode
- [x] Update all the existing templates with minimal mode
- [x] Reload the list of available templates after saving a graph as a template 

## Implementation remarks

- Files written in minimal mode can be identified with the flag "template: True" in the header. If that flag is not detected in an .mg file, it is loaded as a regular file and not a template.
- .mg files written normally (with all the parameters saved) can still be opened as templates, but they will not be robust to any node description changes.
- test_multiviewPipeline now tests both "minimal" and "regular" modes: the first part of the test opens the default Photogrammetry template (which has been updated to minimal mode) while the second part of the test saves the graph obtained after reading the template into a normal .mg file (in which all the input/output parameters are written).

### DIFF
--- a/meshroom/core/attribute.py
+++ b/meshroom/core/attribute.py
@@ -283,7 +283,7 @@ class Attribute(BaseObject):
         return copy.copy(self.desc.value)
 
     def _isDefault(self):
-        return self._value == self.defaultValue()
+        return self.value == self.defaultValue()
 
     def getPrimitiveValue(self, exportDefault=True):
         return self._value
--- a/meshroom/core/graph.py
+++ b/meshroom/core/graph.py
@@ -261,6 +261,9 @@ class Graph(BaseObject):
         self.header = fileData.get(Graph.IO.Keys.Header, {})
         nodesVersions = self.header.get(Graph.IO.Keys.NodesVersions, {})
 
+        # check whether the file was saved as a template in minimal mode
+        isTemplate = self.header.get("template", False)
+
         with GraphModification(self):
             # iterate over nodes sorted by suffix index in their names
             for nodeName, nodeData in sorted(graphData.items(), key=lambda x: self.getNodeIndexFromName(x[0])):
@@ -273,7 +276,8 @@ class Graph(BaseObject):
                 #   3. fallback to no version "0.0": retro-compatibility
                 if "version" not in nodeData:
                     nodeData["version"] = nodesVersions.get(nodeData["nodeType"], "0.0")
-                n = nodeFactory(nodeData, nodeName)
+
+                n = nodeFactory(nodeData, nodeName, template=isTemplate)
 
                 # Add node to the graph with raw attributes values
                 self._addNode(n, nodeName)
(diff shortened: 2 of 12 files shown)


meshroom/core/attribute.py
```diff
@@ -283,6 +283,8 @@ def defaultValue(self):
         return copy.copy(self.desc.value)
 
     def _isDefault(self):
+        if self.isLink:
+            return self.getLinkParam().value == self.defaultValue()
         return self._value == self.defaultValue()
```
fabiencastan, line 288:
```suggestion
        return self.value == self.defaultValue()
```
Not fully sure if it's evaluated properly, else:
`self._get_value() == self.defaultValue()`

meshroom/core/graph.py
```diff
@@ -1015,17 +1023,50 @@ def save(self, filepath=None, setupProjectFile=True):
...
+                Graph.IO.Keys.Graph: self.toDict()
+            }
 
         with open(path, 'w') as jsonFile:
             json.dump(data, jsonFile, indent=4)
 
         if path != self._filepath and setupProjectFile:
             self._setFilepath(path)
 
+    def getModifiedKeys(self):
```
fabiencastan, line 1045: getModifiedKeys(self) => getNonDefaultAttributes(self)

We are not returning only keys (myDict.keys(), myDict.values()).

meshroom/core/node.py
```diff
@@ -1465,3 +1465,78 @@ def nodeFactory(nodeDict, name=None):
             node = node.upgrade()
 
     return node
+
+def templateNodeFactory(nodeDict, name=None):
```
fabiencastan, line 1469: A bit too much duplication.
Maybe we can just add options to the nodeFactory function.
Using a functor for the filtering condition and an option to force nodes upgrade.

fabiencastan (changes requested):
.

meshroom/core/attribute.py
cbentejac: Done

meshroom/core/graph.py

meshroom/core/node.py
```diff
@@ -1463,5 +1464,7 @@ def nodeFactory(nodeDict, name=None):
         if not internalFolder and nodeDesc:
             logging.warning("No serialized output data: performing automatic upgrade on '{}'".format(name))
             node = node.upgrade()
+        if template:  # if the node comes from a template file and there is a conflict, it should be upgraded anyway
```
fabiencastan, line 1467: Maybe `elif` to ensure that we are not calling the upgrade twice in any case?

fabiencastan (approved):

PR #1758, opened 2022-08-19
Support copying/pasting a node in the Graph Editor

cbentejac:
## Description

This PR adds support for copying (Ctrl+C) and pasting (Ctrl+V) a node.
- Copy: if a node is currently selected, its content will be formatted to JSON (in the same manner as when a graph is saved to an .mg file) and copied to the system's clipboard. It can then be pasted anywhere.
- Paste: if the system's clipboard contains a correctly formatted node description, even if it does not contain any attribute, a node corresponding to that description will be created and added to the graph. If the clipboard does not contain a valid node description and there still is an attempt to paste its content in the Graph Editor, nothing will happen. The clipboard's content must be formatted to JSON to be considered valid. For example, `{
"nodeType": "CameraInit"
}` is a valid description of a default CameraInit node (with all its attributes set to their default values). 

## Features list

- [X] Format to JSON and copy a selected node to the system's clipboard when Ctrl+C is pressed in the Graph Editor
- [X] Parse the clipboard's content and, if it contains a valid and correctly formatted node description, create and add a node to the graph with that description when Ctrl+V is pressed in the Graph Editor

## Implementation remarks

- The clipboard must be cleared when exiting Meshroom because of an issue in the QClipboard object that occurs on Unix systems with X11/XCB. See https://bugreports.qt.io/browse/QTBUG-56456. 
- If the node description in the clipboard contains a position, it is ignored when pasting it into the Graph Editor. The node will be created and placed where the mouse is placed (the top left corner of the node will be placed right under the mouse).

### DIFF
--- a/meshroom/core/graph.py
+++ b/meshroom/core/graph.py
@@ -396,11 +396,11 @@ class Graph(BaseObject):
 
         return duplicates
 
-    def pasteNode(self, nodeType, **kwargs):
+    def pasteNode(self, nodeType, position, **kwargs):
         name = self._createUniqueNodeName(nodeType)
         node = None
         with GraphModification(self):
-            node = Node(nodeType, **kwargs)
+            node = Node(nodeType, position=position, **kwargs)
             self._addNode(node, name)
             self._applyExpr()
         return node
--- a/meshroom/core/graph.py
+++ b/meshroom/core/graph.py
@@ -396,6 +396,15 @@ class Graph(BaseObject):
 
         return duplicates
 
+    def pasteNode(self, nodeType, **kwargs):
+        name = self._createUniqueNodeName(nodeType)
+        node = None
+        with GraphModification(self):
+            node = Node(nodeType, **kwargs)
+            self._addNode(node, name)
+            self._applyExpr()
+        return node
+
     def outEdges(self, attribute):
         """ Return the list of edges starting from the given attribute """
         # type: (Attribute,) -> [Edge]
--- a/meshroom/ui/commands.py
+++ b/meshroom/ui/commands.py
@@ -199,14 +199,15 @@ class PasteNodeCommand(GraphCommand):
     """
     Handle node pasting in a Graph.
     """
-    def __init__(self, graph, nodeType, parent=None, **kwargs):
+    def __init__(self, graph, nodeType, position=None, parent=None, **kwargs):
         super(PasteNodeCommand, self).__init__(graph, parent)
         self.nodeType = nodeType
+        self.position = position
         self.nodeName = None
         self.kwargs = kwargs
 
     def redoImpl(self):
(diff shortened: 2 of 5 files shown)


fabiencastan:
It needs to work with multiple nodes selected.
We should add Copy/Paste entries in the Edit menu and in right click menu.

We should also maybe:
- add an "Import" option to import another Meshroom scene into the current one
- not reconnect to external nodes (but reconnect nodes between them)
- be able to modify the text in a text editor and copy it back into meshroom

cbentejac:
> It needs to work with multiple nodes selected. We should add Copy/Paste entries in the Edit menu and in right click menu.
> 
> We should also maybe:
> 
> * add an "Import" option to import another Meshroom scene into the current one
> * not reconnect to external nodes (but reconnect nodes between them)

Done:
- Copying/pasting selections containing multiple nodes is supported and the Copy/Paste entries have been added in the Edit and right click menus
- An "Import Scene" entry has been added to the File menu 
- Nodes are reconnected between themselves but not to external nodes

fabiencastan:
Some internal connections between the pasted nodes are lost during the copy/paste.

cbentejac:
> Some internal connections between the pasted nodes are lost during the copy/paste.

Fixed, sorry, I had added a "break" that shouldn't have been there in my previous push...

meshroom/ui/qml/GraphEditor/Node.qml
```diff
@@ -31,6 +31,8 @@ Item {
     readonly property color defaultColor: isCompatibilityNode ? "#444" : activePalette.base
     property color baseColor: defaultColor
 
+    property point mousePosition
```
fabiencastan, line 34: `property point mousePosition: Qt.point(mouseArea.mouseX, mouseArea.mouseY)`

meshroom/ui/qml/GraphEditor/Node.qml
```diff
@@ -108,6 +110,7 @@ Item {
         drag.threshold: 2
         hoverEnabled: true
         acceptedButtons: Qt.LeftButton | Qt.RightButton
+        onPositionChanged: root.mousePosition = mapToItem(mouseArea, mouseArea.mouseX, mouseArea.mouseY)
```
fabiencastan, line 113: Probably no need to map mouseArea to itself.

fabiencastan: In that case, better to keep declarative code (instead of imperative  code).

cbentejac: This line is actually not needed anymore since `property point mousePosition: Qt.point(mouseArea.mouseX, mouseArea.mouseY)` does the job by itself.

PR #1777, opened 2022-09-21
[core] Set the "template" flag to "false" when saving a project as a regular file

cbentejac:
## Description

When a scene is saved as a template, a flag named "template" is set to "true" in the header of the .mg file. 

Prior to this PR, and since #1754, that flag was never explicitly set to "false" when saving a scene as a regular project, meaning that any pipeline built on top of a template pipeline was necessarily saved with "template" set to "true".

### DIFF
--- a/meshroom/core/graph.py
+++ b/meshroom/core/graph.py
@@ -1200,9 +1200,10 @@ class Graph(BaseObject):
             for p in usedNodeTypes
         }
 
+        self.header["template"] = template
+
         data = {}
         if template:
-            self.header["template"] = True
             data = {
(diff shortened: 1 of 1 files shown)


meshroom/core/graph.py
```diff
@@ -1208,6 +1208,7 @@ def save(self, filepath=None, setupProjectFile=True, template=False):
                 Graph.IO.Keys.Graph: self.getNonDefaultAttributes()
             }
         else:
+            self.header["template"] = False
```
fabiencastan, line 1211: `self.header["template"] = template` out of the if/else

fabiencastan (approved):

PR #1784, opened 2022-09-26
Fix and prevent mismatches between an attribute's type and its default value's type

cbentejac:
## Description

This PR fixes a long-standing issue with nodes' attributes that was recently highlighted by #1747. 

Some nodes' attributes had default values that did not match with the attribute's type, and, provided these attributes could invalidate their node, this would trigger some UID changes when loading a graph with such nodes, compared to their initial values when the graph was written. 

One of the direct consequences of this issue was that a pipeline that was created and submitted to the render farm needed to be reloaded in Meshroom (to trigger the undesired change of UIDs) for the nodes' status to be updated; otherwise, the nodes' status remained in the "waiting" state, since these nodes' UIDs did not correspond to those submitted.

We address this issue as follows:
- All the attributes that had incorrect default values have been fixed for all the existing nodes in Meshroom.
- The unit tests now include a comparison between a graph's nodes' UIDs when said graph is saved, and when it is loaded.
- Upon Meshroom's start, when all the nodes are being loaded, every node's description is parsed to ensure that every attribute has a default value and range whose type matches with the attribute's. If no mismatch is detected, the node is loaded as usual. However, if a single attribute has an erroneous default value or range, then the node is rejected altogether: it will not be loaded into Meshroom, and a message will explicitly display which attribute(s) was invalid.

## Features list

- [X] Fix the type of default values to match with their attribute's type on all nodes
- [X] Check, before loading a node in Meshroom, that its description is valid: if there is a single attribute whose default value / range does not match the attribute's type, the node's description is deemed invalid, and that node will not be loaded
- [X] Add a check on the UIDs before writing a graph and after loading it in the unit tests


meshroom/core/__init__.py
```diff
@@ -94,6 +104,76 @@ def loadPlugins(folder, packageName, classType):
...
+
+    Returns:
+        errorList (list): the list of invalid parameters if there are any, empty list otherwise
+    """
+    errorList = []
+
+    def validateBasicTypes(param, errors):
+        # Validate base types (int, float, string, bool)
+        # ChoiceParam and File are not checked
+        if isinstance(param, desc.IntParam) and (not isinstance(param.value, int)
```
fabiencastan, line 125: `param.validateValueType(param.value)`
It should be part of each XXParam class. We should not maintain lists of all known types like this.

cbentejac: Done

fabiencastan (commented):
code LGTM now

meshroom/core/desc.py
```diff
@@ -133,6 +145,24 @@ def validateValue(self, value):
...
+        the group with invalid types.
+        """
+        invalidParams = []
+        for attr in self.groupDesc:
+            name = attr.validateValueType()
+            if name:
+                invalidParams.append(name)
+        if invalidParams:
+            # In group "group", if parameters "x" and "y" (with "y" in nested group "subgroup") are invalid, the
+            # returned string will be: "group:x, group:subgroup:y"
```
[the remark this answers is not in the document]
cbentejac: I'm not sure I understand this comment. Does it address only the comment itself, or the way nested attributes are displayed in the string?

If it is the former, then no, the comment is correct. If we take for example the "x" attribute of "principalPoint" (which is itself in the "intrinsic" group) in the CameraInit node, it would be displayed as "intrinsic:principalPoint:x" (as opposed to, let's say, "focalLength", which is directly in "intrinsic" and which would be displayed as "intrinsic:focalLength").

If it's the latter, I can update it but I think it might be easier to fix the erroneous attributes if the precise "position" of the attributes is given.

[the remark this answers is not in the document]
fabiencastan: just misread

meshroom/core/desc.py
```diff
@@ -185,6 +215,10 @@ def validateValue(self, value):
             raise ValueError('File only supports string input  (param:{}, value:{}, type:{})'.format(self.name, value, type(value)))
         return os.path.normpath(value).replace('\\', '/') if value else ''
 
+    def validateValueType(self):
+        # nothing to validate
```
fabiencastan, line 219: We should check that's a string.

meshroom/core/desc.py
```diff
@@ -44,6 +44,15 @@ def validateValue(self, value):
         """
         raise NotImplementedError("Attribute.validateValue is an abstract function that should be implemented in the derived class.")
 
+    def validateValueType(self):
```
fabiencastan, line 47: In fact, the function name is very similar to `def validateValue(self, value)` which does something very different. It would be good to find another name to avoid the ambiguity between the two.

Maybe `def checkValueTypes(self)`.

cbentejac: Done using "checkValueTypes" (I didn't have a better name idea in store 🙃).

fabiencastan (approved):

PR #1787, opened 2022-09-30
[ui] Elapsed time indicators in log

mugulmd:
## Description

The goal of this PR is to enrich the node log viewer so that finding the steps that require long calculations is easier for users. To achieve that, we retrieve the time value of each log-line (when it exists) and we display the elapsed time between consecutive time values.

We add 2 visual indicators: 
- a colored marker next to the line number
- a tooltip when hovering over the line number cell displaying the elapsed time.

### DIFF
--- a/meshroom/ui/qml/Controls/TextFileViewer.qml
+++ b/meshroom/ui/qml/Controls/TextFileViewer.qml
@@ -197,7 +197,7 @@ Item {
                     width: textView.width
                     spacing: 6
 
-                    property var duration: textView.model.get(index).duration
+                    property var logLine: textView.model.get(index) ? textView.model.get(index) : {"line": "", "duration": -1}
 
                     Item {
                         Layout.minimumWidth: childrenRect.width
@@ -208,7 +208,7 @@ Item {
                             Rectangle {
                                 width: 4
                                 Layout.fillHeight: true
-                                color: Colors.interpolate(Colors.grey, Colors.red, duration/180)
+                                color: Colors.interpolate(Colors.grey, Colors.red, logLine.duration/180)
                             }
                             // Line number
                             Label {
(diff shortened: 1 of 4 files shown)


mugulmd:
For now there are still 2 main issues to work on that I noted: 
- the color scale for the markers is arbitrary and definitely requires some fine-tuning
- when changing the chunk while displaying the log, the following QML warning appear: `TextFileViewer.qml:274: TypeError: Value is undefined and could not be converted to an object`

fabiencastan:
The colorbar should have the line height:
image

fabiencastan:
Time in tooltip should be formatted h:m:s.
The zone to display the Tooltip should include the colorbar.

fabiencastan:
Should interpolate colors.

fabiencastan:
The colorbar should be grey if the timing is 0 (and not the same than the background), so we can always see a continuous vertical line.

meshroom/ui/qml/Utils/Colors.qml
```diff
@@ -61,17 +61,21 @@ QtObject {
         return "magenta"
     }
 
-    function durationColorScale(t) {
```
fabiencastan: You should interpolate between: grey > green > yellow > red.

meshroom/ui/qml/Controls/TextFileViewer.qml
```diff
@@ -203,7 +203,7 @@ Item {
                     Rectangle {
                         width: 4
                         Layout.fillHeight: true
-                        color: duration > 0 ? Colors.durationColorScale(duration) : "transparent"
+                        color: Colors.interpolate(Colors.grey, Colors.red, duration/180)
```
fabiencastan: Need to use Colors.durationColorScale to get something visible. Just add interpolation in it.

fabiencastan (approved):

PR #1788, opened 2022-09-30
[ui] Paste nodes at the center of the Graph Editor when it does not contain the mouse

cbentejac:
## Description

When using the Edit > Paste action menu, or when pressing Ctrl+V while the Graph Editor has the focus but does not contain the mouse, there is no current mouse position within the GraphEditor. The position that is provided to the "pasteNodes" function is the last known mouse position within the Graph Editor, which oftentimes corresponds to its borders.

Instead of using the border of the Graph Editor as the top-left corner of the pasting zone (the zone in which all the nodes contained in the clipboard will be pasted, with their original distance to one another preserved), this PR simulates the mouse's position and sets it at the center of the Graph Editor. This position is then used as the center of the pasting zone instead of its top-left corner. Distances between the pasted nodes remain unchanged.

### DIFF
--- a/meshroom/ui/graph.py
+++ b/meshroom/ui/graph.py
@@ -789,8 +789,8 @@ class UIGraph(QObject):
             return json.dumps(selection, indent=4)
         return ''
 
-    @Slot(str, QPoint, result="QVariantList")
-    def pasteNodes(self, clipboardContent, position=None):
+    @Slot(str, QPoint, bool, result="QVariantList")
+    def pasteNodes(self, clipboardContent, position=None, centerPosition=False):
         """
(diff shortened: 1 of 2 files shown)


fabiencastan:
Can we not just check if the mouse position is in the GraphEditor? If it's not inside, the paste is done at the center.

cbentejac:
> Can we not just check if the mouse position is in the GraphEditor? If it's not inside, the paste is done at the center.

You're right: that way, we can also paste nodes at the center when the Graph Editor still has the focus but does not contain the mouse.

fabiencastan (approved):

PR #1805, opened 2022-10-26
Update bounding box display to use the correct geometric frame

servantftechnicolor:
Bounding box for the meshing node 

The bounding box manipulator was not storing parameters in the correct geometric frame


mh0g (approved):
Offset issue fixed

fabiencastan (approved):

fabiencastan (approved):

PR #1806, opened 2022-10-26
added an explicit debug message for missing nodes

mh0g:
Added an error message to replace 'NoneType' object has no attribute 'processChunk' when encountering a missing node description

### DIFF
--- a/meshroom/core/node.py
+++ b/meshroom/core/node.py
@@ -408,6 +408,8 @@ class NodeChunk(BaseObject):
         self.statThread = stats.StatisticsThread(self)
         self.statThread.start()
         try:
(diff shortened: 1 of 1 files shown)


cbentejac (approved):

meshroom/core/node.py
```diff
@@ -408,6 +408,8 @@ def process(self, forceCompute=False):
         self.statThread = stats.StatisticsThread(self)
         self.statThread.start()
         try:
+            if self.node.nodeDesc is None:
```
fabiencastan, line 411: It should be done at node's creation, not at each compute, isn't it?

Issue #1797, opened 2022-10-17
meshroom_batch change FeatureExtraction:describerTypes syntax 


Dev-hestabit:
Using the command line tool "meshroom_batch" I am able to override parameters like the "forceCpuExtraction" successfully like this:

./meshroom_batch -i /input -o /output --paramOverrides FeatureExtraction:forceCpuExtraction=0

If I wantet o override the "describerTypes" to sift + akaze (for example) as well I tried something like this:

./meshroom_batch -i /input -o /output --paramOverrides FeatureExtraction:forceCpuExtraction=0 FeatureExtraction:describerTypes=sift,akaze

This does not work. The error message I get is the following:

ValueError: ChoiceParam value "s" is not in "['sift', 'sift_float', 'sift_upright', 'dspsift', 'akaze', 'akaze_liop', 'akaze_mldb', 'cctag3', 'cctag4', 'sift_ocv', 'akaze_ocv']".

So obviously the syntax I use to assign the two types via command line is incorrect.
Any help on the correct syntax would be appreciated.

I am running on Linux (arch) with the latest version of meshroom (2021.1.0)

mugulmd:
Hey @Dev-hestabit, thanks for reporting this!

Unfortunately after looking at the source code it looks like what you're trying to achieve here (passing multiple values to a non-exclusive ChoiceParam from the command line) is currently not supported, so there's no correct syntax (at least to my knowledge).

I'll see if I can fix this!

Dev-hestabit:
I have fix this issue by changing the FeatureExtraction, Feature Matching and Structured for motion .pyc file with the git .py file and changes the default value from dspsift to sift,akaze and without passing any paramOverrides it works exactly with command line that i want it to using command line

Dev-hestabit:
Is there anyway apart from CCtag for scaling 3d model to its real life size?

mugulmd:
> Is there anyway apart from CCtag for scaling 3d model to its real life size?

Sorry I'm not sure I see how this is related to this issue..

PR #1811, opened 2022-10-31
[bin] meshroom_batch: allow passing list of values to param overrides

mugulmd:
## Description

Fix https://github.com/alicevision/Meshroom/issues/1797

Currently it is not possible to pass a list of values with the `--paramOverrides` argument of `meshroom_batch`, which is problematic when overriding non-exclusive choice parameters, such as `FeatureExtraction:describerTypes`.

This PR provides the following mechanism to fix this issue: if the input string contains at least one comma it will be interpreted as a list of values (split around the commas), otherwise it will be interpreted as a single value.

### DIFF
--- a/bin/meshroom_batch
+++ b/bin/meshroom_batch
@@ -157,9 +157,6 @@ with multiview.GraphModification(graph):
             if not result:
                 raise ValueError('Invalid param override: ' + str(p))
             node, t, param, value = result.groups()
-            multiValues = value.split(',')
-            if len(multiValues) > 1:
-                value = multiValues
             if t == ':':
                 nodesOfType = graph.nodesOfType(node)
                 if not nodesOfType:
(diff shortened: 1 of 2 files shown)


bin/meshroom_batch
```diff
@@ -157,6 +157,9 @@ with multiview.GraphModification(graph):
             if not result:
                 raise ValueError('Invalid param override: ' + str(p))
             node, t, param, value = result.groups()
+            multiValues = value.split(',')
```
fabiencastan, line 160: Should only be done if the attribute could contain multiple values, like groups or lists.

fabiencastan: See https://github.com/alicevision/Meshroom/blob/develop/meshroom/core/desc.py#L84

fabiencastan (approved):

PR #1812, opened 2022-10-31
Add new file watcher behaviours

cbentejac:
## Description

The file watcher behaviour, prior to this PR, was to check every single chunk's status file every 5 seconds, no matter the current status of the graph or the nodes. 

This PR adds a button in the GraphEditor to change the settings of the file watcher, with 3 different modes:
- Auto-refresh: this is the file watcher's current mode; every single chunk's status file is checked every 5 seconds, with no exception
- Disabled: the file watcher is simply disabled; the watcher thread is either stopped/never started, and no chunk's file status is checked
- Minimal auto-refresh: every 5 seconds, the file watcher checks the status files of currently submitted or running chunks; if there are no submitted or running chunks, then the file watcher does not do anything

Additionally, at the start of Meshroom, if no submitter is detected, the file watcher is disabled by default; if a submitter is detected, then the classic auto-refresh mode is enabled. 

## Features list

- [x] Add a button to change the file watcher's mode
- [x] Add two different behaviours for the file watcher: disabled and minimal auto-refresh

## Implementation remarks

In minimal auto-refresh mode, the thread that periodically checks the status files is still running, whether it has chunks' status files to watch or not. If the list of status files to watch is empty, then it will not do anything, but it will still run. The only mode that prevents the thread from running is the "disable" mode.


meshroom/ui/graph.py
```diff
@@ -118,8 +142,16 @@ def stop(self):
 
     @property
     def statusFiles(self):
```
fabiencastan, line 144: Need a dedicated property named for instance "watchedStatusFiles", as the statusFiles is semantically supposed to always return the list of all status files.

fabiencastan (changes requested):

meshroom/ui/graph.py
cbentejac: Done

meshroom/ui/qml/main.qml
```diff
@@ -902,6 +902,81 @@ ApplicationWindow {
...
+                        onClicked: {
+                            refreshFilesMenu.open()
+                        }
+                        enabled: true
+                        Menu {
+                            id: refreshFilesMenu
+                            width: 210
+                            y: parent.height
+                            x: -width + parent.width
+                            MenuItem {
```
fabiencastan, line 921: Would `autoExclusive: true` change the icon?

meshroom/ui/qml/main.qml
```diff
@@ -902,6 +902,81 @@ ApplicationWindow {
...
+                                ToolTip.delay: 200
+                                onToggled: {
+                                    if (checked) {
+                                        disableAutoRefresh.checked = false
+                                        enableAutoRefresh.checked = false
+                                        _reconstruction.filePollerRefreshChanged(2)
+                                    }
+                                    // Prevents cases where the user unchecks the currently checked option
+                                    minimalAutoRefresh.checked = true
+                                    filePollerRefreshStatus.text = MaterialIcons.sync
```
fabiencastan, line 975: Maybe https://materialui.co/material-icons-two-tone/published-with-changes to have 1 logo per mode?

cbentejac: Done.

meshroom/ui/qml/main.qml
cbentejac: It didn't make a difference: with or without the autoExclusive, the icon is always a state behind unless it is updated "manually" when a MenuItem is selected.

meshroom/ui/qml/main.qml
```diff
@@ -909,8 +909,8 @@ ApplicationWindow {
                     }
                     MaterialToolButton {
                         id: filePollerRefreshStatus
-                        text: _reconstruction.filePollerRefresh == 0 ? MaterialIcons.sync : MaterialIcons.sync_disabled
-                        ToolTip.text: "Set File Refresh Status"
+                        text: (_reconstruction.filePollerRefresh == 0 || _reconstruction.filePollerRefresh == 2) ? MaterialIcons.sync : MaterialIcons.sync_disabled
```
fabiencastan: Use enum values like PollerRefreshStatus.AUTO_ENABLED instead of 0.

meshroom/ui/qml/main.qml
```diff
@@ -939,19 +944,42 @@ ApplicationWindow {
                             }
                             MenuItem {
                                 id: disableAutoRefresh
-                                text: "Disable Auto-Refresh"
+                                text: "Disable File Watcher"
```
fabiencastan, line 964: I do not see the reason to change the wording.
It should be "No Auto-Refresh".

meshroom/ui/qml/main.qml
```diff
@@ -909,8 +909,8 @@ ApplicationWindow {
                     }
                     MaterialToolButton {
                         id: filePollerRefreshStatus
-                        text: _reconstruction.filePollerRefresh == 0 ? MaterialIcons.sync : MaterialIcons.sync_disabled
-                        ToolTip.text: "Set File Refresh Status"
+                        text: (_reconstruction.filePollerRefresh == 0 || _reconstruction.filePollerRefresh == 2) ? MaterialIcons.sync : MaterialIcons.sync_disabled
+                        ToolTip.text: "Set File Watcher Refresh Status"
```
fabiencastan: From a user perspective, we do not care about the notion of file watcher.

ToolTip.text: "Auto-Refresh Nodes Status For External Changes. \n\n"+
                     "If the status of a node is changed by another instance \n" +
                     "on the same network, such as when computing in render farm."

fabiencastan: Warning: this string is update somewhere with " (Current: STATUS)" but I do not see where.

fabiencastan:
Minimal refresh does not update changes on submitted nodes as expected.

meshroom/ui/qml/main.qml
cbentejac: The enum is created on the Python side, and there is no way to expose it to QML with PySide2 (https://bugreports.qt.io/browse/PYSIDE-957).

meshroom/ui/qml/main.qml
cbentejac: Done.

cbentejac:
The issue with the updates of some chunks in minimal auto-refresh has been fixed, and a safeguard has been added to prevent submitting or computing the same chunks twice, in cases where a graph might not be up-to-date.

(and the branch has been rebased on `develop`)

fabiencastan (approved):
Great work!

PR #1813, opened 2022-11-02
[ui] keyboard shortcut: press tab to open node menu

mugulmd:
## Description

This PR introduces a new keyboard shortcut in the graph editor: TAB to open the node menu (same behavior as right clicking).

This has been requested by a user, the main reason being that this shortcut is used in several other programs (Houdini, Maya, Nuke, etc.).

### DIFF
--- a/meshroom/ui/qml/GraphEditor/GraphEditor.qml
+++ b/meshroom/ui/qml/GraphEditor/GraphEditor.qml
@@ -118,19 +118,30 @@ Item {
     }
 
     Keys.onPressed: {
-        if (event.key === Qt.Key_F)
-            fit()
-        if (event.key === Qt.Key_Delete)
-            if (event.modifiers == Qt.AltModifier)
-                uigraph.removeNodesFrom(uigraph.selectedNodes)
-            else
-                uigraph.removeNodes(uigraph.selectedNodes)
-        if (event.key === Qt.Key_D)
-            duplicateNode(event.modifiers == Qt.AltModifier)
-        if (event.key === Qt.Key_C && event.modifiers == Qt.ControlModifier)
-            copyNodes()
-        if (event.key === Qt.Key_V && event.modifiers == Qt.ControlModifier)
-            pasteNodes()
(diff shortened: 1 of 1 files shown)


cbentejac (commented):
For some reason, when you create a node after opening the menu with tab instead of right-click, the node is successfully created and appears to be selected (it is highlighted in blue as expected) but it doesn't respond to any keyboard shortcut (you can't delete it with "suppr" or duplicate it with "D" whereas you can when the menu is opened through the right-click).

You need to unselect and then reselect the node for the keyboard shortcuts to be effective on that node.

cbentejac (approved):

meshroom/ui/qml/GraphEditor/GraphEditor.qml
```diff
@@ -118,19 +118,31 @@ Item {
...
+        }
+        else if (event.key === Qt.Key_C && event.modifiers == Qt.ControlModifier) {
+            copyNodes();
+        }
+        else if (event.key === Qt.Key_V && event.modifiers == Qt.ControlModifier) {
+            pasteNodes();
+        }
+        else if (event.key == Qt.Key_Tab) {
+            event.accepted = true;
+            newNodeMenu.spawnPosition = mouseArea.mapToItem(draggable, mouseArea.mouseX, mouseArea.mouseY);
```
fabiencastan, line 143: Check if the mouse position is in the GraphEditor and if it's not inside, the node creation should be done in the center.

fabiencastan: See https://github.com/alicevision/Meshroom/pull/1788

cbentejac: I don't think the menu should appear if the GraphEditor doesn't have the focus: the "tab" key is ambiguous, it is used by default to select one element after another outside the GraphEditor (and it is the "classic" behaviour for that key). 

If the node creation menu appears in the center of the GraphEditor even when it doesn't have the focus, then this behaviour will be overridden.

PR #1816, opened 2022-11-04
[ui] image gallery search bar

mugulmd:
## Description

This PR introduces a new search bar in the image gallery to filter viewpoints using their filename or their uid.

## TODO

- [X] add search bar UI
- [X] search based on filename 
- [ ] search based on uid
- [ ] focus search bar from gallery with tab

### DIFF
--- a/meshroom/ui/qml/ImageGallery/ImageGallery.qml
+++ b/meshroom/ui/qml/ImageGallery/ImageGallery.qml
@@ -106,6 +106,14 @@ Panel {
     }
 
     headerBar: RowLayout {
+        SearchBar {
+            id: searchBar
+            width: 100
+            onTextChanged: {
+                sortedModel.updateFilter("path", text);
+            }
+        }
+
         MaterialToolButton {
             text: MaterialIcons.more_vert
             font.pointSize: 11


cbentejac (commented):
The filtering works if you're using regular strings, and partially works when using regular expressions (both for the filtering in the Image Gallery and in the metadata display, which is not directly concerned by this PR):
- the "." character works as any character and doesn't seem to generate any errors
- the "*" characters automatically cancels any filtering that it precedes/follows: as soon as it appears in the filtering string, no matter its position, it's like no filtering is applied. It also generates a syntax error warning in meshroom/ui/qml/Utils/SortFilterDelegateModel.qml at line 95 ("SyntaxError: Invalid regular expression").

I haven't tested any complex regex, and this behaviour seems to be the regular one for the metadata filtering (independently from this PR). 

I don't know if we want to support regexes or not in the filtering but in any case, something should probably be done to either support them correctly, or not support them at all. 

Apart from that, the code looks good to me.

mugulmd:
> The filtering works if you're using regular strings, and partially works when using regular expressions (both for the filtering in the Image Gallery and in the metadata display, which is not directly concerned by this PR):
> 
>     * the "." character works as any character and doesn't seem to generate any errors
> 
>     * the "*" characters automatically cancels any filtering that it precedes/follows: as soon as it appears in the filtering string, no matter its position, it's like no filtering is applied. It also generates a syntax error warning in meshroom/ui/qml/Utils/SortFilterDelegateModel.qml at line 95 ("SyntaxError: Invalid regular expression").
> 
> 
> I haven't tested any complex regex, and this behaviour seems to be the regular one for the metadata filtering (independently from this PR).
> 
> I don't know if we want to support regexes or not in the filtering but in any case, something should probably be done to either support them correctly, or not support them at all.
> 
> Apart from that, the code looks good to me.

Since we can't expect users to know about how regexp work, it would feel like a bug to them when entering characters like "." as it would trigger quite a lot of matches, so I removed the support for regexp: we're simply trying to match a sub-string now (which is coherent with how things are done in the new node menu of the graph editor by the way).

fabiencastan:
Maybe one day it would be good to support simple expressions like
https://docs.python.org/3/library/fnmatch.html

mugulmd:
> Maybe one day it would be good to support simple expressions like https://docs.python.org/3/library/fnmatch.html

Added to my TODO list 😄

PR #1819, opened 2022-11-08
New nodes for large memory use in panoramas

servantftechnicolor:
Very large panoramas required to update the panorama seams node and replace image processing node with a specific node to handle large memory consumption

### DIFF
--- a/meshroom/nodes/aliceVision/PanoramaPostProcessing.py
+++ b/meshroom/nodes/aliceVision/PanoramaPostProcessing.py
@@ -1,0 +1,53 @@
+__version__ = "1.0"
+
+import json
+import os
+
+from meshroom.core import desc
+
+
+class PanoramaPostProcessing(desc.CommandLineNode):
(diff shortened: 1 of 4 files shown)


fabiencastan:
We should also change the panorama template accordingly, isn't it?

servantftechnicolor:
> We should also change the panorama template accordingly, isn't it?

waiting for support

meshroom/nodes/aliceVision/PanoramaPostProcessing.py
```diff
@@ -0,0 +1,53 @@
...
+            values=['fatal', 'error', 'warning', 'info', 'debug', 'trace'],
+            exclusive=True,
+            uid=[],
+        ),
+    ]
+
+    outputs = [
+        desc.File(
+            name='outputPanorama',
+            label='Output Folder',
```
fabiencastan, line 48:
```suggestion
            label='Output Panorama',
```

fabiencastan (approved):

PR #1832, opened 2022-11-25
[ui] Correctly determine if a graph is being computed locally and update nodes' statuses accordingly

cbentejac:
## Description

This PR fixes two bugs related to local graph computations:
- When computing a graph in an instance of Meshroom, and opening the same computing graph in another instance of Meshroom, the second instance of Meshroom behaves as if it was performing the computations even though it was not. As a consequence, trying to open a new graph was failing (Meshroom would try stopping the computations, but fail since they were carried out by the first instance), and so was closing the instance, with a "Please stop any local computation before exiting Meshroom" error message. Only the chunks' statuses were taken into account to determine whether local computations were ongoing.
- When loading a new graph while the opened one was locally computing nodes, computations were automatically stopped to be able to load the new graph. However, the computing chunks' statuses were not being updated before the graphs were switched up. This meant that opening that previously computing graph again was leading to a display in which nodes appeared to be computing although there was no ongoing computations. The only way to get back to a correct state was manually unlocking the nodes and clearing the pending statuses.

## Features list

- [X] Take the Meshroom instance's session ID when determining whether computations are made locally, and not only the chunks' statuses.
- [X] Clear locally submitted nodes before loading a new graph to avoid stopping the computations without updating the chunks' statuses.

### DIFF
--- a/meshroom/ui/graph.py
+++ b/meshroom/ui/graph.py
@@ -284,6 +284,9 @@ class UIGraph(QObject):
         """ Set the internal graph. """
         if self._graph:
             self.stopExecution()
+            # Clear all the locally submitted nodes at once before the graph gets changed, as it won't receive further updates
+            if self._computingLocally:
+                self._graph.clearLocallySubmittedNodes()
             self.clear()
         oldGraph = self._graph
         self._graph = g
--- a/meshroom/ui/graph.py
+++ b/meshroom/ui/graph.py
@@ -11,6 +11,7 @@ from multiprocessing.pool import ThreadPool
 from PySide2.QtCore import Slot, QJsonValue, QObject, QUrl, Property, Signal, QPoint
 
 from meshroom import multiview
+from meshroom.core import sessionUid
 from meshroom.common.qt import QObjectListModel
 from meshroom.core.attribute import Attribute, ListAttribute
 from meshroom.core.graph import Graph, Edge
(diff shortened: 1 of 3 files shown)


meshroom/ui/graph.py
```diff
@@ -454,7 +458,9 @@ def submit(self, node=None):
 
     def updateGraphComputingStatus(self):
         # update graph computing status
-        computingLocally = any([ch.status.execMode == ExecMode.LOCAL and ch.status.status in (Status.RUNNING, Status.SUBMITTED) for ch in self._sortedDFSChunks])
+        computingLocally = any([ch.status.execMode == ExecMode.LOCAL and
+                                ch.status.status in (Status.RUNNING, Status.SUBMITTED) and
+                                ch.status.sessionUid == sessionUid for ch in self._sortedDFSChunks])
```
fabiencastan, line 463:
```suggestion
        # Check if we have one or more nodes computing locally with the current identifier of the Meshroom session
        computingLocally = any([
                                (ch.status.execMode == ExecMode.LOCAL and
                                ch.status.sessionUid == sessionUid and
                                ch.status.status in (Status.RUNNING, Status.SUBMITTED))
                                    for ch in self._sortedDFSChunks])
```

meshroom/ui/graph.py
```diff
@@ -283,6 +284,9 @@ def setGraph(self, g):
         """ Set the internal graph. """
         if self._graph:
             self.stopExecution()
+            # Clear all the submitted nodes at once before the graph gets changed, as it won't receive further updates
+            if self._computingLocally:
+                self._graph.clearSubmittedNodes()
```
fabiencastan, line 289: It will change the status of all nodes including those submitted externally in renderfarm.

cbentejac: You're right, nodes submitted externally would also have been cleared if at least one chunk was being computed locally at the same time. This is now fixed!

meshroom/ui/graph.py
cbentejac: Done.

cbentejac:
> Should we not also change the meaning of isExtern?
> 
> https://github.com/alicevision/Meshroom/blob/b08f83552e6b89272e5731c2e01a16d92bd41072/meshroom/core/node.py#L435
> 
> 
> Instead of:
> ```
>     def isExtern(self):
>         return self._status.execMode == ExecMode.EXTERN
> ```
> 
> Use:
> 
> ```
>     def isExtern(self):
>         return self._status.execMode == ExecMode.EXTERN or (
>                 self._status.execMode == ExecMode.LOCAL and self._status.sessionUid != sessionUi)
> ```

You're totally right, it's done!

fabiencastan (approved):

PR #1833, opened 2022-11-29
[ui] GraphEditor: create new pipelines with the node menu

cbentejac:
## Description

This PR adds a "Pipelines" category to the node menu (opened with a right click or a press of the TAB key in the GraphEditor) which contains the list of available templates. Selecting a pipeline through the menu creates said pipeline into the current scene, with its top-left corner being placed on the mouse's position. Pipelines can be searched using the search bar at the same time as any node type.

## Features list

- [X] Add the list of available templates in a "Pipelines" category in the node menu
- [X] If a pipeline in the "Pipelines" category is clicked, create it and add it to the current scene (like any node type)
- [X] Support searching for a pipeline in the node menu's search bar (like any node type)

## Implementation remarks

- Creating a pipeline with the node menu is functionally akin to importing a template into the current project, and moving it to the mouse's position (instead of leaving it below the current graph).
- The "Pipelines" category's creation is hard-coded, and then filled with the list of the available templates' names. 
- Since the "Pipelines" category is added as a regular one, it is alphabetically sorted with the other categories, which only contain node types.

### DIFF
--- a/meshroom/ui/app.py
+++ b/meshroom/ui/app.py
@@ -198,6 +198,9 @@ class MeshroomApp(QApplication):
             templates.append(variant)
         return templates
 
+    def _pipelineTemplateNames(self):
+        return [p["name"] for p in self.pipelineTemplateFiles]
+
     @Slot()
     def reloadTemplateList(self):
         for f in meshroom.core.pipelineTemplatesFolders:
@@ -352,5 +355,6 @@ class MeshroomApp(QApplication):
     pipelineTemplateFilesChanged = Signal()
     recentProjectFilesChanged = Signal()
     pipelineTemplateFiles = Property("QVariantList", _pipelineTemplateFiles, notify=pipelineTemplateFilesChanged)
+    pipelineTemplateNames = Property("QVariantList", _pipelineTemplateNames, notify=pipelineTemplateFilesChanged)
     recentProjectFiles = Property("QVariantList", _recentProjectFiles, notify=recentProjectFilesChanged)
     default8bitViewerEnabled = Property(bool, _default8bitViewerEnabled, constant=True)
--- a/meshroom/ui/commands.py
+++ b/meshroom/ui/commands.py
@@ -245,7 +245,7 @@ class ImportProjectCommand(GraphCommand):
             else:
                 self.graph.node(node.name).position = Position(node.x, node.y + lowestY + self.yOffset)
 
-        return status
+        return importedNodes
 
     def undoImpl(self):
         for nodeName in self.importedNames:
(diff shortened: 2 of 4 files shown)


meshroom/ui/app.py
```diff
@@ -352,5 +355,6 @@ def _default8bitViewerEnabled(self):
     pipelineTemplateFilesChanged = Signal()
     recentProjectFilesChanged = Signal()
     pipelineTemplateFiles = Property("QVariantList", _pipelineTemplateFiles, notify=pipelineTemplateFilesChanged)
+    pipelineTemplateNames = Property("QVariantList", _pipelineTemplateNames, notify=pipelineTemplateFilesChanged)
```
mugulmd, line 358: I don't think this new property is needed, you could directly use `pipelineTemplateFiles` and avoid duplicating data

meshroom/ui/commands.py
```diff
@@ -239,9 +240,12 @@ def redoImpl(self):
 
         for node in importedNodes:
             self.importedNames.append(node.name)
-            self.graph.node(node.name).position = Position(node.x, node.y + lowestY + self.yOffset)
+            if self.position:
```
mugulmd, line 243:
```suggestion
            if self.position is not None:
```

meshroom/ui/graph.py
```diff
@@ -363,8 +363,13 @@ def importProject(self, filepath):
                 localFile = filepath.toString()
         else:
             localFile = filepath
+        if isinstance(position, QPoint):
+            if position.x() == 0 and position.y() == 0:
```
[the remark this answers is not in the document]
cbentejac: Done.

meshroom/ui/app.py
cbentejac: It is not needed per se, but I feel like it simplifies greatly the parsing on the QML side: https://github.com/alicevision/Meshroom/pull/1833/commits/8bd4b6e04751d3425cb7373968631ed054f65741 
The parsing was, admittedly, not as efficient as it could have been but I still think there's a benefit in having `pipelineTemplateFiles`.

meshroom/ui/graph.py
```diff
@@ -363,8 +364,10 @@ def importProject(self, filepath):
                 localFile = filepath.toString()
         else:
             localFile = filepath
+        if isinstance(position, QPoint):
+                position = Position(position.x(), position.y())
```
fabiencastan, line 368: Should we not import at the center of the GraphEditor when we do not have a mouse position, like for the copy/paste nodes?

PR #1836, opened 2022-12-02
[core] Add cgroups support to meshroom

servantftechnicolor:
CGroups on linux may limit the memory and the number of cpus

Without getting the specific information about these limits, we may overshoot our memory usage and core usage, creating crashes.

This PR will read the values as specified by the specifications.

Rely on https://github.com/alicevision/AliceVision/pull/1304.

### DIFF
--- a/meshroom/core/cgroup.py
+++ b/meshroom/core/cgroup.py
@@ -43,10 +43,10 @@ def getCgroupMemorySize():
 
     return size
 
-def parseNumericList(str):
+def parseNumericList(numericListString):
 
     nList = []
-    for item in str.split(','):
+    for item in numericListString.split(','):
         if '-' in item:
             start, end = item.split('-')
             start = int(start)
--- a/meshroom/core/cgroup.py
+++ b/meshroom/core/cgroup.py
@@ -25,7 +25,7 @@ def getCgroupMemorySize():
                     continue
                 if tokens[1] == "memory":
                     cgroup = tokens[2]
-    except IOError:
+    except OSError:
         pass
 
     if cgroup is None:
@@ -38,7 +38,7 @@ def getCgroupMemorySize():
             value = f.read().rstrip("\r\n")
             if value.isnumeric():
                 size = int(value)
-    except IOError:
+    except OSError:
         pass
 
     return size
@@ -80,7 +80,7 @@ def getCgroupCpuCount():
                     continue
                 if tokens[1] == "cpuset":
                     cgroup = tokens[2]
-    except IOError:
+    except OSError:
         pass
 
     if cgroup is None:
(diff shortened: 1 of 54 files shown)


fabiencastan:
We need to find a way for nodes to declare if they supports these arguments or rely on environment variables.
We have many nodes launching external commands (like blender).
We could for instance create an intermediate class for AVCommandLineNode.

mh0g:
we could use a  chunk.node.nodeDesc.commandLine.startswith("aliceVision_") to do this only on the proper nodes

mh0g (approved):

mh0g:
a warning in the log when the feature is used may be usefull

meshroom/core/cgroup.py
```diff
@@ -0,0 +1,101 @@
...
+        return -1
+
+    size = -1
+    filename = f"/sys/fs/cgroup/memory/{cgroup}/memory.limit_in_bytes"
+    try:
+        with open(filename, "r") as f :
+            value = f.read().rstrip("\r\n")
+            if value.isnumeric():
+                size = int(value)
+    except IOError:
```
fabiencastan, line 41: OSError

meshroom/core/desc.py
```diff
@@ -589,6 +593,23 @@ def processChunk(self, chunk):
...
             chunk.subprocess = None
 
+#specific command line node for alicevision apps
+class AVCommandLineNode(CommandLineNode):
+    def buildCommandLine(self, chunk):
+
+        str = super(AVCommandLineNode, self).buildCommandLine(chunk)
+
+        cmdMem = ''
+        memSize = cgroup.getCgroupMemorySize()
```
fabiencastan, line 603: This is a static information that should be estimated once for the lifetime of the session and not each time we evaluate the command line per node.

fabiencastan (changes requested):

meshroom/core/cgroup.py
```diff
@@ -0,0 +1,101 @@
...
+        with open(filename, "r") as f :
+            value = f.read().rstrip("\r\n")
+            if value.isnumeric():
+                size = int(value)
+    except OSError:
+        pass
+
+    return size
+
+def parseNumericList(str):
```
fabiencastan, line 46: str is a python keyword, it would be good to rename this variable.

meshroom/core/desc.py
```diff
@@ -589,6 +593,34 @@ def processChunk(self, chunk):
...
+            AVCommandLineNode.cmdCore = ''
+            coresCount = cgroup.getCgroupCpuCount()
+            if coresCount > 0:
+                AVCommandLineNode.cmdCore = ' --maxCores={coresCount}'.format(coresCount=coresCount)
+
+            AVCommandLineNode.cgroupParsed = True
+
+    def buildCommandLine(self, chunk):
+
+        str = super(AVCommandLineNode, self).buildCommandLine(chunk)
```
fabiencastan, line 621: same

fabiencastan (approved):

PR #1839, opened 2022-12-05
[ui] Fix all "TypeError" QML warnings

cbentejac:
## Description

Some QML properties access exposed Python objects that may or may not be null upon their access. When these objects are accessed while null, QML issues "TypeError" warnings. These warnings have no functional impact as QML correctly handles trying to access null objects, but can spam the logs.

This PR fixes all these warnings by ensuring that the Python objects are not null before they are accessed by QML properties. In case the objects are null, default values are returned instead of still attempting to access them anyway.

### DIFF
--- a/meshroom/ui/qml/Viewer/Viewer2D.qml
+++ b/meshroom/ui/qml/Viewer/Viewer2D.qml
@@ -24,7 +24,7 @@ FocusScope {
     property alias useLensDistortionViewer: displayLensDistortionViewer.checked
     property alias usePanoramaViewer: displayPanoramaViewer.checked
 
-    property var activeNodeFisheye: _reconstruction.activeNodes.get("PanoramaInit").node
+    property var activeNodeFisheye: _reconstruction ? _reconstruction.activeNodes.get("PanoramaInit").node : null
     property bool cropFisheye : activeNodeFisheye ? activeNodeFisheye.attribute("useFisheye").value : false
     property bool enable8bitViewer: MeshroomApp.default8bitViewerEnabled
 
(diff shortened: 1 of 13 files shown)


meshroom/ui/qml/Viewer/Viewer2D.qml
```diff
@@ -220,12 +220,14 @@ FocusScope {
     function getFileAttributePath(node, attrName, viewId) {
         // get output attribute with matching name
         // and parse its value to get the image filepath
-        for (var i = 0; i < node.attributes.count; i++) {
-            var attr = node.attributes.at(i);
-            if (attr.name == attrName) {
-                let pattern = String(attr.value).replace("<VIEW_ID>", viewId);
-                let path = Filepath.globFirst(pattern);
-                return Filepath.stringToUrl(path);
+        if (node) {
```
fabiencastan, line 223:
```
if(!node)
    return "";
```

meshroom/ui/qml/main.qml
```diff
@@ -404,43 +408,45 @@ ApplicationWindow {
...
         function getSelectedNodesName()
         {
             var nodesName = ""
-            for (var i = 0; i < _reconstruction.selectedNodes.count; i++)
-            {
-                if (nodesName !== "")
-                    nodesName += ", "
-                var node = _reconstruction.selectedNodes.at(i)
-                nodesName += node.name
+            if (_reconstruction) {
```
fabiencastan, line 442: if(!_reconstruction)
    return "";

cbentejac: Done

fabiencastan (approved):

PR #1842, opened 2022-12-08
[nodes] new SfMTriangulation node

mh0g:
## Description

Adds a main that is essentially the triangulation step from the sequential SfM.
AliceVision PR: https://github.com/alicevision/AliceVision/pull/1314

## Features list
This allows to run the triangulation without going into the sfm step.

### DIFF
--- a/meshroom/nodes/aliceVision/Triangulate.py
+++ b/meshroom/nodes/aliceVision/Triangulate.py
@@ -1,4 +1,4 @@
-__version__ = "2.0"
+__version__ = "1.0"
 
 from meshroom.core import desc
 
@@ -14,7 +14,7 @@ class Triangulate(desc.AVCommandLineNode):
         desc.File(
             name='input',
             label='SfMData',
-            description='SfMData file.',
+            description='SfMData file. Must contain the camera calibration.',
             value='',
             uid=[0],
         ),
--- a/meshroom/nodes/aliceVision/Triangulate.py
+++ b/meshroom/nodes/aliceVision/Triangulate.py
@@ -8,7 +8,7 @@ class Triangulate(desc.AVCommandLineNode):
     size = desc.DynamicNodeSize('input')
 
     category = 'Sparse Reconstruction'
-    documentation = '''Perfom keypoint triangulation (similarly to what is done in the SfM)'''
+    documentation = '''Perfom keypoint triangulation (as it is done in the SfM)'''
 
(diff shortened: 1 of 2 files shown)


mugulmd:
When testing this node it kept failing with the following error: 
`ERROR: unrecognised option '--maxMemory=9223372036854771712'`

mugulmd:
Tested on 3 datasets by using the output of the `StructureFromMotion` node as input of the `Triangulate` node. 
Results: 
- the generated landmarks are similar to the ones generated by `StructureFromMotion`, with a few differences (which is expected since there is no bundle adjustment)
- when increasing the `minNumberOfObservationsForTriangulation` the number of landmarks decreases (which is expected).

I also ran the following sanity check: when using an `SfMData` without poses as input, no landmark is generated.

mh0g:
> Tested on 3 datasets by using the output of the `StructureFromMotion` node as input of the `Triangulate` node. Results:
> 
>     * the generated landmarks are similar to the ones generated by `StructureFromMotion`, with a few differences (which is expected since there is no bundle adjustment)
> 
>     * when increasing the `minNumberOfObservationsForTriangulation` the number of landmarks decreases (which is expected).
> 
> 
> I also ran the following sanity check: when using an `SfMData` without poses as input, no landmark is generated.

Tested on windows, seems to work as intended.
Tested on a benchmark dataset with ground truth poses and features/matches from meshroom. Small differences vs running with poses from SfM (as expected). Can also confirm that the landmarks are recomputed when using the ouptut of the Sfm node.

meshroom/nodes/aliceVision/Triangulate.py
```diff
@@ -0,0 +1,164 @@
+__version__ = "1.0"
+
+from meshroom.core import desc
+
+
+class Triangulate(desc.AVCommandLineNode):
```
fabiencastan, line 6: Triangulation

fabiencastan:
It would be good to also add a node for https://github.com/alicevision/AliceVision/blob/develop/src/software/pipeline/main_computeStructureFromKnownPoses.cpp
Or fuse them in a single node.

fabiencastan:
After checking, no need to integrate computeStructureFromKnownPoses, already possible with FeatureMatching+Triangulation.

fabiencastan (approved):

fabiencastan (approved):

PR #1852, opened 2022-12-20
[ui] Load 3D Depth Map: minor improvements

mugulmd:
## Description

This PR introduces some minor improvements around loading depth maps in 3D: 
- the "View Depth Map in 3D" button in the 2D viewer is enabled only when the currently viewed attribute corresponds to a depth map
- when loading a depth map in the 3D viewer with this button, a custom label is used in the 3D inspector composed of the node name, attribute name and view id.

### DIFF
--- a/meshroom/ui/qml/Viewer/Viewer2D.qml
+++ b/meshroom/ui/qml/Viewer/Viewer2D.qml
@@ -1170,14 +1170,17 @@ FocusScope {
 
                         MaterialToolButton {
                             id: displayImageOutputIn3D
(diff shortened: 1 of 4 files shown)


fabiencastan (changes requested):
When we double click on the depth map node, it creates an empty entry in the 3D Viewer, and when we click on the "View DepthMap in 3D" button, it creates another entry:
image

fabiencastan (approved):
In fact, it's working fine. Probably just a mistake in my setup during my previous test.

fabiencastan:
In fact, the behavior of opening the Node as an empty element in the 3D occurs when the node is in "Compatibility mode".

PR #1858, opened 2023-01-03
[ui] Checkbox to enable/disable 8-bit viewer

mugulmd:
## Description

This PR introduces a new checkbox to enable/disable the 8-bit viewer: 

checkbox_8bit_viewer

The default value of this checkbox can be changed with the `MESHROOM_USE_8BIT_VIEWER` environment variable.

### DIFF
--- a/meshroom/ui/qml/Viewer/Viewer2D.qml
+++ b/meshroom/ui/qml/Viewer/Viewer2D.qml
@@ -26,7 +26,7 @@ FocusScope {
 
     property var activeNodeFisheye: _reconstruction ? _reconstruction.activeNodes.get("PanoramaInit").node : null
     property bool cropFisheye : activeNodeFisheye ? activeNodeFisheye.attribute("useFisheye").value : false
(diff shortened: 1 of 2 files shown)


cbentejac (commented):
Looks good to me, I couldn't find a case that was not already handled. :)

I simply have a comment: since there is now an explicit mention of the 8-bit viewer for all the users, wouldn't it be appropriate to add somewhere that the 8-bit viewer is what you revert to when the HDR viewer is disabled? Maybe in the tooltip of `enable8bitViewerAction`?

fabiencastan (approved):

mugulmd:
> Looks good to me, I couldn't find a case that was not already handled. :)
> 
> I simply have a comment: since there is now an explicit mention of the 8-bit viewer for all the users, wouldn't it be appropriate to add somewhere that the 8-bit viewer is what you revert to when the HDR viewer is disabled? Maybe in the tooltip of `enable8bitViewerAction`?

PR #1860, opened 2023-01-05
[ui] ImageGallery: fix the DB path in the "Edit Sensor Database" dialog

cbentejac:
## Description

When opening the dialog to edit the sensor database, the displayed path of the database was not resolved. Instead, the environment variable that contains it was displayed:

This PR displays the resolved path to the sensor database instead. As a direct consequence, the "Copy Path" button does indeed copy the absolute path to the sensor database (rather than the environment variable itself), and the "Open in External Editor" button can be used again to open the database in the set editor.

### DIFF
--- a/meshroom/ui/qml/ImageGallery/ImageGallery.qml
+++ b/meshroom/ui/qml/ImageGallery/ImageGallery.qml
@@ -143,7 +143,7 @@ Panel {
 
     SensorDBDialog {
         id: sensorDBDialog
-        sensorDatabase: cameraInit ? Filepath.stringToUrl(cameraInit.attribute("sensorDatabase").value) : ""
+        sensorDatabase: cameraInit ? Filepath.stringToUrl(cameraInit.attribute("sensorDatabase").evalValue) : ""
         readOnly: _reconstruction ? _reconstruction.computing : false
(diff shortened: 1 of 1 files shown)


mugulmd (approved):
Changes are working fine, however the "Open in external editor" button did not work, but this is does not seem to be related to this PR.

cbentejac:
> Changes are working fine, however the "Open in external editor" button did not work, but this is does not seem to be related to this PR.

If the "Open in External Editor" fails with error 31, it is because there is no application associated with the sensor database file type, and in that case it is expected (you should get a similar message if you were to try opening it directly in your file explorer). If the error is not an error 31, then it is not normal :)

mugulmd:
> > Changes are working fine, however the "Open in external editor" button did not work, but this is does not seem to be related to this PR.
> 
> If the "Open in External Editor" fails with error 31, it is because there is no application associated with the sensor database file type, and in that case it is expected (you should get a similar message if you were to try opening it directly in your file explorer). If the error is not an error 31, then it is not normal :)

It was an error 31 indeed, thanks :)

Garoli (approved):
This is working perfectly now ! :+1:

fabiencastan (approved):

PR #1861, opened 2023-01-05
[ui] Thumbnail cache

mugulmd:
## Description

This PR introduces a new caching mechanism for thumbnails to improve interactivity when navigating in the Image Gallery.

This cache associates to each image a unique thumbnail, which is stored on disk for later reuse.

### Cache directory

The storing location is global (i.e not specific to a Meshroom scene).
The default location can be overriden with the `MESHROOM_THUMBNAIL_DIR` environment variable.

The thumbnails dimensions cannot exceed 256x256, and are stored on disk in JPEG format.
On the tested dataset, their size vary between 6 and 11 Ko (per thumbnail).

### Automatic cleaning process

When launching the application, the cache directory is scanned and all thumbnails that haven't been used for more than a certain time limit are removed. This process also makes sure that the number of thumbnails on disk does not exceed a certain limit.

The default time limit is 90 days, and can be overriden with the `MESHROOM_THUMBNAIL_TIME_LIMIT` environment variable.

The default maximum number of thumbnails on disk is 100000, and can be overriden with the `MESHROOM_MAX_THUMBNAILS_ON_DISK` environment variable.

### Asynchronous thumbnail creation

Since thumbnail creation can be quite long on large image datasets it is performed asynchronously (using 3 worker threads) to avoid blocking the UI.
A loading circle is displayed during the thumbnail creation process to let the user know that computations are undergoing.

When a thumbnail is created, a signal is sent to the Image Gallery, and the grid view is in charge of dispatching it to the corresponding Image Delegate.

#### Details on thumbnail creation requests

Thumbnail creation requests are stored in a LIFO (last-in-first-out) structure: this way, when scrolling through the image gallery, the items that the user is currently viewing have priority in the calculations.

Also, to avoid situations in which thumbnails are never retrieved, a timer is used to re-send some requests every 2 seconds if needed.

Finally, all requests are cleared when the viewpoints change. This way, when changing the scene, if thumbnail calculations of the previous scene are not finished (which can happen if the image dataset is quite large), they won't slow done calculations of the current scene.

## Tests

Tested on Windows and Linux, using images with different formats (JPEG, EXR, RAW), aspect ratio and Exif orientation tags.

Tested varying the storage time limit and maximum number of thumbnails on disk.

Tested with existing and non-existing thumbnail directory (in the second case, the directory is created recursively).

Stress tested on a 12000 high resolution images dataset and compared to current develop branch to check that fluidity has improved.

### DIFF
--- a/meshroom/ui/components/thumbnail.py
+++ b/meshroom/ui/components/thumbnail.py
@@ -102,6 +102,11 @@ class ThumbnailCache(QObject):
         Scan the thumbnail directory and
         remove all thumbnails that have not been used for more than storageTimeLimit days.
         """
+        # Check if thumbnail directory exists
+        if not os.path.exists(ThumbnailCache.thumbnailDir):
+            logging.debug('[ThumbnailCache] Thumbnail directory does not exist yet.')
+            return
+
         # Get current time
         now = time.time()
 
--- a/meshroom/ui/components/thumbnail.py
+++ b/meshroom/ui/components/thumbnail.py
@@ -10,6 +10,7 @@ import hashlib
 import time
 import logging
 from threading import Thread
+from multiprocessing.pool import ThreadPool
 
 
 class ThumbnailCache(QObject):
@@ -57,10 +58,9 @@ class ThumbnailCache(QObject):
     thumbnailCreated = Signal(QUrl, int)
 
     # Threads info and LIFO structure for running clean and createThumbnail asynchronously
-    maxWorkerThreads = 3
-    activeWorkerThreads = 0
     requests = []
     cleaningThread = None
+    workerThreads = ThreadPool(processes=3)
 
     @staticmethod
     def initialize():
@@ -211,11 +211,9 @@ class ThumbnailCache(QObject):
             return source
 
         # Thumbnail does not exist
-        # Create request and start a thread if needed
+        # Create request and submit to worker threads
         ThumbnailCache.requests.append((imgSource, callerID))
-        if ThumbnailCache.activeWorkerThreads < ThumbnailCache.maxWorkerThreads:
-            thread = Thread(target=self.handleRequestsAsync)
-            thread.start()
+        ThumbnailCache.workerThreads.apply_async(func=self.handleRequestsAsync)
 
         return None
 
@@ -264,15 +262,9 @@ class ThumbnailCache(QObject):
     def handleRequestsAsync(self):
         """Process thumbnail creation requests in LIFO order.
 
-        This method is only meant to be called by worker threads,
-        hence it also takes care of registering/unregistering the calling thread.
-
         Note: this operation waits for the cleaning process to finish before starting,
         in order to avoid synchronization issues.
         """
-        # Register worker thread
-        ThumbnailCache.activeWorkerThreads += 1
-
         # Wait for cleaning thread to finish
         if ThumbnailCache.cleaningThread is not None and ThumbnailCache.cleaningThread.is_alive():
             ThumbnailCache.cleaningThread.join()
@@ -284,8 +276,7 @@ class ThumbnailCache(QObject):
                 self.createThumbnail(req[0], req[1])
         except IndexError:
             # No more request to process
-            # Unregister worker thread
-            ThumbnailCache.activeWorkerThreads -= 1
+            return
 
     @Slot()
     def clearRequests(self):
--- a/meshroom/ui/components/thumbnail.py
+++ b/meshroom/ui/components/thumbnail.py
@@ -200,7 +200,10 @@ class ThumbnailCache(QObject):
         # Remove all thumbnails marked as removable
         for path in toRemove:
             logging.debug(f'[ThumbnailCache] Remove {path}')
-            os.remove(path)
+            try:
+                os.remove(path)
+            except FileNotFoundError:
+                logging.error(f'[ThumbnailCache] Tried to remove {path} but this file does not exist')
 
         # Check if number of thumbnails on disk exceeds limit
         if len(remaining) > ThumbnailCache.maxThumbnailsOnDisk:
@@ -212,4 +215,7 @@ class ThumbnailCache(QObject):
             for i in range(nbToRemove):
                 path = remaining[i][0]
                 logging.debug(f'[ThumbnailCache] Remove {path}')
-                os.remove(path)
+                try:
+                    os.remove(path)
+                except FileNotFoundError:
+                    logging.error(f'[ThumbnailCache] Tried to remove {path} but this file does not exist')
--- a/meshroom/ui/components/thumbnail.py
+++ b/meshroom/ui/components/thumbnail.py
@@ -26,6 +26,7 @@ class ThumbnailCache(QObject):
     i.e. scanning this directory and removing thumbnails that have not been used for too long.
     This operation also ensures that the number of thumbnails on disk does not exceed a certain limit, 
     by removing thumbnails if necessary (from least recently used to most recently used).
+    Since this operation is done at application startup, it is also performed asynchronously.
 
     The default time limit is 90 days, 
     and can be overriden with the MESHROOM_THUMBNAIL_TIME_LIMIT environment variable.
@@ -55,10 +56,11 @@ class ThumbnailCache(QObject):
     # - an identifier for the caller, e.g. the component that sent the request (useful for faster dispatch in QML)
     thumbnailCreated = Signal(QUrl, int)
 
-    # Threads info and LIFO structure for running createThumbnail asynchronously
+    # Threads info and LIFO structure for running clean and createThumbnail asynchronously
     maxWorkerThreads = 3
     activeWorkerThreads = 0
     requests = []
+    cleaningThread = None
 
     @staticmethod
     def initialize():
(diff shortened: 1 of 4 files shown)


mugulmd:
> Tested loading on windows with jpg datasets from 16 to 1000 images, loading seems to work. After saving and loading the project, doing a clear image freezes

After some investigation it seems that this is not related to this PR. 

More specifically the problem is that for large image datasets (noticed on datasets with 1000 images) the "Clear Image" action will freeze the app.

mugulmd:
Remark: 
When loading images that do not have a thumbnail, the app blocks while the thumbnails are being computed.
It would be nice to have at least a loading icon.

mugulmd:
> Remark: When loading images that do not have a thumbnail, the app blocks while the thumbnails are being computed. It would be nice to have at least a loading icon.

Actually now thumbnail creation is asynchronous so the app doesn't block.

meshroom/ui/components/thumbnail.py
```diff
@@ -0,0 +1,210 @@
...
+class ThumbnailCache(QObject):
+    """ThumbnailCache provides an abstraction for the thumbnail cache on disk, available in QML.
+
+    For a given image file, it ensures the corresponding thumbnail exists (by creating it if necessary)
+    and gives access to it.
+    Since creating thumbnails can be long (as it requires to read the full image from disk)
+    it is performed asynchronously to avoid blocking the main thread.
+
+    The default cache location is a subdirectory of the user's home directory:
+    ~/Meshroom/thumbnails.
```
fabiencastan, line 23: https://doc.qt.io/qt-5/qstandardpaths.html
QStandardPaths::CacheLocation

meshroom/ui/components/thumbnail.py
```diff
@@ -0,0 +1,210 @@
...
+    and can be overriden with the MESHROOM_THUMBNAIL_TIME_LIMIT environment variable.
+
+    The default maximum number of thumbnails on disk is 100000, 
+    and can be overriden with the MESHROOM_MAX_THUMBNAILS_ON_DISK.
+
+    The main use case for thumbnails in Meshroom is in the ImageGallery.
+    """
+
+    # Thumbnail cache directory
+    thumbnailDir = os.path.join(Path.home(), 'Meshroom', 'thumbnails')
```
fabiencastan, line 41: https://doc.qt.io/qt-5/qstandardpaths.html
QStandardPaths::CacheLocation

meshroom/ui/components/thumbnail.py
```diff
@@ -0,0 +1,210 @@
...
+
+        # Scan thumbnail directory and gather all thumbnails to remove
+        toRemove = []
+        remaining = []
+        for entry in os.scandir(ThumbnailCache.thumbnailDir):
+            if not entry.is_file():
+                continue
+
+            # Compute storage duration since last usage of thumbnail
+            lastUsage = os.path.getmtime(entry.path)
```
fabiencastan, line 183:
```suggestion
            lastUsage = entry.stat(follow_symlinks=False).st_mtime
```
https://docs.python.org/3/library/os.html#os.DirEntry.stat
"On Unix, this method always requires a system call. On Windows, it only requires a system call if follow_symlinks is True and the entry is a reparse point (for example, a symbolic link or directory junction)."

meshroom/ui/components/thumbnail.py
```diff
@@ -0,0 +1,210 @@
...
+            return
+
+        # Get current time
+        now = time.time()
+
+        # Scan thumbnail directory and gather all thumbnails to remove
+        toRemove = []
+        remaining = []
+        for entry in os.scandir(ThumbnailCache.thumbnailDir):
+            if not entry.is_file():
```
fabiencastan, line 179:
```suggestion
            if not entry.is_file(follow_symlinks=False):
```

fabiencastan: "The result is cached on the os.DirEntry object, with a separate cache for follow_symlinks True and False. Call [os.stat()](https://docs.python.org/3/library/os.html#os.stat) along with [stat.S_ISDIR()](https://docs.python.org/3/library/stat.html#stat.S_ISDIR) to fetch up-to-date information."

It will allow to have a single system call.

meshroom/ui/qml/ImageGallery/ImageDelegate.qml
```diff
@@ -77,14 +86,17 @@ Item {
...
                     anchors.margins: 4
-                    source: root.source
-                    sourceSize: Qt.size(100, 100)
                     asynchronous: true
                     autoTransform: true
                     fillMode: Image.PreserveAspectFit
                 }
+                BusyIndicator {
+                    anchors.centerIn: parent
+                    running: !Filepath.exists(thumbnail.source)
```
fabiencastan, line 98: Now it is set only when the file already exists.
```suggestion
                    running: thumbnail.source
```

meshroom/ui/components/thumbnail.py
```diff
@@ -0,0 +1,210 @@
...
+                logging.debug(f'[ThumbnailCache] {entry.name} exceeded storage time limit')
+                toRemove.append(entry.path)
+            else:
+                # Store path and last usage time for potentially sorting and removing later
+                remaining.append((entry.path, lastUsage))
+
+        # Remove all thumbnails marked as removable
+        for path in toRemove:
+            logging.debug(f'[ThumbnailCache] Remove {path}')
+            os.remove(path)
```
fabiencastan, line 198: The various call to `os.remove` could raise if the file does not exist (for instance another instance of Meshroom).
The whole function should be in a try/except FileNotFoundError.

meshroom/ui/components/thumbnail.py
```diff
@@ -0,0 +1,210 @@
...
+
+    # Maximum number of thumbnails in the cache directory
+    maxThumbnailsOnDisk = 100000
+
+    # Signal to notify listeners that a thumbnail was created and written on disk
+    # This signal has one argument: the url of the image that the thumbnail is associated to
+    thumbnailCreated = Signal(QUrl)
+
+    # Thread pool for running createThumbnail asynchronously on a fixed number of worker threads
+    pool = ThreadPool()
```
fabiencastan, line 57: We need to limit the number of processes (by default will use the number of cpus on the workstation) as loading too many images at the same time will not improve the overall time and will slow down the load of the first image.
```suggestion
    pool = ThreadPool(processes=3)
```

meshroom/ui/qml/ImageGallery/ImageGallery.qml
```diff
@@ -201,6 +200,19 @@ Panel {
                 }
             }
 
+            // Update grid item when corresponding thumbnail is computed
+            Connections {
+                target: ThumbnailCache
+                function onThumbnailCreated(imgSource) {
```
fabiencastan, line 206: send the index in `ThumbnailCache.thumbnail(root.source);` and emit it in the signal to retrieve it here and avoid the quadratic loop here.
```
function onThumbnailCreated(imgSource, index) {
    if(index => grid.count)
        return
    let item = grid.itemAtIndex(i);
    if(!item)
        return
    return item.updateThumbnail();
}
```

fabiencastan (changes requested):
Need to change default path for meshroom cache and some minor improvements.

meshroom/ui/components/thumbnail.py
```diff
@@ -173,22 +174,28 @@ def clean():
         # Scan thumbnail directory and gather all thumbnails to remove
         toRemove = []
         remaining = []
-        for entry in os.scandir(ThumbnailCache.thumbnailDir):
```
fabiencastan: scandir was fine to me

meshroom/ui/qml/ImageGallery/ImageGallery.qml
```diff
@@ -201,6 +200,17 @@ Panel {
...
             }
 
+            // Update grid item when corresponding thumbnail is computed
+            Connections {
+                target: ThumbnailCache
+                function onThumbnailCreated(imgSource, callerID) {
+                    let item = grid.itemAtIndex(callerID);  // item is an ImageDelegate
+                    if (item && item.source == imgSource) {
+                        item.updateThumbnail();
+                    }
```
fabiencastan, line 210: Maybe add an else with a full search of the imgSource, as if the filter is enabled the filteredIndex changed and the item is never updated.
After a night, the busyindicator is still running.

fabiencastan (changes requested):

meshroom/ui/components/thumbnail.py
```diff
@@ -0,0 +1,281 @@
...
+        if timeLimit is not None:
+            ThumbnailCache.storageTimeLimit = float(timeLimit)
+
+        # User specifed maximum number of thumbnails on disk
+        maxOnDisk = os.getenv('MESHROOM_MAX_THUMBNAILS_ON_DISK')
+        if maxOnDisk is not None:
+            ThumbnailCache.maxThumbnailsOnDisk = int(maxOnDisk)
+
+        # Clean thumbnail directory
+        ThumbnailCache.clean()
```
fabiencastan, line 85: Could take large amount of time. Should be done async.

fabiencastan (changes requested):

PR #1862, opened 2023-01-05
[ui] Display nodes computed in another Meshroom instance as "Computed Externally"

cbentejac:
## Description

This PR contains a bugfix and a minor UI improvement:
- Bugfix: a node used to be considered as computed externally if and only if its first chunk's execution mode was external. This is true in most cases but it does not cover the cases of nodes that are computed both locally and externally: if computations are started locally, interrupted and resumed externally, any chunk that completed its computations locally will have its execution mode set as such. When resuming the computations externally, the first chunk will still indicate that its execution mode is local, so the whole node will be treated as computed locally.
A node is now considered as computed externally if at least one of its chunks' execution mode is external.
- The "Computed Externally" icon used to be display on a node only when it was computing on the render farm. It is now also displayed when a node is computing in another instance of Meshroom.

## Features list

- [x] Fix the method determining whether a node is submitted externally: a node is submitted externally if at least one of its chunks' execution mode is external
- [x] Display the "Computed Externally" icon on a node if it is being computed in another instance of Meshroom (or on a render farm)

### DIFF
--- a/meshroom/core/node.py
+++ b/meshroom/core/node.py
@@ -1085,6 +1085,7 @@ class BaseNode(BaseObject):
 
     globalExecModeChanged = Signal()
     globalExecMode = Property(str, globalExecMode.fget, notify=globalExecModeChanged)
+    isExternal = Property(bool, isExtern, notify=globalExecModeChanged)
     isComputed = Property(bool, _isComputed, notify=globalStatusChanged)
     aliveChanged = Signal()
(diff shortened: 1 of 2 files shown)


Garoli (commented):
Tested only for the case with 2 meshroom instances, it works. :+1:

fabiencastan (approved):

PR #1864, opened 2023-01-09
[ui] Use the location of the most recently imported images as the base folder for the "Import Images" dialog

cbentejac:
## Description

Since #1778, all the actions from the "File" menu that open dialogs use the folder of the .mg file that was opened last as the base folder. 

This is also the case for the "Import Images" dialog, which means that at each import, the dialog opens in the folder containing the most recently opened .mg file rather than the most recently imported images. 

This PR changes that behaviour by saving the location of the last imported images, and using it as the base folder specifically for the "Import Images" dialog. The new behaviour can be summed up as follows:
- The last 3 locations of the imported images are saved across Meshroom sessions and used as a base folder (if the most recent location is not valid anymore, it is dropped and the second most recent location is used, and so on until either the list is empty or a valid location in it has been found);
- If no location has been saved when Meshroom is started (or if they are all invalid) and a graph with images has already been loaded, then the base folder will be the location of the graph's images;
- If no location has been saved when Meshroom is started (or if they are all invalid), and if the "Import Images" is the first dialog to be opened (or the already loaded graph has no image), then the same base folder as the other file dialogs is used.

The base folder for all the other file dialogs ("Open", "Save As", "Import Project", "Save As Template") remains unchanged. 

## Features list

- [x] Save the locations of the most recently imported images across sessions
- [x] Use the saved locations of the most recently imported images as the "Import Images" dialog's base folder

### DIFF
--- a/meshroom/ui/qml/ImageGallery/ImageGallery.qml
+++ b/meshroom/ui/qml/ImageGallery/ImageGallery.qml
@@ -23,6 +23,7 @@ Panel {
     readonly property string currentItemSource: grid.currentItem ? grid.currentItem.source : ""
     readonly property var currentItemMetadata: grid.currentItem ? grid.currentItem.metadata : undefined
     readonly property int centerViewId: (_reconstruction && _reconstruction.sfmTransform) ? parseInt(_reconstruction.sfmTransform.attribute("transformation").value) : 0
+    readonly property alias galleryGrid: grid
 
     property int defaultCellSize: 160
     property bool readOnly: false
--- a/meshroom/ui/qml/main.qml
+++ b/meshroom/ui/qml/main.qml
@@ -46,7 +46,18 @@ ApplicationWindow {
     SystemPalette { id: activePalette }
     SystemPalette { id: disabledPalette; colorGroup: SystemPalette.Disabled }
 
-    property url imagesFolder: ""
+    property url imagesFolder: {
+        var recentImportedImagesFolders = MeshroomApp.recentImportedImagesFolders
+        if (recentImportedImagesFolders.length > 0) {
+            for (var i = 0; i < recentImportedImagesFolders.length; i++) {
+                if (Filepath.exists(recentImportedImagesFolders[i]))
+                    return Filepath.stringToUrl(recentImportedImagesFolders[i])
+                else
+                    MeshroomApp.removeRecentImportedImagesFolder(Filepath.stringToUrl(recentImportedImagesFolders[i]))
+            }
+        }
+        return ""
+    }
 
     Settings {
         id: settings_General
@@ -331,6 +342,7 @@ ApplicationWindow {
         onAccepted: {
             _reconstruction.importImagesUrls(importImagesDialog.fileUrls)
             imagesFolder = Filepath.dirname(importImagesDialog.fileUrls[0])
+            MeshroomApp.addRecentImportedImagesFolder(imagesFolder)
         }
     }
 
@@ -480,6 +492,7 @@ ApplicationWindow {
 
     function initFileDialogFolder(dialog, importImages = false) {
         let folder = "";
+
         if (_reconstruction.graph && _reconstruction.graph.filepath) {
             folder = Filepath.stringToUrl(Filepath.dirname(_reconstruction.graph.filepath));
             if (imagesFolder.toString() === "" && workspaceView.imageGallery.galleryGrid.itemAtIndex(0) !== null) {
--- a/meshroom/ui/qml/main.qml
+++ b/meshroom/ui/qml/main.qml
@@ -482,6 +482,9 @@ ApplicationWindow {
         let folder = "";
         if (_reconstruction.graph && _reconstruction.graph.filepath) {
             folder = Filepath.stringToUrl(Filepath.dirname(_reconstruction.graph.filepath));
+            if (imagesFolder.toString() === "" && workspaceView.imageGallery.galleryGrid.itemAtIndex(0) !== null) {
+                imagesFolder = Filepath.dirname(workspaceView.imageGallery.galleryGrid.itemAtIndex(0).source);
+            }
         } else {
             var projects = MeshroomApp.recentProjectFiles;
             if (projects.length > 0 && Filepath.exists(projects[0])) {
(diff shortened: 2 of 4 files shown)


Garoli (approved):
This works well ! 
Maybe we could add something similar when a project is loaded ? For the moment the path is still the one of the .mg file. It could be the path of the first or last image loaded ?

meshroom/ui/qml/main.qml
```diff
@@ -46,6 +46,8 @@ ApplicationWindow {
     SystemPalette { id: activePalette }
     SystemPalette { id: disabledPalette; colorGroup: SystemPalette.Disabled }
 
+    property string imagesFolder: ""
```
mugulmd, line 49: Should use a `url` property instead of `string` since it represents a location on the file system

meshroom/ui/qml/main.qml
```diff
@@ -475,15 +478,22 @@ ApplicationWindow {
...
         } else {
             var projects = MeshroomApp.recentProjectFiles;
             if (projects.length > 0 && Filepath.exists(projects[0])) {
-                dialog.folder = Filepath.stringToUrl(Filepath.dirname(projects[0]));
+                folder = Filepath.stringToUrl(Filepath.dirname(projects[0]));
             }
         }
+
+        if (importImages) {
+            folder = imagesFolder === "" ? folder : imagesFolder
```
mugulmd, line 493: Should check with `Filepath.exists` or something similar instead to ensure the dialog folder is properly initialized

meshroom/ui/qml/main.qml
cbentejac: Done.

mugulmd:
Could also be interesting to make this feature work in-between sessions

meshroom/ui/qml/main.qml
cbentejac: Done.

cbentejac:
> Could also be interesting to make this feature work in-between sessions

Done, the folders are now saved across sessions! 

> This works well ! Maybe we could add something similar when a project is loaded ? For the moment the path is still the one of the .mg file. It could be the path of the first or last image loaded ?

Also done. If no folder has been saved yet (or if the settings have all been reset), and the user opens a .mg file that contains images before trying to import images, the base folder for the "Import Images" dialog will be the location of the project's images.

meshroom/ui/qml/main.qml
```diff
@@ -475,15 +490,26 @@ ApplicationWindow {
...
-    function initFileDialogFolder(dialog) {
-        if(_reconstruction.graph && _reconstruction.graph.filepath) {
-            dialog.folder = Filepath.stringToUrl(Filepath.dirname(_reconstruction.graph.filepath));
+    function initFileDialogFolder(dialog, importImages = false) {
+        let folder = "";
+
+        if (_reconstruction.graph && _reconstruction.graph.filepath) {
+            folder = Filepath.stringToUrl(Filepath.dirname(_reconstruction.graph.filepath));
+            if (imagesFolder.toString() === "" && workspaceView.imageGallery.grid.itemAtIndex(0) !== null) {
+                imagesFolder = Filepath.stringToUrl(Filepath.dirname(workspaceView.imageGallery.grid.itemAtIndex(0).source));
```
fabiencastan, line 499: Why do we check the path of the first input image only if we have saved the current scene?

meshroom/ui/qml/main.qml
```diff
@@ -475,15 +490,26 @@ ApplicationWindow {
...
 
-    function initFileDialogFolder(dialog) {
-        if(_reconstruction.graph && _reconstruction.graph.filepath) {
-            dialog.folder = Filepath.stringToUrl(Filepath.dirname(_reconstruction.graph.filepath));
+    function initFileDialogFolder(dialog, importImages = false) {
+        let folder = "";
+
+        if (_reconstruction.graph && _reconstruction.graph.filepath) {
+            folder = Filepath.stringToUrl(Filepath.dirname(_reconstruction.graph.filepath));
+            if (imagesFolder.toString() === "" && workspaceView.imageGallery.grid.itemAtIndex(0) !== null) {
```
fabiencastan, line 498: We check `imagesFolder.toString() === ""` but it has not been initialized before.

meshroom/ui/qml/ImageGallery/ImageGallery.qml
```diff
@@ -23,6 +23,7 @@ Panel {
     readonly property string currentItemSource: grid.currentItem ? grid.currentItem.source : ""
     readonly property var currentItemMetadata: grid.currentItem ? grid.currentItem.metadata : undefined
     readonly property int centerViewId: (_reconstruction && _reconstruction.sfmTransform) ? parseInt(_reconstruction.sfmTransform.attribute("transformation").value) : 0
+    readonly property GridView grid: grid
```
fabiencastan, line 26: grid = grid?

fabiencastan: maybe better to use a different name for each var.
If it's to expose it externally, maybe use "property alias".

meshroom/ui/qml/main.qml
fabiencastan, line 498: check that the grid is not empty before accessing the first element!

fabiencastan (changes requested):
Need some modifications

meshroom/ui/qml/ImageGallery/ImageGallery.qml
cbentejac: Done.

meshroom/ui/qml/main.qml
cbentejac: In case `imagesFolder` was empty because the user had not imported any image yet and a project with images had been opened, the idea was to use the location of these images to set `imagesFolder` rather than waiting for the user to do a first "Import Images".

Waiting to have a _reconstruction.graph.filepath was initially done because we didn't want to use the location of dropped images as the base folder: it could only be "actively" set by using the "Import Images" action , or "passively" by opening a project file with images (and so dropping images in the gallery would never enter the `if (_reconstruction.graph.filepath)` condition). 

This makes less sense now that the locations are saved across sessions. Since the number of cases for which `imagesFolder` is not set is virtually 0 now (it will happen once the first time a user starts Meshroom, and never again unless the settings are reset, all 3 saved locations have become invalid or the user never uses the "Import Images" action), it might be better to move it outside of the `if`. That way we prevent as much as possible opening the dialog on Meshroom's install location for the rare cases in which `imagesFolder` is not set. 

The location will still be only saved across sessions if the "Import Images" dialog is used.

fabiencastan (approved):

PR #1865, opened 2023-01-10
[ui] GraphEditor: use maxZoom to fit on nodes

Garoli:
## Description
Put the scale to 1 when fitting the graph to prevent huge node/text when there are only a few nodes. 

## Features list
- [X] Scale to 1 when few nodes on graph display

### DIFF
--- a/meshroom/ui/qml/GraphEditor/GraphEditor.qml
+++ b/meshroom/ui/qml/GraphEditor/GraphEditor.qml
@@ -14,6 +14,7 @@ Item {
     property variant uigraph: null  /// Meshroom ui graph (UIGraph)
     readonly property variant graph: uigraph ? uigraph.graph : null  /// core graph contained in ui graph
     property variant nodeTypesModel: null  /// the list of node types that can be instantiated
+    property real maxZoom: 2.0
 
     property var edgeAboutToBeRemoved: undefined
 
@@ -152,7 +153,6 @@ Item {
         anchors.fill: parent
         property double factor: 1.15
         property real minZoom: 0.1
-        property real maxZoom: 2.0
         // Activate multisampling for edges antialiasing
         layer.enabled: true
         layer.samples: 8
@@ -800,7 +800,7 @@ Item {
         // compute bounding box
         var bbox = boundingBox()
         // rescale to fit the bounding box in the view, max zoom is limited to 120% to prevent huge text
-        draggable.scale = Math.min(Math.min(root.width/bbox.width, root.height/bbox.height),1.2)
+        draggable.scale = Math.min(Math.min(root.width/bbox.width, root.height/bbox.height),maxZoom)
         // recenter
         draggable.x = bbox.x*draggable.scale*-1 + (root.width-bbox.width*draggable.scale)*0.5
         draggable.y = bbox.y*draggable.scale*-1 + (root.height-bbox.height*draggable.scale)*0.5
--- a/meshroom/ui/qml/GraphEditor/GraphEditor.qml
+++ b/meshroom/ui/qml/GraphEditor/GraphEditor.qml
@@ -799,7 +799,7 @@ Item {
     function fit() {
         // compute bounding box
         var bbox = boundingBox()
-        // rescale to fit the bounding box in the view, max zoom is limited to 120% to prevent huge text
+        // rescale to fit the bounding box in the view, zoom is limited to prevent huge text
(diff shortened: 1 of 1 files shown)


meshroom/ui/qml/GraphEditor/GraphEditor.qml
```diff
@@ -799,8 +799,8 @@ Item {
     function fit() {
         // compute bounding box
         var bbox = boundingBox()
-        // rescale
-        draggable.scale = Math.min(root.width/bbox.width, root.height/bbox.height)
+        // rescale, if low number of node, zomm 100% to prevent huge text
+        draggable.scale = nodeRepeater.count < 4 ? 1 :  Math.min(root.width/bbox.width, root.height/bbox.height)
```
mugulmd, line 803: Not sure a threshold on the number of nodes is the right approach here as it doesn't take the graph layout into account and can lead to unexpected situations when we are right below/above the threshold.

Maybe try imposing a limit on the min/max scale instead ?

fabiencastan: zoom

Garoli: > Not sure a threshold on the number of nodes is the right approach here as it doesn't take the graph layout into account and can lead to unexpected situations when we are right below/above the threshold.
> 
> Maybe try imposing a limit on the min/max scale instead ?

Done, the max is now 120% !

> zoom

Done :face_with_spiral_eyes:

Garoli:
> Not sure a threshold on the number of nodes is the right approach here as it doesn't take the graph layout into account and can lead to unexpected situations when we are right below/above the threshold.
> 
> Maybe try imposing a limit on the min/max scale instead ?

Done, the max is now 120% !

> zoom

Done :face_with_spiral_eyes:

meshroom/ui/qml/GraphEditor/GraphEditor.qml
```diff
@@ -799,8 +799,8 @@ Item {
     function fit() {
         // compute bounding box
         var bbox = boundingBox()
-        // rescale
-        draggable.scale = Math.min(root.width/bbox.width, root.height/bbox.height)
+        // rescale, if low number of node, zoom 100% to prevent huge text
```
fabiencastan, line 802:
```suggestion
        // rescale to fit the bounding box in the view, max zoom is limited to 120% to prevent huge text
```

fabiencastan (approved):

fabiencastan (changes requested):
In fact there is already a maxZoom defined that needs to be moved in a global var of the GraphEditor and used in both places.

Garoli:
> In fact there is already a maxZoom defined that needs to be moved in a global var of the GraphEditor and used in both places.

Done

meshroom/ui/qml/GraphEditor/GraphEditor.qml
```diff
@@ -14,6 +14,7 @@ Item {
     property variant uigraph: null  /// Meshroom ui graph (UIGraph)
     readonly property variant graph: uigraph ? uigraph.graph : null  /// core graph contained in ui graph
     property variant nodeTypesModel: null  /// the list of node types that can be instantiated
+    property real maxZoom: 2.0
```
fabiencastan, line 17: keep min/maxZoom together

Garoli: Done

meshroom/ui/qml/GraphEditor/GraphEditor.qml
```diff
@@ -799,8 +799,8 @@ Item {
     function fit() {
         // compute bounding box
         var bbox = boundingBox()
-        // rescale
-        draggable.scale = Math.min(root.width/bbox.width, root.height/bbox.height)
+        // rescale to fit the bounding box in the view, max zoom is limited to 120% to prevent huge text
```
cbentejac, line 802: The comment needs to be updated since `maxZoom = 2.0`.

cbentejac (changes requested):
Tested and works fine, a minor comment modification is just needed.

meshroom/ui/qml/GraphEditor/GraphEditor.qml
fabiencastan, line 802:
```suggestion
        // rescale to fit the bounding box in the view, zoom is limited to prevent huge text
```

fabiencastan (approved):

PR #1867, opened 2023-01-11
[ui] node time computation and chunks count in node editor header

Garoli:
## Description
Improvement of computation information in node editor header

## Features list
[X] update computation time every 2.5 seconds in header
[X] computation time when error occurs 
[X] chunk computed counter in node editor header

## Implementation remarks
The previous label text logic is moved to the nodeChanged signal because the logic is overwritten by the timer when setting the text time. 
I don't know if 2.5 seconds is the best choice, feel free to give other refresh timing ideas.

### DIFF
--- a/meshroom/core/node.py
+++ b/meshroom/core/node.py
@@ -411,11 +411,12 @@ class NodeChunk(BaseObject):
         try:
             self.node.nodeDesc.processChunk(self)
         except Exception as e:
+            self._status.elapsedTime = time.time() - startTime
             if self._status.status != Status.STOPPED:
-                self._status.elapsedTime = time.time() - startTime
                 self.upgradeStatusTo(Status.ERROR)
             raise
         except (KeyboardInterrupt, SystemError, GeneratorExit) as e:
+            self._status.elapsedTime = time.time() - startTime
             self.upgradeStatusTo(Status.STOPPED)
             raise
         finally:
@@ -806,6 +807,14 @@ class BaseNode(BaseObject):
             shutil.rmtree(self.internalFolder)
             self.updateStatusFromCache()
 
+    @Slot(result=str)
+    def getStartDateTime(self):
+        """ Return the date (str) of the first running chunk """
+        if not self.isAlreadySubmittedOrFinished() or len(self._chunks) == 0:
+            return ""
+        dateTime = [chunk._status.startDateTime for chunk in self._chunks if chunk._status.startDateTime != ""]
+        return min(dateTime) if len(dateTime) != 0 else ""
+
     def isAlreadySubmitted(self):
         for chunk in self._chunks:
             if chunk.isAlreadySubmitted():
@@ -828,13 +837,10 @@ class BaseNode(BaseObject):
                 return True
         return False
 
-    @Slot(result=str)
-    def getFirstChunkRunning(self):
-        """ Return the date (str) of the first running chunk """
-        if not self.isAlreadySubmittedOrFinished() or len(self._chunks) == 0:
-            return ""
-        dateTime = [chunk._status.startDateTime for chunk in self._chunks if chunk._status.startDateTime != ""]
-        return min(dateTime) if len(dateTime) != 0 else ""
+    @Slot(result=bool)
+    def isRunning(self):
+        """ Return True if at least one chunk of this Node is running, False otherwise. """
+        return any(chunk.isRunning() for chunk in self._chunks)
 
     @Slot(result=bool)
     def isFinishedOrRunning(self):
--- a/meshroom/core/node.py
+++ b/meshroom/core/node.py
@@ -412,6 +412,7 @@ class NodeChunk(BaseObject):
             self.node.nodeDesc.processChunk(self)
         except Exception as e:
             if self._status.status != Status.STOPPED:
+                self._status.elapsedTime = time.time() - startTime
                 self.upgradeStatusTo(Status.ERROR)
             raise
         except (KeyboardInterrupt, SystemError, GeneratorExit) as e:
--- a/meshroom/core/node.py
+++ b/meshroom/core/node.py
@@ -404,6 +404,7 @@ class NodeChunk(BaseObject):
         global runningProcesses
         runningProcesses[self.name] = self
         self._status.initStartCompute()
+        exceptionStatus = None
         startTime = time.time()
         self.upgradeStatusTo(Status.RUNNING)
         self.statThread = stats.StatisticsThread(self)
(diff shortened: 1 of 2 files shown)


meshroom/core/node.py
```diff
@@ -750,6 +750,14 @@ def isSubmittedOrRunning(self):
                 return True
         return False
 
+    @Slot(result=str)
+    def getFirstChunkRunning(self):
```
fabiencastan, line 755: node.getFirstChunkRunning() => the name seems to indicate that it will return a chunk
Could be replaced by:
node.getStartDateTime() => the notion of chunk is not mentioned, but we can consider it an implementation detail

meshroom/ui/qml/GraphEditor/NodeEditor.qml
```diff
@@ -17,30 +17,42 @@ Panel {
...
 
     signal attributeDoubleClicked(var mouse, var attribute)
     signal upgradeRequest()
 
     title: "Node" + (node !== null ? " - <b>" + node.label + "</b>" : "")
     icon: MaterialLabel { text: MaterialIcons.tune }
 
+    onNodeChanged: {
+        nodeStartDateTime=""
+        if(node !== null && node.isSubmittedOrRunning()){
```
fabiencastan, line 30: only when running (submitted has no impact here)

meshroom/ui/qml/GraphEditor/NodeEditor.qml
```diff
@@ -17,30 +17,42 @@ Panel {
...
+            timer.start()
+        }
+        else if (node !== null && node.isFinishedOrRunning()) {
+            computationInfo.text = Format.getTimeStr(node.elapsedTime)
```
fabiencastan, line 34: Seems to be the right thing to do here.
But I'm not sure if elapsedTime is correctly updated when the node is running for now.

meshroom/ui/qml/GraphEditor/NodeEditor.qml
```diff
@@ -17,30 +17,42 @@ Panel {
...
-                        return "Running for: " + Format.getTimeStr(node.elapsedTime)
-                    } else {
-                        return (node.chunks.count > 1) ? "First chunk running" : "Node running"
+            id : computationInfo
+            Timer{
+                id:timer
+                interval: 2500
+                triggeredOnStart: true
+                repeat: true
+                running: node !== null && node.isSubmittedOrRunning()
```
fabiencastan, line 49: node.isRunning()

meshroom/ui/qml/GraphEditor/NodeEditor.qml
```diff
@@ -118,7 +130,10 @@ Panel {
                 MenuItem {
                     enabled: root.node !== null
                     text: "Clear Pending Status"
-                    onClicked: node.clearSubmittedChunks()
+                    onClicked: {
+                        node.clearSubmittedChunks()
+                        timer.stop()
```
fabiencastan, line 146: the change of "running" status should already stop it

meshroom/ui/qml/GraphEditor/NodeEditor.qml
```diff
@@ -50,9 +50,20 @@ Panel {
...
                     parent.text=Format.getTimeStr((now-nodeStartDateTime)/1000)
+
+                    var chunkCompletion=0
+                    if( node.chunks.count>1){
+                        for (var i = 0; i < node.chunks.count; i++) {
+                            if(node.chunks.at(i).statusName == "SUCCESS"){
+                                chunkCompletion++
+                            }
+                        }
+                        parent.text+= " | "+ chunkCompletion + "/" + node.chunks.count + " chunks"
```
fabiencastan, line 64: avoid +=
Setup an intermediate variable with the "chunk completion" string that is always added but can be empty.

meshroom/core/node.py
```diff
@@ -411,6 +411,7 @@ def process(self, forceCompute=False):
             self.node.nodeDesc.processChunk(self)
         except Exception as e:
             if self._status.status != Status.STOPPED:
+                self._status.elapsedTime = time.time() - startTime
```
fabiencastan, line 414: move out of the if statement
and add also in the other except

fabiencastan (changes requested):
need some modifications

meshroom/core/node.py
```diff
@@ -728,6 +730,14 @@ def clearData(self):
             shutil.rmtree(self.internalFolder)
             self.updateStatusFromCache()
 
+    @Slot(result=str)
+    def getStartDateTime(self):
+        """ Return the date (str) of the first running chunk """
+        if not self.isAlreadySubmittedOrFinished() or len(self._chunks) == 0:
```
fabiencastan, line 736: Submitted has no start time.
Status that can have a start time:
``` 
 43 class Status(Enum):
  44     """
  45     """
  46     NONE = 0 => no
  47     SUBMITTED => no
  48     RUNNING => yes
  49     ERROR => yes
  50     STOPPED => yes
  51     KILLED => yes
  52     SUCCESS => yes
```

fabiencastan: Maybe remove the status check on the node itself and add it on the loop over chunks.

meshroom/core/node.py
```diff
@@ -728,6 +730,14 @@ def clearData(self):
...
+            return ""
+        dateTime = [chunk._status.startDateTime for chunk in self._chunks if chunk._status.startDateTime != ""]
```
fabiencastan, line 738: Something like:
```suggestion
        dateTime = [chunk._status.startDateTime for chunk in self._chunks if chunk._status.status not in (Status.NONE, Status.SUBMITTED) and chunk._status.startDateTime != ""]
```

meshroom/core/node.py
```diff
@@ -728,6 +730,14 @@ def clearData(self):
             shutil.rmtree(self.internalFolder)
             self.updateStatusFromCache()
 
+    @Slot(result=str)
+    def getStartDateTime(self):
+        """ Return the date (str) of the first running chunk """
+        if not self.isAlreadySubmittedOrFinished() or len(self._chunks) == 0:
+            return ""
```
fabiencastan, line 737: len(dateTime) already checked at the end
```suggestion
```

fabiencastan (dismissed):
If we change the selected node and the status is the same, the information is not updated (and we display the timing info of another node).

cbentejac (changes requested):
If you select a node, start its computations, and wait without unselecting it, the timer never stops if the status is updated to `ERROR`: the displayed computation time keeps on increasing as the timer keeps on being triggered.
If you unselect the node, the timer is immediately stopped, and if you reselect it, the displayed computation time is correct.

I don't know where it happens, but it seems like the `globalStatusChanged` signal is not emitted when the node's status changes to `ERROR`.

On another note, since we compute the elapsedTime when we stop the computations (cf the `except` clauses in node.py), couldn't we display it in that case as well? (it is an open question)

fabiencastan (approved):

PR #1876, opened 2023-01-23
[sfmtransform] Transformation parameter availability

servantftechnicolor:
Transformation parameter is available when auto_from_cameras_x_axis is selected in sfmTransform.
This way, the user can select which view is the center in the panorama.


cbentejac (approved):
Tested with https://github.com/alicevision/AliceVision/pull/1334.

PR #1888, opened 2023-02-01
[ui] fix "Sync Camera with Image Selection"

mugulmd:
## Description

This PR fixes the "Sync Camera with Image Selection" feature in the 3D viewer.

## Features

- use the correct formulas for the camera rotation and translation
- apply exif orientation tag of input image to 3D viewer in order to match 2D viewer.

## Implementation remarks

The image used as overlay is not exactly the input image but its undistorted version computed by the PrepareDenseScene node. This means that if the PrepareDenseScene node has not been computed or does not exist (for example in a CameraTracking pipeline), there won't be a background image.
A more robust solution should be imagined for handling such cases.

### DIFF
--- a/meshroom/ui/qml/Controls/ExifOrientedViewer.qml
+++ b/meshroom/ui/qml/Controls/ExifOrientedViewer.qml
@@ -1,8 +1,9 @@
 import QtQuick 2.11
+import Utils 1.0
 
 /**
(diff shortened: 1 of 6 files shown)


cbentejac (approved):
Tested on images with orientation 1 and 6, and everything seemed to work well!

fabiencastan (approved):

PR #1893, opened 2023-02-13
[ui] fix thumbnail cache bugs

mugulmd:
## Description

In this PR we fix several bugs related to the thumbnail cache mechanism.

### Endless request loop

We fix a problem that can occur with the thumbnail cache when input images cannot be read: 
since each image delegate has a Timer that will send a new request every 5 seconds until a thumbnail can be loaded, in the case of images that cannot be read these requests will be sent endlessly.

The proposed fix is to stop the Timer after the 1st request, since it is the only one that is actually useful for preventing unwanted behaviour from the QML engine.

### Reload image after modification

Another problem that we fixed is that when an image was loaded once, then modified for any reason and then reloaded later on, its corresponding thumbnail will not be updated.

The proposed fix is to remove a thumbnail if its corresponding image has been modified after the thumbnail creation: this way, a new thumbnail will be generated, and it will match the modified image.

### DIFF
--- a/meshroom/ui/qml/ImageGallery/ImageDelegate.qml
+++ b/meshroom/ui/qml/ImageGallery/ImageDelegate.qml
@@ -41,18 +41,14 @@ Item {
         updateThumbnail();
     }
 
-    // Send a new request every 5 seconds until thumbnail is loaded
+    // Send a new request after 5 seconds if thumbnail is not loaded
     // This is meant to avoid deadlocks in case there is a synchronization problem
     Timer {
         interval: 5000
-        repeat: true
         running: true
         onTriggered: {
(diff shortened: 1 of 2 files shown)


meshroom/ui/qml/ImageGallery/ImageDelegate.qml
```diff
@@ -41,19 +41,15 @@ Item {
         updateThumbnail();
     }
 
-    // Send a new request every 5 seconds until thumbnail is loaded
+    // Send a new request after 5 seconds if thumbnail is loaded
```
fabiencastan, line 44: is not loaded

fabiencastan (approved):

PR #1897, opened 2023-02-15
[ui] Improve "Clear Images" action's behaviour and performance

cbentejac:
## Description

This PR modifies the behaviour of the "Clear Images" menu action as follows:
- It is now executed on the Python side instead of the QML side, which makes it more than 4x faster:
   - For 500 PNG images, it used to take ~38s, it now takes less than 0.5s.
   - For 1000 PNG images, it used to take ~2m39, it now takes 1s.
- "Clear Images" now only clears the images for the active group (meaning the active CameraInit node), instead of clearing all the images for all groups (all CameraInit nodes).
- When using "Clear Images", the "Undo" stack used to be filled with individual intrinsics and viewpoints images. Calling "Clear Images" when 500 images were in the gallery was thus requiring at least 500 calls to "Undo" to be fully reverted. It is now considered as a single action, meaning a single call to "Undo" is enough to recover all the cleared images.

A "Clear All Images" action has been added in the "Advanced" sub-menu to clear the images of all the groups.

<div align="center"></div>

Tooltips have been added over both "Clear Images" and "Clear All Images" to make the distinction clear for users. 

## Features list

- [x] Execute "Clear Images" on the Python side rather than the QML side;
- [x] Make "Clear Images" remove only the images of the current CameraInit node;
- [x] Add a "Clear All Images" action in the "Advanced" sub-menu to remove the images of all the CameraInit nodes.

### DIFF
--- a/meshroom/ui/commands.py
+++ b/meshroom/ui/commands.py
@@ -350,6 +350,31 @@ class ListAttributeRemoveCommand(GraphCommand):
         listAttribute.insert(self.index, self.value)
 
 
+class ClearImagesCommand(GraphCommand):
+    def __init__(self, graph, cameraInitNodes, parent=None):
+        super(ClearImagesCommand, self).__init__(graph, parent)
+        self.cameraInits = cameraInitNodes
+        self.viewpoints = { cameraInit.name: cameraInit.attribute("viewpoints").getExportValue() for cameraInit in self.cameraInits }
+        self.intrinsics = { cameraInit.name: cameraInit.attribute("intrinsics").getExportValue() for cameraInit in self.cameraInits }
+        self.title = "Clear{}Images".format(" " if len(self.cameraInits) == 1 else " All ")
+        self.setText(self.title)
+
+    def redoImpl(self):
+        for i in range(len(self.cameraInits)):
+            # Reset viewpoints
+            self.cameraInits[i].viewpoints.resetValue()
+            self.cameraInits[i].viewpoints.valueChanged.emit()
+            # Reset intrinsics
+            self.cameraInits[i].intrinsics.resetValue()
+            self.cameraInits[i].intrinsics.valueChanged.emit()
+
+    def undoImpl(self):
+        for cameraInit in self.viewpoints:
+            with GraphModification(self.graph):
+                self.graph.node(cameraInit).viewpoints.value = self.viewpoints[cameraInit]
+                self.graph.node(cameraInit).intrinsics.value = self.intrinsics[cameraInit]
+
+
 class MoveNodeCommand(GraphCommand):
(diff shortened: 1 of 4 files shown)


mugulmd:
I'm not sure we should be using "Ctrl-D" and "Ctrl-Shift-D" for clearing images since the letter D is usually reserved for duplicating. If that doesn't create any conflict I would rather use the letter X instead, what do you think ?

fabiencastan:
We don't need to do that very often, so I think we can just remove the shortcut.

cbentejac:
> We don't need to do that very often, so I think we can just remove the shortcut.

I dropped the commit that added the shortcuts.

meshroom/ui/reconstruction.py
```diff
@@ -694,6 +694,33 @@ def allViewIds(self):
...
+    @Slot()
+    def clearAllImages(self):
+        """ Clear the list of viewpoints and intrinsics for all the CameraInit nodes. """
+        with self.groupedGraphModification("Clear All Images"):
+            for cameraInit in self._cameraInits:
+                self.deleteIntrinsics(cameraInit.attribute("viewpoints"), cameraInit.attribute("intrinsics"))
+
+    def deleteIntrinsics(self, viewpoints, intrinsics):
+        """ Delete a set of provided viewpoints and intrinsics. """
+        # Delete all viewpoints
```
fabiencastan, line 712: Maybe better to use "resetValue" https://github.com/alicevision/Meshroom/blob/develop/meshroom/core/attribute.py#L358

cbentejac: Done (and infinitely better!).

meshroom/ui/commands.py
```diff
@@ -350,6 +350,31 @@ def undoImpl(self):
...
+            self.cameraInits[i].viewpoints.resetValue()
+            self.cameraInits[i].viewpoints.valueChanged.emit()
+            # Reset intrinsics
+            self.cameraInits[i].intrinsics.resetValue()
+            self.cameraInits[i].intrinsics.valueChanged.emit()
+
+    def undoImpl(self):
+        for i in range(len(self.cameraInits)):
+            with GraphModification(self.graph):
+                self.cameraInits[i].viewpoints.value = self.viewpoints[i]
```
fabiencastan, line 374: Sounds dangerous to rely on the ordering of cameraInit nodes.
It would be more straightforward to use a map with the unique name of the node as key.

fabiencastan (changes requested):

meshroom/ui/commands.py
cbentejac: Done.

fabiencastan (approved):

PR #1903, opened 2023-02-22
Fix missing Publish nodes in templates

### LINKED ISSUE #1900: [question] Got error when run meshroom_batch
Environment：Ubuntu20.04， build AliceVision and meshroom， try to run in the command line.
The command：python3 bin/meshroom_batch -i ./dataset_monstree/mini3/ --toNode DepthMap_1 -o ./test_result/，it got error：
RuntimeError: meshroom_batch requires a pipeline graph with exactly one 'Publish' node, 0 found.

I have no idea where is the "Publish" node and I searched the issues but no related one.
Hope for reply~

hammady:
Closes #1900.

## Description

This PR fixes an issue in `meshroom_batch` which was caused by the refactoring of hard-coded pipelines to .mg files. Specifically, this commit: 8fb0c778d12af5d3d51567407a6d38132e53d796 omitted the `Publish` node from the photogrammetry pipeline. This PR adds it back. Before this PR, if the `--output` argument was passed to `meshroom_batch`, it throws an error complaining from no Publish nodes as explained in the linked issue.

### DIFF
--- a/bin/meshroom_batch
+++ b/bin/meshroom_batch
@@ -91,15 +91,6 @@ if args.verbose:
     logging.getLogger().setLevel(logStringToPython[args.verbose])
 
 
-def getOnlyNodeOfType(g, nodeType):
-    """ Helper function to get a node of 'nodeType' in the graph 'g' and raise if no or multiple candidates. """
-    nodes = g.nodesOfType(nodeType)
-    if len(nodes) != 1:
-        raise RuntimeError("meshroom_batch requires a pipeline graph with exactly one '{}' node, {} found."
-                           .format(nodeType, len(nodes)))
-    return nodes[0]
-
-
 def getInitNode(g):
     """
     Helper function to get the Init node in the graph 'g' and raise an exception if there is no or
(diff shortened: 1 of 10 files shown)


fabiencastan:
Hi,
Thanks for your contribution!
We need to have the Publish node in the templates, as you are proposing, to have it available in the batch command.
@cbentejac But we probably we need to filter it out when we create the pipeline in the UI.

cbentejac:
@fabiencastan On it!

Hi @hammady, thank you for identifying and fixing the issue!
If that is alright with you, I'll push the modifications we need to filter out the "Publish" nodes from templates in the UI directly to your branch, on top of your commit. That way, this PR will fix #1900 without impacting the way templates are currently used in the UI (contrary to `meshroom_batch`, "Publish" nodes are neither needed nor useful when using the default pipelines in the UI, because all the nodes' outputs are already easy to access).

hammady:
@cbenteja please go ahead!

cbentejac:
Hi,

I have pushed two commits:
- The first one adds a `Publish` node to all the other existing templates, so they can all be used with `meshroom_batch`.
- The second one adds an optional parameter in the graph loading method to determine whether `Publish` nodes from templates should be ignored during the graph's construction (by default, they will be). In `meshroom_batch`, that option is set depending on whether the `--output` argument is set: if it is then the `Publish` node is kept, otherwise it is ignored. `meshroom_batch` also does not throw a runtime error anymore if there are more than 1 `Publish` nodes. Instead, all the `Publish` nodes will push to the folder specified by `--output`, with the risk of overwriting each other if they are connected to the same type of nodes.

hammady:
Can we merge this now?

bin/meshroom_batch
```diff
@@ -140,8 +140,15 @@ with multiview.GraphModification(graph):
         graph.setVerbose(args.verbose)
 
     if args.output:
-        publish = getOnlyNodeOfType(graph, 'Publish')
```
fabiencastan, line 143: getOnlyNodeOfType is not used anymore

fabiencastan (changes requested):

bin/meshroom_batch
cbentejac: Done, `getOnlyNodeOfType` has been completely removed.

fabiencastan (approved):

PR #1915, opened 2023-03-02
[ui] Viewer2D: fix displayed metadata

mugulmd:
## Description

When double-clicking on a node with image output attributes, even though the image shown is not the view itself, the metadata used are the viewpoint's metadata, taken from the CameraInit's sfm data.
This can lead to problems, for example with the orientation metadata: if the viewpoint and the image displayed have a different orientation, the image will appear flipped and/or rotated.

However, whenever we use the FloatImage viewer, the displayed image's metadata are available (directly loaded from disk using the aliceVision::image module).
Therefore the purpose of this PR is to use in priority the metadata retrieved by the FloatImage viewer when it is available.

### DIFF
--- a/meshroom/ui/qml/Viewer/Viewer2D.qml
+++ b/meshroom/ui/qml/Viewer/Viewer2D.qml
@@ -43,6 +43,10 @@ FocusScope {
             if (floatImageViewerLoader.active) {
                 return floatImageViewerLoader.item.metadata;
             }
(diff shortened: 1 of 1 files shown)


fabiencastan (changes requested):
The CameraInit adds some information about the sensor, etc that will not be visible anymore.

The problem is more that when we load an external image, we should know that the we are not displaying the input image anymore and should not use unrelated metadata.

fabiencastan (approved):
That's great!

PR #1920, opened 2023-03-03
[ui] Add a new advanced menu action to load templates like regular projects

cbentejac:
## Description

This PR adds a new menu action, "Load Template", in the "Advanced" sub-menu of "File".

"Load Template" allows to load an .mg file as a regular project file, without taking into account if it is a template.

If the project file is not a template, it will be opened exactly as if the "Open File" menu had been used. If it is a template and it contains "Publish" nodes, they will not be filtered out (whereas they will be if the template is opened with "Open File" or through the "New" actions).

## Features list

- [x] Add a new menu action to load templates like regular project files


meshroom/ui/qml/main.qml
```diff
@@ -482,6 +494,20 @@ ApplicationWindow {
         onTriggered: graphEditor.pasteNodes()
     }
 
+    Action {
+        id: loadTemplateAction
+
+        property string tooltip: "Load a template like a regular project file (any \"Publish\" node will be displayed)"
+        text: "Load Template"
+        shortcut: "Ctrl+Shift+O"
```
fabiencastan, line 502: Maybe avoid shortcuts for so specific options.
```suggestion
```

fabiencastan (approved):

meshroom/ui/qml/main.qml
cbentejac: Done, the shortcut has been removed.

fabiencastan (approved):

PR #1921, opened 2023-03-04
Set GLTF as the default output format in Texturing

hammady:
## Description

This sets GLTF as the default output format in Texturing. It applies to the UI and built-in scripts like `meshroom_batch`.

## Implementation remarks
Note that GLTF generates an additional .bin file.
For this, I had to include .* in the output file pattern. An alternative was to add an additional output parameter to be enabled only when GLTF is selected (like the MTL file). Also note that with this .* pattern, the MTL output is redundant.

### DIFF
--- a/meshroom/nodes/aliceVision/Texturing.py
+++ b/meshroom/nodes/aliceVision/Texturing.py
@@ -74,7 +74,7 @@ Many cameras are contributing to the low frequencies and only the best ones cont
             name='outputMeshFileType',
             label='Mesh File Type',
             description='File Type',
-            value='obj',
+            value='gltf',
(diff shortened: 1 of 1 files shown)


simogasp:
I'm not against having gltf as default option because it was the missing standard and hopefully soon it will be used everywhere. That being said, there are still many programs and pipelines that for one reason or another need e.g. an obj. And everyone has her/his own needs.
It would be nice if we could set this default value through the existing config file that Meshroom uses for e.g. tracking recent projects etc. I don't know if there is an easy way to retrieve those settings and set the default with the relevant key. This could be extended to other defaults of other nodes (e.g. the type of features to extract by default etc).

PR #1924, opened 2023-03-08
[ui] Reflect changes made in QtAliceVision refactorize PR

mugulmd:
## Description

In [this QtAliceVision PR](https://github.com/alicevision/QtAliceVision/pull/31) several changes has been made to the QML types defined in the plugin.
The goal of this PR is to reflect these changes in Meshroom's QML code to keep it up to date.

### DIFF
--- a/meshroom/ui/qml/Viewer/FeaturesInfoOverlay.qml
+++ b/meshroom/ui/qml/Viewer/FeaturesInfoOverlay.qml
@@ -22,6 +22,11 @@ FloatingPane {
     ColumnLayout {
         // Header
         RowLayout {
+            // Node used to read features
+            Label {
+                text: _reconstruction ? _reconstruction.activeNodes.get("featureProvider").node.label : ""
+                Layout.fillWidth: true
+            }
             // Settings menu
             Loader {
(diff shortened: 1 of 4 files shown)


meshroom/ui/qml/Viewer/FeaturesInfoOverlay.qml
```diff
@@ -16,16 +16,12 @@ FloatingPane {
...
+    property var msfmdata: null
 
     ColumnLayout {
         // Header
         RowLayout {
-            // FeatureExtraction node name
-            Label {
-                text: featureExtractionNode ? featureExtractionNode.label : ""
-                Layout.fillWidth: true
-            }
```
cbentejac, line 28: As the name of the active `FeatureExtraction` node is not displayed anymore in the FeatureViewer, it is impossible to know what the viewer is showing when there are several branches / `FeatureExtraction` nodes. I understand that using the `FeatureExtraction` node alone is not good solution _per se_ as the Feature Viewer does not only display information coming from the `FeatureExtraction` node, but we need something to indicate what we are displaying.

cbentejac (approved):

PR #1925, opened 2023-03-09
Hide disabled File attributes and their connections

cbentejac:
## Description

This PR effectively hides `File` attributes that are disabled. If these attributes have connections, the edges representing connections are hidden but preserved. 

- When the edges are hidden, if one of the connected attributes is still enabled, its pin becomes empty as if it was not connected to anything. If the disabled attribute is re-enabled and the connection has not been broken (if the enabled attribute on the other side of the edge has not been reconnected to another attribute, for example), the edge re-appears:
<div align="center"></div>

- If the connection has been broken, then the attribute will re-appear as unconnected, as the edge has been removed:
<div align="center"></div>

- Hidden connections are saved in project files in the same fashion as regular connections: no distinction is made between visible and hidden connections, only the `enabled` status of the connected attributes will determine whether they are hidden in the GraphEditor or not. When a project file containing some hidden attributes and connections is loaded, they are correctly taken into account and hidden.

- The list of the loaded node's viewable outputs in the 2D Viewer is updated accordingly: any attribute in the list that gets disabled will be automatically removed, and any disabled attribute that gets re-enabled will be added to it.

This PR also uncomments the conditional `enabled` status of some outputs in the DepthMap node.

## Features list

- [x] Hide the disabled File attributes in the GraphEditor and their connections, but preserve them;
- [x] Update the list of viewable outputs in the 2D Viewer depending on the loaded node's enabled outputs.

### DIFF
--- a/meshroom/ui/qml/GraphEditor/GraphEditor.qml
+++ b/meshroom/ui/qml/GraphEditor/GraphEditor.qml
@@ -383,7 +383,7 @@ Item {
                     property var src: root._attributeToDelegate[edge.src]
                     property var dst: root._attributeToDelegate[edge.dst]
                     property bool isValidEdge: src !== undefined && dst !== undefined
-                    visible: isValidEdge
+                    visible: isValidEdge && src.visible && dst.visible
 
                     property bool inFocus: containsMouse || (edgeMenu.opened && edgeMenu.currentEdge === edge)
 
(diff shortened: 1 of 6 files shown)


meshroom/ui/qml/GraphEditor/GraphEditor.qml
```diff
@@ -393,7 +393,7 @@ Item {
                     property var src: root._attributeToDelegate[edge.src]
                     property var dst: root._attributeToDelegate[edge.dst]
                     property bool isValidEdge: src != undefined && dst != undefined
-                    visible: isValidEdge
+                    visible: isValidEdge && root._attributeToDelegate[edge.src].visible && root._attributeToDelegate[edge.dst].visible
```
fabiencastan, line 396: there are vars shortcuts "src" and "dst"

bmaujean (approved):

PR #1941, opened 2023-03-20
[ui] ImageGallery: Allow image drop if the active group is not computing

cbentejac:
## Description

This PR fixes the behaviour of the Image Gallery, which used to switch to read-only as soon as a node in the entire graph was being computed. This meant that no image could be dropped in the gallery if there was any ongoing computation, even if the nodes that were computed did not include or involve the active `CameraInit` node.

The computation status of the active `CameraInit` node is now taken into account when determining whether the Image Gallery should be read-only or not.

## Features list

- [x] Switch the Image Gallery to read-only if and only if the active `CameraInit` node is locked because of computations.

### DIFF
--- a/meshroom/ui/qml/ImageGallery/ImageGallery.qml
+++ b/meshroom/ui/qml/ImageGallery/ImageGallery.qml
@@ -39,7 +39,7 @@ Panel {
         property variant currentCameraInit: _reconstruction && _reconstruction.tempCameraInit ? _reconstruction.tempCameraInit : root.cameraInit
         property variant viewpoints: currentCameraInit ? currentCameraInit.attribute('viewpoints').value : undefined
         property variant intrinsics: currentCameraInit ? currentCameraInit.attribute('intrinsics').value : undefined
-        property bool readOnly: root.readOnly || displayHDR.checked
+        property bool readOnly: ((_reconstruction && currentCameraInit) ? currentCameraInit.locked : root.readOnly) || displayHDR.checked
 
         onViewpointsChanged: {
             ThumbnailCache.clearRequests()


meshroom/ui/qml/ImageGallery/ImageGallery.qml
```diff
@@ -34,12 +34,16 @@ Panel {
     title: "Image Gallery"
     implicitWidth: (root.defaultCellSize + 2) * 2
 
+    onReadOnlyChanged: {
+        m.readOnly = (_reconstruction && m.currentCameraInit) ? _reconstruction.isGroupComputing(m.currentCameraInit) || displayHDR.checked : root.readOnly || displayHDR.checked
```
fabiencastan, line 38:
```suggestion
        m.readOnly = (_reconstruction && m.currentCameraInit) ? (_reconstruction.isGroupComputing(m.currentCameraInit) || displayHDR.checked) : )root.readOnly || displayHDR.checked)
```

meshroom/ui/reconstruction.py
```diff
@@ -1116,6 +1116,17 @@ def getPoseRT(self, viewpoint):
 
         return R, T
 
+    @Slot(QObject, result=bool)
+    def isGroupComputing(self, cameraInit):
```
mugulmd, line 1120: Maybe I'm missing the point but I don't really see why we need a new slot, can't we just use the `currentCameraInit.locked` property ?

cbentejac: You are completely right! This has been removed and we do directly use `currentCameraInit.locked` instead.

meshroom/ui/qml/ImageGallery/ImageGallery.qml
cbentejac: Done.

meshroom/ui/qml/ImageGallery/ImageGallery.qml
```diff
@@ -34,12 +34,16 @@ Panel {
     title: "Image Gallery"
     implicitWidth: (root.defaultCellSize + 2) * 2
 
+    onReadOnlyChanged: {
```
mugulmd, line 37: I don't think this is necessary since the `m.readOnly` property already depends on `root.readOnly` (the QML engine should update it dynamically).
Same for `onCurrentCameraInitChanged` below.

meshroom/ui/qml/ImageGallery/ImageGallery.qml
```diff
@@ -34,12 +34,16 @@ Panel {
...
+        m.readOnly = (_reconstruction && m.currentCameraInit) ? (m.currentCameraInit.locked || displayHDR.checked) : (root.readOnly || displayHDR.checked)
+    }
+
     QtObject {
         id: m
         property variant currentCameraInit: _reconstruction && _reconstruction.tempCameraInit ? _reconstruction.tempCameraInit : root.cameraInit
         property variant viewpoints: currentCameraInit ? currentCameraInit.attribute('viewpoints').value : undefined
         property variant intrinsics: currentCameraInit ? currentCameraInit.attribute('intrinsics').value : undefined
-        property bool readOnly: root.readOnly || displayHDR.checked
+        property bool readOnly: (_reconstruction && currentCameraInit) ? (currentCameraInit.locked || displayHDR.checked) : (root.readOnly || displayHDR.checked)
```
mugulmd, line 46: You can factor out the `displayHDR.checked` to avoid writing it twice

cbentejac: Done.

meshroom/ui/qml/ImageGallery/ImageGallery.qml
cbentejac: This wasn't fully reliable before accessing `currentCameraInit.locked` (I'm not entirely sure why) but it is now. Thanks for your help!

mugulmd (approved):

PR #1944, opened 2023-03-21
[panorama] Panorama preview size

servantftechnicolor:
New parameter to panoramaPostProcessing : 

Panorama preview width (Which has to be smaller than the panorama original width).


cbentejac (approved):
I rebased the PR on top of develop (it was originally on top of the `dev/contactSheet` branch) and fixed the range of the `previewSize` parameter, as the slider only accepted 1000 or 0 as values.

PR #1951, opened 2023-04-02
Update build-ubuntu.sh

rakhnin:
fix docker build ubuntu script


cbentejac (approved):
Thank you for the fix!

PR #1954, opened 2023-04-04
[sfmTransform] add auto mode

servantftechnicolor:
New sfmTransform Node

- Use cameras X axis to estimate the horizontal plane (and the horizon)
- Use GPS measurements (if available) to align the Z axis with the north
- Use GPS measurements (if available) to set the scale (otherwise scaled such that stddev=1)
- Use cameras center mean as frame origin


meshroom/nodes/aliceVision/SfMTransform.py
```diff
@@ -45,7 +46,7 @@ class SfMTransform(desc.AVCommandLineNode):
                         " * from_markers: Align specific markers to custom coordinates\n"
                         " * from_gps: Align with the gps positions from the image metadata",
             value='auto_from_landmarks',
```
fabiencastan, line 48:
```suggestion
            value='auto',
```

fabiencastan (approved):

PR #1966, opened 2023-04-07
[core] Correctly parse status in version names when it exists

cbentejac:
## Description

`Version` objects were meant to be used only with numerical characters, separated by periods. Now that the version of Meshroom can also contain its status ("develop" when it is in develop mode), the `Version` class needs to be able to parse the status string (when it exists) as well as the version components.

This PR addresses this by handling the status as a separate string, which is returned in addition to the tuple of integers when the whole version string is parsed in the first place.

This prevents issues when opening files that have been saved with a version like "x.x.x-develop".

### DIFF
--- a/meshroom/core/__init__.py
+++ b/meshroom/core/__init__.py
@@ -147,16 +147,19 @@ class Version(object):
         """
         if len(args) == 0:
             self.components = tuple()
+            self.status = str()
         elif len(args) == 1:
             versionName = args[0]
             if isinstance(versionName, str):
-                self.components = Version.toComponents(versionName)
+                self.components, self.status = Version.toComponents(versionName)
             elif isinstance(versionName, (list, tuple)):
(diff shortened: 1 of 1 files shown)


meshroom/core/__init__.py
```diff
@@ -206,17 +209,24 @@ def __le__(self, other):
...
+            tuple of int, string: split version numbers, status if any (or empty string)
         """
         if not versionName:
-            return ()
-        return tuple([int(v) for v in versionName.split(".")])
+            return (), str()
+
+        status = str()
+        # If there is a status, it is placed after a "-"
+        splitComponents = versionName.split("-")
```
fabiencastan, line 226:
```suggestion
        splitComponents = versionName.split("-", maxsplit=1)
```

cbentejac: Done!

fabiencastan (approved):

PR #1974, opened 2023-04-12
Node to split reconstructed and not reconstructed cameras

servantftechnicolor:
Take an input sfmdata

- Create 1 sfmdata with reconstructed views only (and landmarks)
- Create 1 sfmdata with non reconstructed views only (without landmarks)

TCSR-305
see https://github.com/alicevision/AliceVision/pull/1409

### DIFF
--- a/meshroom/nodes/aliceVision/SfMSplitReconstructed.py
+++ b/meshroom/nodes/aliceVision/SfMSplitReconstructed.py
@@ -1,8 +1,6 @@
-__version__ = "2.0"
+__version__ = "1.0"
 
 from meshroom.core import desc
-
-import os.path
 
 
 class SfMSplitReconstructed(desc.AVCommandLineNode):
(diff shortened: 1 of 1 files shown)


meshroom/nodes/aliceVision/SfMSplitReconstructed.py
```diff
@@ -0,0 +1,50 @@
+__version__ = "2.0"
```
mugulmd, line 1: 1.0

meshroom/nodes/aliceVision/SfMSplitReconstructed.py
```diff
@@ -0,0 +1,50 @@
...
+
+import os.path
+
+
+class SfMSplitReconstructed(desc.AVCommandLineNode):
+    commandLine = 'aliceVision_sfmSplitReconstructed {allParams}'
+    size = desc.DynamicNodeSize('input')
+
+    category = 'Utils'
+    documentation = '''
```
mugulmd, line 13: a little documentation never hurt anybody 😇

meshroom/nodes/aliceVision/SfMSplitReconstructed.py
```diff
@@ -0,0 +1,50 @@
+__version__ = "2.0"
+
+from meshroom.core import desc
+
+import os.path
```
mugulmd, line 5: I don't think we need that import

mugulmd (approved):

PR #1975, opened 2023-04-12
[ui] add new patterns to load images in viewer2D

Garoli:
## Description
This PR adds the support of new patterns to load images in the 2D image viewer. It will also be easier to add ones. 

## Features list

- [x] Add **< FILENAME >** and **< INSTRINSIC_ID >** pattern
- [x] patterns are given through a dictionary


meshroom/ui/components/filepath.py
```diff
@@ -49,6 +49,12 @@ def extension(self, path):
         """ Returns the extension (.ext) of a pathname """
         return os.path.splitext(self.asStr(path))[-1]
 
+    @Slot(str, result=str)
+    @Slot(QUrl, result=str)
+    def removeExtension(self, path):
+        """ Returns the pathname without its extension (.ext)"""
+        return os.path.splitext(path)[0]
```
mugulmd, line 56: You should use `self.asStr(path)` as in the methods above to properly handle both strings and QUrls

meshroom/ui/qml/Viewer/Viewer2D.qml
```diff
@@ -219,34 +219,47 @@ FocusScope {
...
         }
         if (_reconstruction && (!displayedNode || outputAttribute.name == "gallery")) {
-            return getViewpointPath(_reconstruction.selectedViewId);
+            return Filepath.stringToUrl(getViewpointAttribute("path",_reconstruction.selectedViewId));
         }
-        return getFileAttributePath(displayedNode, outputAttribute.name, _reconstruction ? _reconstruction.selectedViewId : -1);
+
+        var viewId = -1;
+        var intrinsicId = -1;
+        var fileName = -1;
```
mugulmd, line 227:
```suggestion
        var fileName = "";
```

meshroom/ui/qml/Viewer/Viewer2D.qml
```diff
@@ -219,34 +219,47 @@ FocusScope {
...
         if (!node)
             return "";
         for (var i = 0; i < node.attributes.count; i++) {
             var attr = node.attributes.at(i);
             if (attr.name == attrName) {
-                let pattern = String(attr.value).replace("<VIEW_ID>", viewId);
-                let path = Filepath.globFirst(pattern);
+                let path = String(attr.value)
+                for (var key in patterns) {
+                    if (patterns.hasOwnProperty(key)) {
```
mugulmd, line 247: I think I'm missing the point, why do we need `hasOwnProperty` here ? 😅

meshroom/ui/qml/Viewer/Viewer2D.qml
```diff
@@ -219,34 +219,47 @@ FocusScope {
...
                 return Filepath.stringToUrl(path);
             }
         }
         return "";
     }
 
-    function getViewpointPath(viewId) {
-        // get viewpoint from cameraInit with matching id
-        // and get its image filepath
+    function getViewpointAttribute(attributeName, viewId) {
```
mugulmd, line 257: Since this function can now return various attributes with potentially different types, its default return value should not be an empty string but `undefined` or `null`

meshroom/ui/qml/Viewer/Viewer2D.qml
Garoli: Just took the snipet from [stackoverflow](https://stackoverflow.com/questions/684672/how-do-i-loop-through-or-enumerate-a-javascript-object) saying that one should verify the key is a property of the object 
I can remove this if you want to !

mugulmd: Thanks for the reference, that's fine by me let's keep it !

mugulmd (approved):

mugulmd:
Your 2nd commit is labeled `TO SQUASH`, can you squash it so the git history is clean ? Thanks 😄

meshroom/ui/qml/Viewer/Viewer2D.qml
fabiencastan: javascript hell

PR #2009, opened 2023-05-15
[doc] Update readme for custom pipelines and nodes

fabiencastan:


README.md
```diff
@@ -51,6 +51,19 @@ Overall presentation of the protogrammetry practice with Meshroom and detailed p
 Presentation of the Meshroom software with a focus on using it for Match Moving.
 
 
+## Customization
+
+### Custom Pipelines
+
+You can create custom pipelines in the user interface and save it as template: "File > Advanced > Save As Template".
+You can define the `MESHROOM_PIPELINE_TEMPLATES_PATH` environment variable to specific folders to makes these pipelines available in Meshroom.
```
simogasp, line 59:
```suggestion
You can define the `MESHROOM_PIPELINE_TEMPLATES_PATH` environment variable to specific folders to make these pipelines available in Meshroom.
```

simogasp (changes requested):

README.md
```diff
@@ -51,6 +51,19 @@ Overall presentation of the protogrammetry practice with Meshroom and detailed p
...
+
+### Custom Nodes
+
+You can also create custom nodes in python and make them available in Meshroom using the `MESHROOM_NODES_PATH` environment variable.
```
natowi, line 63: Maybe we could have a custom_nodes folder in the future that is used by default and you only need to copy the code in this folder, without specifying a variable.

fabiencastan: People can add custom python code in the subfolder meshroom/pipelines, but yes a folder in the root may be easier.

cbentejac (approved):

PR #2030, opened 2023-05-29
[nodes] rewrite RenderAnimatedCamera

mugulmd:
## Description

The goal of this PR is to rewrite the RenderAnimatedCamera node and the Blender-Python script it runs behind the scenes to make it work properly and support wireframe rendering.

This node now takes as input:
- `cameras`: SfMData in JSON format containing the views, poses and intrinsics of the scene to render
- `model`: scene model, can be either a mesh in OBJ format or a point cloud in Alembic format
- `pointCloudParams`: rendering parameters if the input is a point cloud
- `meshParams`: rendering parameters if the input is a mesh, the most important one being the `shading` method (wireframe or line art)
- (optional) `undistortedImages`: undistorted images to use as background of the rendered views for reference.

The output of this node is one frame per view corresponding to the input geometry rendered in Blender using the input camera parameters.
This output can be directly visualized in Meshroom's 2D viewer by double-clicking the node.

Also the node has been renamed `ScenePreview` for clarity and now belongs to the `Utils` category.

## Notes

Since the `cameras` input must be in JSON format, users may need to use a `ConvertSfMFormat` node


meshroom/nodes/blender/RenderAnimatedCamera.py
```diff
@@ -1,47 +1,48 @@
...
         ),
         desc.File(
-            name='scriptPath',
+            name='script',
             label='Script Path',
-            description='''Path to the internal script for rendering in Blender''',
+            description='Path to the internal script for rendering in Blender',
             value=os.path.join(currentDir, 'scripts' ,'renderAnimatedCameraInBlender.py'),
-            uid=[],
+            uid=[0],
```
fabiencastan, line 31: We use an absolute path here. So I'm not sure if it should invalidate, as we may move the Meshroom install...

cbentejac (approved):
Tested with pretty much all the possible combinations and it works as expected!

A couple notes for a (potential) ulterior PR:
- It'd probably be best to group the `Display Background` and `Undistorted Images` together. `Undistorted Images` is optional but `Display Background`, which enables it, is always by default, and the relationship between the two is not obvious. Grouping them would prevent having errors because `Undistorted Images` is willingly not set while `Display Background` is still on.
- Using a logger object instead of `print()` would be nice, as it would allow us to have timestamps in the logs.

PR #2031, opened 2023-05-30
[ui] ImageGallery: Reset viewpoints and intrinsics when removing all the images

cbentejac:
## Description

This PR fixes an issue that occurred when all the images from the Gallery were manually removed one by one (either with the "Del" key or the "Remove" option): the list of viewpoints was emptied as expected (but never reset) but the list of intrinsics was never modified (neither emptied nor reset). This could cause errors later on, when re-filling the Gallery with new images that were unrelated to the remaining intrinsics.

When the list of viewpoints is now emptied, both the list of viewpoints and the list of intrinsics are reset, in the same fashion as when "Clear Images" is called.

## Features list

- [x] Reset the lists of viewpoints and intrinsics when the last image of the active group is removed manually from the Gallery.

### DIFF
--- a/meshroom/ui/qml/ImageGallery/ImageGallery.qml
+++ b/meshroom/ui/qml/ImageGallery/ImageGallery.qml
@@ -29,6 +29,7 @@ Panel {
     property bool readOnly: false
 
     signal removeImageRequest(var attribute)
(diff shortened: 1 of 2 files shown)


demoulinv (approved):
PR successfully tested

PR #2032, opened 2023-06-02
Add a template for the HDR fusion

cbentejac:
## Description

This PR adds a new template to perform a stand-alone HDR fusion, which used to be only available in the HDR panorama templates.

As the `LdrToHdrMerge` node's only output was an SfMData file, its output folder is now exposed so that results of the fusion are available. 

The `Publish` node's support for inputs has also been extended to folders: if a folder is provided in the list of inputs, its content will be recursively published to the output folder.  

## Features list

- [x] Add support of folders for the inputs of the `Publish` node;
- [x] Expose the output folder of the `LdrToHdrMerge` node;
- [x] Add a new template dedicated to the HDR fusion.

### DIFF
--- a/meshroom/pipelines/hdrFusion.mg
+++ b/meshroom/pipelines/hdrFusion.mg
@@ -1,0 +1,88 @@
+{
+    "header": {
+        "nodesVersions": {
+            "Publish": "1.3",
+            "LdrToHdrSampling": "4.0",
+            "LdrToHdrMerge": "4.1",
+            "LdrToHdrCalibration": "3.0",
+            "CameraInit": "9.0"
+        },
+        "releaseVersion": "2023.2.0-develop",
+        "fileVersion": "1.1",
+        "template": true
+    },
+    "graph": {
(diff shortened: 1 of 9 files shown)


meshroom/pipelines/hdrFusion.mg
```diff
@@ -0,0 +1,99 @@
...
+                    "radial3",
+                    "brown",
+                    "fisheye1",
+                    "3deanamorphic4",
+                    "3deradial4",
+                    "3declassicld"
+                ]
+            }
+        },
+        "PanoramaPrepareImages_1": {
```
fabiencastan, line 76: No need for panoramaPrepareImages in this case

fabiencastan (changes requested):

meshroom/pipelines/hdrFusion.mg
cbentejac: Done.

mugulmd (approved):

PR #2052, opened 2023-06-13
[ui] Display an icon on nodes that have viewable 3D outputs

cbentejac:
## Description

This PR follows #2047 and extends the display of the icon for viewable outputs to outputs that are viewable in the 3D Viewer. Any `File` output that has an extension supported by the 3D Viewer can be loaded.

The distinction between outputs that can be loaded in the 2D Viewer and those that can be loaded in the 3D Viewer is made, and the tooltip that appears when hovering over the icon reflects that. The tooltip message has also been simplified. 

The case where a node has both 2D and 3D outputs is handled with a corresponding tooltip message, although there is currently no such node.

### DIFF
--- a/meshroom/ui/qml/GraphEditor/Node.qml
+++ b/meshroom/ui/qml/GraphEditor/Node.qml
@@ -308,7 +308,8 @@ Item {
 
                             MaterialLabel {
                                 id: nodeImageOutput
-                                visible: node.hasImageOutput && ["SUCCESS"].includes(node.globalStatus) && node.chunks.count > 0
+                                visible: (node.hasImageOutput || node.has3DOutput)
+                                         && ["SUCCESS"].includes(node.globalStatus) && node.chunks.count > 0
                                 text: MaterialIcons.visibility
                                 padding: 2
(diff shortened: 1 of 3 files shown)


meshroom/ui/qml/GraphEditor/Node.qml
```diff
@@ -317,8 +318,14 @@ Item {
                                     id: nodeImageOutputTooltip
                                     parent: header
                                     visible: nodeImageOutputMA.containsMouse && nodeImageOutput.visible
-                                    text: "This node has at least one output that can be loaded in the 2D Viewer.\n" +
-                                          "Double-clicking on this node will load it in the 2D Viewer."
+                                    text: {
+                                        if (node.hasImageOutput)
```
fabiencastan, line 322:
```suggestion
                                        if (node.hasImageOutput && !node.has3DOutput)
```

cbentejac: Done

fabiencastan (approved):

PR #2054, opened 2023-06-14
Fix submit through simpleFarm

mugulmd:
## Description

Bug fix for simpleFarm submission:
use node name instead of node type when naming tasks as they are guaranteed to be unique

### DIFF
--- a/meshroom/submitters/simpleFarmSubmitter.py
+++ b/meshroom/submitters/simpleFarmSubmitter.py
@@ -76,7 +76,7 @@ class SimpleFarmSubmitter(BaseSubmitter):
         allRequirements.extend(self.config['GPU'].get(node.nodeDesc.gpu.name, []))
 
         task = simpleFarm.Task(
-            name=node.nodeType,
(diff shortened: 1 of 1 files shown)


meshroom/core/graph.py
```diff
@@ -1094,7 +1094,7 @@ def dfsMaxEdgeLength(self, startNodes=None):
         """
         nodesStack = []
         edgesScore = defaultdict(lambda: 0)
-        visitor = Visitor(reverse=False, dependenciesOnly=False)
+        visitor = Visitor(reverse=False, dependenciesOnly=True)
```
fabiencastan, line 1097: `dfsMaxEdgeLength` is not something used for nodes alignment in the UI? In which case, dependenciesOnly should be False.

fabiencastan: Maybe not the case anymore.
Propably better to just expose dependenciesOnly as a parameter of `dfsMaxEdgeLength`.

fabiencastan (approved):

PR #2057, opened 2023-06-14
[ui] thumbnails: fallback if thumbnailDir could not be created

mugulmd:
## Description

Avoid crashes and unwanted behaviours when thumbnailDir cannot be created (for example if user does not have write permissions in target directory).

### DIFF
--- a/meshroom/ui/components/thumbnail.py
+++ b/meshroom/ui/components/thumbnail.py
@@ -89,7 +89,11 @@ class ThumbnailCache(QObject):
         ThumbnailCache.cleaningThread.start()
 
         # Make sure the thumbnail directory exists before writing into it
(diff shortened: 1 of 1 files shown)


fabiencastan (approved):

meshroom/ui/components/thumbnail.py
```diff
@@ -89,7 +89,10 @@ def initialize():
         ThumbnailCache.cleaningThread.start()
 
         # Make sure the thumbnail directory exists before writing into it
-        os.makedirs(ThumbnailCache.thumbnailDir, exist_ok=True)
+        try:
+            os.makedirs(ThumbnailCache.thumbnailDir, exist_ok=True)
+        except OSError:
+            pass
```
fabiencastan, line 95: Maybe add a warning with the path of the cache folder.

fabiencastan (approved):

PR #2072, opened 2023-06-19
Advanced pixelwise merging

demoulinv:
## Description

Add option and parameters in LdrToHdrMerge node for advanced pixelwise HDR merging.


mugulmd (approved):

mugulmd:
### Note

When merging RAW images with the `HDR Fusion` pipeline, better results are achieved with the `Calibration Method` and `Working Color Space` set to `linear` in the `LdrToHdrSampling` node.
It might be a good idea to use these as the default parameters in the future.

PR #2076, opened 2023-06-20
semantic image segmentation

servantftechnicolor:
Create a node which calls AliceVision semantic segmentation


mugulmd:
## Note: Usage with the FeatureExtraction node

To use the masks generated by the `ImageSegmentation` node in the `FeatureExtraction` node, users must:
- connect the `Masks Folder` output and input
- set the `Mask Extension` to `exr`
- turn on `Invert Mask`

PR #2087, opened 2023-06-28
[BoundingBox] Add visualization for automatically computed bounding box

almarouk:
Linked to https://github.com/alicevision/AliceVision/pull/1473


meshroom/core/nodes/__init__.py
```diff
@@ -0,0 +1,23 @@
...
+    type[Node]
+        the corresponding type class
+    """
+    if nodeType=="Meshing":
+        from meshroom.core.nodes.MeshingNode import MeshingNode
+        cls = MeshingNode
+    else:
+        from meshroom.core.node import Node
+        cls = Node
+    return cls
```
fabiencastan, line 23: The NodeDesc should be able to create a node, with a default implementation that returns "Node".
And then a NodeDesc could override the default Node without having hardcoded node names in the engine.

fabiencastan (changes requested):
We cannot have hardcoded information from nodes in the code engine.

meshroom/core/nodes/__init__.py
almarouk: Proposed solution (see commit https://github.com/alicevision/Meshroom/pull/2087/commits/ce0aa3ef7a9380eccff8a94f3a9ab225cfa1a719):
- move bounding box update handling to the Meshing node description
- give Node object reference to the corresponding node description
- add a third attribute dictionary in Node (in addition to internal and IO attributes dictionaries) that gives a flexibility to add runtime-related attributes that can be used in qml code

PR #2088, opened 2023-06-28
[codeFactor] add SEMANTIC enum for attribute description

almarouk:
Semantic enum added for easier tracking of available and used semantics


cbentejac:
This PR should not be merged until #1882 is merged.
With our current version of PySide2, enums cannot be exposed to QML (https://bugreports.qt.io/browse/PYSIDE-957). Rather than sending the enum values as integers, we shall wait until our version of PySide is updated so that enums can be exposed to QML.

cbentejac:
This is not as straightforward as I thought it would be, even with PySide2.5.1.5. I'm changing this PR's status to "Draft", we'll need to rework it.

PR #2113, opened 2023-07-13
tolerant bracket size selection

servantftechnicolor:
Bracket size selection is now tolerant to at most two spurious image in the images list.

rule : (Bracket size * countGroupsWithThisSize) >= totalImages - 2

### DIFF
--- a/meshroom/nodes/aliceVision/LdrToHdrCalibration.py
+++ b/meshroom/nodes/aliceVision/LdrToHdrCalibration.py
@@ -1,6 +1,7 @@
 __version__ = "3.0"
 
 import json
+import os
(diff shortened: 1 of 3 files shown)


cbentejac (approved):
I added a detection for new exposure groups based on the path of the images to correctly handle cases where the outlier(s) might have an exposure level that would be coherent with the current exposure group, thus causing an erroneous bracket detection.

For example, if an outlier's exposure level is (22.0, 10.6, 200) and it gets sorted before a dataset whose first image has an exposure level of (22.0, 10.2, 200), both images would be put into the same exposure group unless we use the path information that's available.

PR #2114, opened 2023-07-13
[pipelines] Add colors for CameraTracking and Photog+CamTrack templates

cbentejac:
## Description

This PR adds default colors for the nodes from the "Camera Tracking" and "Photogrammetry and Camera Tracking" templates so that the different processing blocks of these rather complex pipelines can be identified more easily.

"Camera Tracking" template:
CamTrackPipeline

"Photogrammetry and Camera Tracking" template:
PhotogCamTrackPipeline

### DIFF
--- a/meshroom/pipelines/cameraTracking.mg
+++ b/meshroom/pipelines/cameraTracking.mg
@@ -1,34 +1,34 @@
 {
     "header": {
         "pipelineVersion": "2.2",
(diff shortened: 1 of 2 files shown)


natowi:
It would be great if you could also add pipeline specific documentation to the nodes

mugulmd (approved):

PR #2145, opened 2023-08-07
[ui] Preserve last `CameraInit` index when updating the CameraInits list

cbentejac:
## Description

Prior to this PR, when a `CameraInit` node was added to or removed from the list of `CameraInit` nodes, the active `CameraInit` was always reset to the first one in the list, independently from its previous value. This meant that whenever a `CameraInit` node was added to or removed from the graph while the Image Gallery was displaying a `CameraInit` node that was not the first one in the list, the Image Gallery was automatically updated to display the first `CameraInit` node without any intervention from the user, as shown below:
<div align="center"></div>

This pull request changes that behaviour by only modifying the active `CameraInit` node if no `CameraInit` node has been assigned yet, or if the active `CameraInit` node does not exist anymore (meaning it has been removed).

This requires to emit the `cameraInitChanged()` signal every single time the list of existing `CameraInit` nodes is modified even if the active one remains the same, and to connect it to the ImageGallery in order to ensure the index from the combo box always corresponds to the currently active `CameraInit` node. Indeed, when the list of CameraInits is updated, the
model for the combo box is reset, and it needs to be re-updated with the correct non-default value.
<div align="center"></div>

### DIFF
--- a/meshroom/ui/reconstruction.py
+++ b/meshroom/ui/reconstruction.py
@@ -573,7 +573,12 @@ class Reconstruction(UIGraph):
         if set(self._cameraInits.objectList()) == set(cameraInits):
             return
         self._cameraInits.setObjectList(cameraInits)
-        self.cameraInit = cameraInits[0] if cameraInits else None
+
+        if self.cameraInit is None or self.cameraInit not in cameraInits:
+            self.cameraInit = cameraInits[0] if cameraInits else None
+
+        # Manually emit the signal to ensure the active CameraInit index is always up-to-date in the UI
+        self.cameraInitChanged.emit()
 
     def getCameraInitIndex(self):
         if not self._cameraInit:
--- a/meshroom/ui/qml/ImageGallery/ImageGallery.qml
+++ b/meshroom/ui/qml/ImageGallery/ImageGallery.qml
@@ -35,6 +35,14 @@ Panel {
     title: "Image Gallery"
     implicitWidth: (root.defaultCellSize + 2) * 2
 
+    Connections {
+        target: _reconstruction
+
+        function onCameraInitChanged() {
+            nodesCB.currentIndex = root.cameraInitIndex
+        }
+    }
+
     QtObject {
         id: m
         property variant currentCameraInit: _reconstruction && _reconstruction.tempCameraInit ? _reconstruction.tempCameraInit : root.cameraInit


meshroom/ui/reconstruction.py
```diff
@@ -572,8 +572,20 @@ def updateCameraInits(self):
...
+        if self.cameraInit in cameraInits and self.getCameraInitIndex() != -1:
+            idx = self.getCameraInitIndex()
+
+        # Reset the index with the first CameraInit in the list
         self.cameraInit = cameraInits[0] if cameraInits else None
 
+        # If the previously active CameraInit node still exists, set it back as the active one
+        if idx != -1 and self.cameraInit is not None:
+            self.setCameraInitIndex(idx)
+
```
fabiencastan, line 588: Maybe simpler to update the CameraInit only when we need it (first assignment or when the old cameraInit is no more part of the scene, like when we load another project):
```suggestion
        if self.cameraInit is None or self.cameraInit not in cameraInits:
            self.cameraInit = cameraInits[0] if cameraInits else None
```

fabiencastan (changes requested):

meshroom/ui/reconstruction.py
cbentejac: I tested that solution and it does not work.

When the list of `CameraInits` is updated, the model from the combobox is emptied and refilled, and if a "cameraInitIndexChanged" signal is not emitted by actively setting `self.cameraInit`, no further `setCameraInitIndex` is effective as the corresponding model does not exist anymore. We need to "transition" by the first `CameraInit` in the list to keep the correct index when the list is updated.

fabiencastan: We could emit 'cameraInitChanged' manually, isn't it?
In all case, we could avoid using the index here and reset self.cameraInit directly.

cbentejac: Done:
- the active `CameraInit` is only modified if it hasn't been assigned yet or if it has been removed from the list
- the `cameraInitChanged()` signal is emitted every single time the list of `CameraInits` is modified and it is connected to the QML to ensure that the combo box index, which is reset every single time the list is modified no matter what, is re-updated with the correct value

fabiencastan (commented):
Do we need to check if we have removed all cameraInit before accessing [0]?

cbentejac:
> Do we need to check if we have removed all cameraInit before accessing [0]?

If all the CameraInits have been removed, then the test in `self.cameraInit = cameraInits[0] if cameraInits else None` will fail, so there's no risk of accessing an index that does not exist.

fabiencastan (approved):

PR #2154, opened 2023-08-16
[nodes] HDR Fusion: Use the same bracket detection as in AliceVision

cbentejac:
## Description

The bracket detection performed in Meshroom used to differ from the one performed in AliceVision.

The metadata that were retrieved to perform the exposure comparisons were not the same, and where AliceVision was actually computing an exposure value, Meshroom was just performing a comparison between the shutter speed, fnumber and ISO values, which resulted in less accurate groups.

This PR updates the HDR Fusion nodes so that they use the same bracket detection as AliceVision.

The `getExposure()` static method that is added to the `LdrToHdrSampling`, `LdrToHdrCalibration` and `LdrToHdrMerge` node is the pythonic version of the `getExposure()` method from the View class in AliceVision.

### DIFF
--- a/meshroom/nodes/aliceVision/LdrToHdrCalibration.py
+++ b/meshroom/nodes/aliceVision/LdrToHdrCalibration.py
@@ -1,6 +1,7 @@
 __version__ = "3.0"
 
 import json
+import math
 import os
 from collections import Counter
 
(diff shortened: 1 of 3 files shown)


servantftechnicolor (approved):
Tested

Issue #494, opened 2019-06-03
Manually associating points between image pairs


ion1:
I would like to reconstruct the relative camera poses from certain photographs but I have trouble making progress with automatic feature extraction and matching.

[Hugin](http://hugin.sourceforge.net/) has a convenient UI for manually marking common points on pairs of images (with automatic subpixel precision fine-tuning on roughly marked points) but it only does panorama stitching. Is it possible to have something like this with Meshroom? Thank you!

![Hugin Control Point UI](https://user-images.githubusercontent.com/39660/58797631-2f6cd500-8609-11e9-8893-bdb1957d9167.png)

skinkie:
@ion1 https://github.com/alicevision/meshroom/issues/450#issuecomment-485497671 It is already on the wishlist I read.

With respect to Hugin. I hardly would consider that interface convenient. Have you ever tried the 'line based' notation? And wouldn't you expect a tool to rotate the image for you to compare it? ;-)

In my perspective a fast way to orient the images by a human (for example based on points at the horizon) could significantly reduce the search space. If this is done in concert with the FeatureExtraction that might be a very nice method.

ion1:
There seems to be a commercial product with this feature, [this video demonstrates the UI (7:22–8:22)](https://youtu.be/3Lm8fiDF1us?t=442). This would be very nice to have in an open source tool.

skinkie:
@ion1 the iWitness thing is like Hugin, it is rather uninformed. I would expect from a tool that a single click would already zoom in to the portion of the screen that is relevant. PhotoModeler looks to be some sort of assisted 3d modellerer which is pretty cool conceptually speaking but also drives only on human input. Defining a user input flow, or have a human state that the recovered direction is accurate might reduce the search space, and therefore the candidates that are proposed. I'll try to come up with such input and its benefits.

french-paragon:
Agisoft photoscan (now metashape) has a landmark feature which is really usefull to manually specify points in images to improve alignement.

natowi:
Here is a short draft description on the Feature Extraction file format as starting point for implementation:
https://gist.github.com/natowi/ad9ce6d9f912e089bba9b69ec91d7115

skinkie:
I think this might be another good starting point for a GUI approach. https://fspy.io/

natowi:
@skinkie How so?

skinkie:
[...]

PR #2177, opened 2023-09-04
Add visualisation mode for landmarks observations

almarouk:
- add an option to visualize the landmarks observations
- observed landmarks for current camera + links to cameras observing the same landmarks
- linked to the 2D viewer to zoom on specific landmarks

landmarks


natowi:
It would be great if an [edit mode](https://github.com/alicevision/meshroom/issues/494) could be added on top of this feature in a following contribution :)

simogasp:
That's very interesting!
Another interesting feature would be colour-coding the 3D points based on their visibility (i.e., the number of observations). That gives a **hint** of how "reliable" the 3D point is.

PR #2200, opened 2023-09-21
NodalSfm new node

servantftechnicolor:
Create new node for nodal sfm (pure rotation tracking)

see https://github.com/alicevision/AliceVision/pull/1548

### DIFF
--- a/meshroom/nodes/aliceVision/RelativePoseEstimating.py
+++ b/meshroom/nodes/aliceVision/RelativePoseEstimating.py
@@ -50,6 +50,13 @@ class RelativePoseEstimating(desc.AVCommandLineNode):
             uid=[0],
             joinChar=",",
         ),
+        desc.BoolParam(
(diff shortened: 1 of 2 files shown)


meshroom/nodes/aliceVision/RelativePoseEstimating.py
```diff
@@ -50,6 +50,13 @@ class RelativePoseEstimating(desc.AVCommandLineNode):
             uid=[0],
             joinChar=",",
         ),
+        desc.BoolParam(
```
simogasp, line 53: maybe a bit late but why not "RelativePoseEstimation.py"? It's coherent with all the other names

servantftechnicolor: I've been asked to name the nodes with "*ing"

mugulmd (approved):

PR #2243, opened 2023-11-07
[Photometric Stereo] MultiView fusion in Texturing

jmelou:
-  Texturing with output of Photometric Stereo: albedo and normal maps
-  Lighting calibration now supports spherical harmonic lighting

Based on https://github.com/alicevision/AliceVision/pull/1590


fabiencastan (commented):
Need to add "SH" in LightingCalibration methods, isn't it?

meshroom/nodes/aliceVision/PhotometricStereo.py
```diff
@@ -62,6 +62,15 @@ class PhotometricStereo(desc.CommandLineNode):
             advanced=True,
             uid=[0]
         ),
+        #desc.FloatParam(
```
fabiencastan, line 65: ?

fabiencastan (approved):

PR #2252, opened 2023-11-16
added support for unknown image detector/descriptor type

mh0g:
## Description

Added support for an 'unknown' feature type for importing features that are not computed by alicevision.
https://github.com/alicevision/AliceVision/pull/1598


mh0g:
pr obsolete, unknown type already in develop https://github.com/alicevision/Meshroom/blob/develop/meshroom/core/utils.py#L7

PR #2258, opened 2023-11-23
[core] Only update the running chunk to `STOPPED` when stopping computations

cbentejac:
## Description

When stopping the local computations of the graph with the "Stop" button, the status of the chunks that have already been updated to `SUCCESS` is preserved, and the submitted chunks are reset. However, when stopping the computation of a specific node with the "Stop Computation" option, all the chunks are updated to `STOPPED` independently from their previous status.

This PR ensures that chunks of a node that are already in the `SUCCESS` state will not be updated when stopping the computations, and those that are `SUBMITTED` but not `RUNNING` will be reset to `NONE`; only the `RUNNING` chunk will be updated to `STOPPED`.

### DIFF
--- a/meshroom/core/node.py
+++ b/meshroom/core/node.py
@@ -429,7 +429,11 @@ class NodeChunk(BaseObject):
         self.upgradeStatusTo(Status.SUCCESS)
 
     def stopProcess(self):
-        self.upgradeStatusTo(Status.STOPPED)
+        if not self.isExtern():
+            if self._status.status == Status.RUNNING:
+                self.upgradeStatusTo(Status.STOPPED)
+            elif self._status.status == Status.SUBMITTED:
+                self.upgradeStatusTo(Status.NONE)
         self.node.nodeDesc.stopProcess(self)
 
     def isExtern(self):
@@ -942,8 +946,7 @@ class BaseNode(BaseObject):
     def stopComputation(self):
         """ Stop the computation of this node. """
         for chunk in self._chunks.values():
-            if not chunk.isExtern():
-                chunk.stopProcess()
+            chunk.stopProcess()
 
     def getGlobalStatus(self):
         """


meshroom/core/node.py
```diff
@@ -940,10 +940,15 @@ def endSequence(self):
...
-            if not chunk.isExtern():
+        """ Stop the computation of this node. Chunks that are running locally will be stopped, and those that
+        are submitted locally will be reset.
+        """
+        for chunk in [c for c in self._chunks.values() if c.status.status == Status.RUNNING or
+                                                          c.status.status == Status.SUBMITTED]:
+            if not chunk.isExtern() and chunk.status.status == Status.RUNNING:
                 chunk.stopProcess()
+            elif not chunk.isExtern():
+                chunk.upgradeStatusTo(Status.NONE)
```
fabiencastan, line 951:
```suggestion
            if not chunk.isExtern():
                if chunk.status.status == Status.RUNNING:
                    chunk.stopProcess()
                else:
                    chunk.upgradeStatusTo(Status.NONE)
```

fabiencastan:
Or maybe more readable to do it directly in the NodeChunk:
https://github.com/alicevision/Meshroom/blob/bb9195f99561f4c8f63040638518a24d1ba9bf26/meshroom/core/node.py#L431
Only update the status to STOPPED if it was RUNNING.

meshroom/core/node.py
```diff
@@ -940,10 +940,15 @@ def endSequence(self):
...
 
     def stopComputation(self):
-        """ Stop the computation of this node. """
-        for chunk in self._chunks.values():
-            if not chunk.isExtern():
+        """ Stop the computation of this node. Chunks that are running locally will be stopped, and those that
+        are submitted locally will be reset.
+        """
+        for chunk in [c for c in self._chunks.values() if c.status.status == Status.RUNNING or
+                                                          c.status.status == Status.SUBMITTED]:
```
fabiencastan, line 947: this means a double loop. More simple to do the if inside the main loop.

cbentejac: Done

meshroom/core/node.py
cbentejac: Done: the double `for` loop has been removed, and the `else` from this suggestion has been replaced by an `elif chunk.status.status == Status.SUBMITTED` as it otherwise all the chunks would be reset, including those which are in the `SUCCESS` state.

fabiencastan (approved):

PR #2280, opened 2023-12-21
Update `ListAttributes` identically when removing edges or nodes

cbentejac:
## Description

The behaviour regarding `ListAttributes` was different depending on whether an edge or a node was removed:
- if an edge was removed, the `ListAttribute` that was the destination of that edge had its corresponding element completely removed;
- if a node was removed, the `ListAttribute` that was the destination of one of the node's edges had its corresponding element reset, but not removed.

With this behaviour, a user had different UIDs depending on whether a single edge or the whole node had been removed where the UID should always be identical.

This PR ensures that removing a node has the same effect on a connected node as removing an edge: it thus removes completely an element connected to a `ListAttribute` when the node is removed.

## Implementation remarks

When collecting the outgoing edges, we now check whether the edge is connected to a `ListAttribute`. If it is, we associate the attribute of the outgoing edge to a tuple containing the `ListAttribute` it is connected to and its value (if it has any). We then remove the corresponding element in the `ListAttribute` the edge is connected to.

When recreating a node that has been removed (either by performing an "Undo" action or by upgrading a node), this map is used to check whether the edge that will be re-added is to be connected to a `ListAttribute`. If it is, a new element is added to the `ListAttribute` with the value stored in the map. Re-creating the entry in the list is necessary: if it does not exist, the edge will not have any destination attribute to connect to.

### DIFF
--- a/meshroom/core/graph.py
+++ b/meshroom/core/graph.py
@@ -645,18 +645,38 @@ class Graph(BaseObject):
     @changeTopology
     def removeNode(self, nodeName):
         """
-        Remove the node identified by 'nodeName' from the graph
-        and return in and out edges removed by this operation in two dicts {dstAttr.getFullNameToNode(), srcAttr.getFullNameToNode()}
+        Remove the node identified by 'nodeName' from the graph.
+        Returns:
+            - a dictionary containing the incoming edges removed by this operation: {dstAttr.getFullNameToNode(), srcAttr.getFullNameToNode()}
+            - a dictionary containing the outgoing edges removed by this operation: {dstAttr.getFullNameToNode(), srcAttr.getFullNameToNode()}
+            - a dictionary containing the values, indices and keys of attributes that were connected to a ListAttribute prior to the removal of all edges:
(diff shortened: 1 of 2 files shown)


meshroom/core/graph.py
```diff
@@ -645,18 +645,30 @@ def nodeOutEdges(self, node):
...
             for edge in self.nodeOutEdges(node):
                 self.removeEdge(edge.dst)
                 outEdges[edge.dst.getFullNameToNode()] = edge.src.getFullNameToNode()
+
+                # Remove the corresponding attributes from the ListAttributes instead of just emptying their values
+                if isinstance(edge.dst.root, ListAttribute):
+                    outListAttributes[edge.dst.getFullNameToNode()] = (edge.dst.root.getFullNameToNode(), edge.dst.value if edge.dst.value else None)
+                    index = edge.dst.root.index(edge.dst)
+                    edge.dst.root.remove(index)
+
```
cbentejac, line 671: Save index as well

fabiencastan (changes requested):
Need to preserve index in lists during undo

meshroom/core/graph.py
cbentejac: Done.

fabiencastan (approved):

PR #2282, opened 2023-12-21
Use the correct response file to display the graph of the Camera Response Function

cbentejac:
## Description

Following the changes introduced by https://github.com/alicevision/AliceVision/pull/1484, the response file is not always named "response.csv" anymore, but is instead named "response_{INTRINSIC_ID}.csv".

This PR updates the corresponding output attribute in the `LdrToHdrCalibration` node by adding the intrinsic ID, which is never fixed beforehand, to the string of the response file's path. 

Since the intrinsic ID must be resolved, the way the CSV file is provided to the Camera Response Graph widget is also modified: instead of providing the `LdrToHdrCalibration` node to the Camera Response Graph and retrieving the path of the response file within the widget, the path is retrieved and resolved before activating the Camera Response Graph, and is then passed as a parameter to the component.

## Features list

- [x] Add the intrinsic ID to the name of the output response CSV file to match with what is generated by the AliceVision `LdrToHdrCalibration` program.
- [x] Provide the resolved response file path to the Camera Response Graph instead of having the widget trying to retrieve it.

### DIFF
--- a/meshroom/nodes/aliceVision/LdrToHdrCalibration.py
+++ b/meshroom/nodes/aliceVision/LdrToHdrCalibration.py
@@ -1,4 +1,4 @@
-__version__ = "3.0"
+__version__ = "3.1"
 
 import json
 import math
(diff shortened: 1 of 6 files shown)


servantftechnicolor (approved):
I'm not sure to fully understand the code. However, the behavior at runtime is correct.

PR #2337, opened 2024-03-15
[nodes] New node MeshRemoveUnseenFaces

fabiencastan:
Remove triangles from the mesh when the vertices are not visible by any camera.

Based on https://github.com/alicevision/AliceVision/pull/1681


mh0g:
Tested using the 72 first frames of ALAB. The mesh went from 21K+ faces to less than 10k.
Inner faces were removed successfully.

mh0g (approved):

PR #2342, opened 2024-03-20
[ui] drag&drop: common behavior for graph editor and image gallery

Just-Kiel:
In Image Gallery :
- drop 1 .mg open the scene
- drop images either create new camera or augment the reconstruction

In Graph Editor :
- drop 1 .mg open the scene
- drop images create new camera at position of mouse


meshroom/ui/reconstruction.py
```diff
@@ -761,38 +775,49 @@ def handleFilesDrop(self, drop, cameraInit):
...
+                    Message(
+                        "Too much Meshroom Scenes",
+                        "You should only import 1 .mg file"
+                    )
+                )
+            else:
+                self.loadUrl(filesByType["meshroomScene"][0])
+
+
+        if not filesByType["images"] and not filesByType["videos"] and not filesByType["panoramaInfo"] and not filesByType["meshroomScene"]:
```
cbentejac, line 800:
```suggestion
        if not filesByType["images"] and not filesByType["videos"] and not filesByType["panoramaInfo"] and not filesByType["meshroomScenes"]:
```

meshroom/ui/reconstruction.py
```diff
@@ -761,38 +775,49 @@ def handleFilesDrop(self, drop, cameraInit):
...
+    def getFilesByTypeFromDrop(self, urls):
         """
 
         Args:
-            drop:
+            urls: list of filepaths
 
         Returns:
-            <images, otherFiles> List of recognized images and list of other files
+            {images, otherFiles}: Map of recognized images and list of other files
```
cbentejac, line 819:
```suggestion
            {images, videos, panoramaInfo, meshroomScenes, otherFiles}: Map containing the lists of paths for recognized images, videos, Meshroom scenes and other files.
```

meshroom/ui/reconstruction.py
```diff
@@ -801,7 +826,7 @@ def getFilesByTypeFromDrop(drop):
                 filesByType.extend(multiview.findFilesByTypeInFolder(localFile))
             else:
                 filesByType.addFile(localFile)
-        return filesByType
+        return {"images": filesByType.images, "videos": filesByType.videos, "panoramaInfo": filesByType.panoramaInfo, "meshroomScene": filesByType.meshroomScene, "other": filesByType.other}
```
cbentejac, line 829:
```suggestion
        return {"images": filesByType.images, "videos": filesByType.videos, "panoramaInfo": filesByType.panoramaInfo, "meshroomScene": filesByType.meshroomScenes, "other": filesByType.other}
```

meshroom/ui/qml/ImageGallery/ImageGallery.qml
```diff
@@ -438,7 +441,16 @@ Panel {
                 anchors.fill: parent
                 enabled: !m.readOnly && !intrinsicsFilterButton.checked
                 keys: ["text/uri-list"]
-                // TODO: onEntered: call specific method to filter files based on extension
+                onEntered: {
+                    isMeshroomScene = false
+                    nbFilesDropped = drag.urls.length
```
cbentejac, line 446: Not a very accurate variable name since no file has been dropped yet. As we're still in the "dragging" phase, `nbDraggedFiles` might be more explicit.

meshroom/ui/qml/ImageGallery/ImageGallery.qml
```diff
@@ -463,7 +475,17 @@ Panel {
...
                         horizontalAlignment: Text.AlignHCenter
                         verticalAlignment: Text.AlignVCenter
-                        text: "Add Images"
+                        text: {
+                            if (isMeshroomScene) {
+                                if(nbFilesDropped == 1) {
+                                    return "Load Project"
+                                } else {
+                                    return "Only one project"
+                                }
```
cbentejac, line 484: This `if/else` can never be fully entered: if `isMeshroomScene` is true, then the number of files is necessarily 1 already.

So the "Only one project" can never be reached. The current behaviour is that nothing happens when hovering with more than 1 mg file. It might be better to count the number of .mg files / not .mg files contained in the dragging: if at least one file is not an .mg one, the counter of .mg file is reset, and that counter is what the message when hovering is based on.

meshroom/ui/qml/ImageGallery/ImageGallery.qml
```diff
@@ -463,7 +475,17 @@ Panel {
...
+                            }else if (!isMeshroomScene){
```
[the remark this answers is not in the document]
cbentejac: Couldn't this be just an "else {}", actually?

cbentejac (changes requested):
Aside from the comment I made regarding the if/else in the Image Gallery, everything seems to be working correctly and most of the comments are merely cosmetic.

cbentejac (approved):

PR #2346, opened 2024-03-25
Custom loader for .pc.ply pointcloud

servantftechnicolor:
Load PLY as pointcloud for 3D viewer if the extension is .pc.ply

See https://github.com/alicevision/AliceVision/pull/1689


mh0g:
to support display on node click, we can add ".pc.ply" [the list of supported extentions](https://github.com/alicevision/Meshroom/blob/7a3babf5f0574bc48c7677e9e0ee1d5603312d7b/meshroom/ui/qml/Viewer3D/Viewer3DSettings.qml#L16)

cbentejac (approved):

PR #2350, opened 2024-03-27
[core] Add new type of ChoiceParam that changes dynamically

Just-Kiel:
Add new type of ChoiceParam that changes dynamically according to other values.

When value of an attribute is changed onAttributeChanged is called, allowing to have unique reaction within node files.

Also add of callDesc function to be able to have other functions such as onNodeCreated at creation of node.


meshroom/core/desc.py
```diff
@@ -313,12 +313,11 @@ def __init__(self, name, label, description, value, values, exclusive, uid, grou
...
-    def conformValue(self, val):
+    def conformValue(self, value):
         """ Conform 'val' to the correct type and check for its validity """
-        val = self._valueType(val)
-        if val not in self.values:
-            raise ValueError('ChoiceParam value "{}" is not in "{}".'.format(val, str(self.values)))
-        return val
+        if not isinstance(value, self._valueType):
+            raise ValueError('ChoiceParam value should be a {} (param:{}, value:{}, type:{})'.format(self._valueType, self.name, value, type(value)))
+        return value
```
fabiencastan, line 320:
```suggestion
        return self._valueType(value)
```

meshroom/core/attribute.py
```diff
@@ -392,6 +399,43 @@ def wrapper(attr, *args, **kwargs):
...
+    def __init__(self, node, attributeDesc, isOutput, root=None, parent=None):
+        super(ChoiceParam, self).__init__(node, attributeDesc, isOutput, root, parent)
+        self._values = None
+
+    def getValues(self):
+        return self._values if self._values is not None else self.desc._values
+
+    def conformValue(self, val):
+        """ Conform 'val' to the correct type and check for its validity """
+        return self.desc._valueType(val)
```
fabiencastan, line 413:
```suggestion
        return self.desc.conformValue(val)
```

meshroom/core/desc.py
fabiencastan: We just need to remove the check that the value should be in the list of possible values.

fabiencastan (approved):
Looks good to me, just one minor thing to fix.

meshroom/core/attribute.py
Just-Kiel: Done

meshroom/core/desc.py
Just-Kiel: This cannot be done as such because some types cannot be cast to others (e.g. if the user provides a string for a ChoiceParam that expects integers, the cast will fail and cause errors).

cbentejac (approved):

PR #2351, opened 2024-03-27
[nodes] New SelectConnectedViews node

fabiencastan:
Select Connected Views based on SfM landmarks.

Based on https://github.com/alicevision/AliceVision/pull/1692


mh0g (approved):
seems to work as expected (tested on alab)

PR #2358, opened 2024-04-02
[ui] Add new FilterComboBox for ChoiceParam attributes

Just-Kiel:
Filter ComboBox created as single component and used in AttributeEditor.

It has a filter text section making the results sorted by relevance (first if starting by the filter text and after if includes it). Filter TextArea becomes red when no results are found and display all the original list.

Linked to https://github.com/alicevision/Meshroom/pull/2350 to have a better UI for it.


meshroom/ui/qml/Controls/FilterComboBox.qml
```diff
@@ -0,0 +1,111 @@
...
+                id: filterTextArea
+                leftPadding: 12
+                anchors.left: parent.left
+                anchors.right: parent.right
+                anchors.top: parent.top
+
+                selectByMouse: true
+                hoverEnabled: true
+                wrapMode: TextEdit.WrapAnywhere
+                placeholderText: "Filter"
```
fabiencastan, line 72: I'll avoid using a placeholder to keep things light.

cbentejac: I feel like the placeholder's quite useful to understand that you can type in the field. It doesn't seem that obvious if you take it away.

cbentejac (approved):

PR #2364, opened 2024-04-09
New InputNode for nodes without computation and support for all param types in output (and no more limited to File type)

Just-Kiel:
Node without computation implemented with creation of InputNode class for easy description of Node.

Also changes for UI :
- Hide buttons in menu for Compute, Submit...
- Different header color to see the difference
- Tabs depending on chunks hidden
- Output Attributes in Viewer2D doesn't include "Image gallery" when displayedNode is not computable

### DIFF
--- a/meshroom/core/desc.py
+++ b/meshroom/core/desc.py
@@ -144,11 +144,11 @@ class GroupAttribute(Attribute):
             # invalidKeys = set(value.keys()).difference([attr.name for attr in self._groupDesc])
             # if invalidKeys:
             #     raise ValueError('Value contains key that does not match group description : {}'.format(invalidKeys))
-            if self._groupDesc:
+            if self._groupDesc and value.keys():
                 commonKeys = set(value.keys()).intersection([attr.name for attr in self._groupDesc])
                 if not commonKeys:
-                    raise ValueError('Value contains no key that matches with the group description: {}'.format(commonKeys))
-        elif isinstance(value, (list, tuple)):
+                    raise ValueError(f'Value contains no key that matches with the group description (name={self.name}, values={value.keys()}, desc={[attr.name for attr in self._groupDesc]})')
+        elif isinstance(value, (list, tuple, set)):
             if len(value) != len(self._groupDesc):
                 raise ValueError('Value contains incoherent number of values: desc size: {}, value size: {}'.format(len(self._groupDesc), len(value)))
         else:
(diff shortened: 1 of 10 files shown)


meshroom/core/node.py
```diff
@@ -1210,6 +1213,7 @@ def has3DOutputAttribute(self):
     globalExecMode = Property(str, globalExecMode.fget, notify=globalExecModeChanged)
     isExternal = Property(bool, isExtern, notify=globalExecModeChanged)
     isComputed = Property(bool, _isComputed, notify=globalStatusChanged)
+    isComputable = Property(bool, lambda self: self.getGlobalStatus() != Status.INPUT, notify=globalStatusChanged)
```
cbentejac, line 1216: Correct but I'd rather have the lambda in a separate function instead, like we're doing for all the other properties.

```
def _isComputable(self):
    return self.getGlobalStatus() != Status.INPUT

...

isComputable = Property(bool, _isComputable, notify=globalStatusChanged)
```

meshroom/core/taskManager.py
```diff
@@ -333,8 +333,11 @@ def checkNodesDependencies(self, graph, toNodes, context):
         """
         ready = []
         computed = []
+        inputNodes = []
         for node in toNodes:
-            if context == "COMPUTATION":
+            if node.isComputable == False:
```
cbentejac, line 338:
```suggestion
            if not node.isComputable:
```

meshroom/core/desc.py
```diff
@@ -587,6 +587,13 @@ def stopProcess(self, chunk):
     def processChunk(self, chunk):
         raise NotImplementedError('No processChunk implementation on node: "{}"'.format(chunk.node.name))
 
+class InputNode(Node):
```
cbentejac, line 590: Not entirely satisfied with using "InputNode" for non-computable nodes because I feel like there could be other use cases, but I unfortunately don't have a much better idea to designate them 😒

cbentejac (changes requested):
Tested, works well, and the whole implementation is clean. 

Tiny cosmetic changes to do, and see if we can come up with something that's less oriented than "InputNode" for the class' name.

meshroom/core/desc.py
Just-Kiel: What do you think of "ValueNode" ?
I feel like that it might better because it would not be necessarily in input.

fabiencastan (approved):

PR #2366, opened 2024-04-11
[ui] Right click on text element in AttributeEditor open Copy/Paste menu

Just-Kiel:
Right click on text element in AttributeEditor open Copy/Paste menu
- Able to paste nodeContent in text element

### DIFF
--- a/meshroom/ui/qml/GraphEditor/AttributeItemDelegate.qml
+++ b/meshroom/ui/qml/GraphEditor/AttributeItemDelegate.qml
@@ -112,6 +112,21 @@ RowLayout {
                                 updateAttributeLabel()
                             }
                         }
+                        MenuItem {
+                            text: "Copy"
+                            enabled: attribute.value != ""
+                            onTriggered: {
+                                Clipboard.clear()
+                                Clipboard.setText(attribute.value)
+                            }
+                        }
+                        MenuItem {
+                            text: "Paste"
+                            enabled: Clipboard.getText() != "" && root.editable
+                            onTriggered: {
+                                _reconstruction.setAttribute(attribute, Clipboard.getText())
+                            }
+                        }
 
                         MenuSeparator {
                             visible: paramMenu.isFileAttribute
@@ -236,7 +251,6 @@ RowLayout {
                         else if (drop.hasText && drop.text != '')
                             setTextFieldAttribute(drop.text)
                     }
-
                 }
                 MouseArea {
                     anchors.fill: parent
--- a/meshroom/ui/qml/GraphEditor/AttributeItemDelegate.qml
+++ b/meshroom/ui/qml/GraphEditor/AttributeItemDelegate.qml
@@ -211,9 +211,9 @@ RowLayout {
                 text: attribute.value
                 selectByMouse: true
                 onEditingFinished: setTextFieldAttribute(text)
+                persistentSelection: true
                 onAccepted: {
                     setTextFieldAttribute(text)
-                    root.forceActiveFocus()
                 }
                 Component.onDestruction: {
                     if (activeFocus)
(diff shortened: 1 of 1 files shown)


meshroom/ui/qml/GraphEditor/AttributeItemDelegate.qml
```diff
@@ -222,6 +223,38 @@ RowLayout {
...
+                            text: "Copy"
+                            enabled: attribute.value != ""
+                            onTriggered: {
+                                Clipboard.clear()
+                                Clipboard.setText(attribute.value)
+                            }
+                        }
+                        MenuItem {
+                            text: "Paste"
+                            enabled: Clipboard.getText() != ""
```
cbentejac, line 252: Tiny remark here: the "paste" menu should not only appear if the clipboard is not empty, it should also appear if the text field we're right-clicking on is not read-only.

With the current behaviour, we can right-click on an output attribute and have the "paste" menu appearing, but when pasting there, nothing happens (which is the expected behaviour, since output File attributes are read-only). The "paste" menu should be disabled in that case.

cbentejac (dismissed):
The "paste" menu should be disabled even if the clipboard has a valid content when the target is a read-only text field.

Otherwise, everything works fine.

meshroom/ui/qml/GraphEditor/AttributeItemDelegate.qml
```diff
@@ -227,6 +228,38 @@ RowLayout {
...
+                        }
+                    }
+
+                    property Component menuCopy : Menu {
+                        MenuItem {
+                            text: "Copy"
+                            enabled: attribute.value != ""
+                            onTriggered: {
+                                Clipboard.clear()
+                                Clipboard.setText(attribute.value)
```
fabiencastan, line 252: Not tested:
```suggestion
                                textField.copy()
```

meshroom/ui/qml/GraphEditor/AttributeItemDelegate.qml
```diff
@@ -206,6 +206,7 @@ RowLayout {
         Component {
             id: textField_component
             TextField {
+                id: textField
```
fabiencastan, line 209: The problem comes from the `root.forceActiveFocus()` in the onAccepted event.

fabiencastan (changes requested):

meshroom/ui/qml/GraphEditor/AttributeItemDelegate.qml
```diff
@@ -227,6 +228,38 @@ RowLayout {
                         else if (drop.hasText && drop.text != '')
                             setTextFieldAttribute(drop.text)
                     }
+
+                }
+                MouseArea {
+                    anchors.fill: parent
+                    acceptedButtons: Qt.AllButtons
```
fabiencastan, line 235:
```suggestion
                    acceptedButtons: Qt.RightButton
```

meshroom/ui/qml/GraphEditor/AttributeItemDelegate.qml
```diff
@@ -227,6 +228,38 @@ RowLayout {
...
+                    onClicked: {
+                        if (mouse.button == Qt.RightButton) {
+                            var menu = menuCopy.createObject(parameterLabel)
+                            menu.parent = parameterLabel
+                            menu.popup()
```
fabiencastan, line 240: This kills the text selection, I'm not sure how to fix that.

fabiencastan: maybe with persistentSelection...

meshroom/ui/qml/GraphEditor/AttributeItemDelegate.qml
```diff
@@ -227,6 +228,38 @@ RowLayout {
...
                             setTextFieldAttribute(drop.text)
                     }
+
+                }
+                MouseArea {
+                    anchors.fill: parent
+                    acceptedButtons: Qt.AllButtons
+                    onClicked: {
+                        if (mouse.button == Qt.RightButton) {
+                            var menu = menuCopy.createObject(parameterLabel)
```
fabiencastan: parent should be textField

fabiencastan (approved):

PR #2369, opened 2024-04-12
[core/ui] Hide output attributes flagged for visualisation

Just-Kiel:
Hide output attributes flagged for visualisation when specifying visible in the description of the node.
True is for seeing it and False is for hiding the output attribute in the node in the GraphEditor.


mh0g (approved):
Tested on a custom node. Masked some output attributes, as expected they are hidden in the node but still are behaving adequately (the display still work).

cbentejac (approved):

PR #2375, opened 2024-04-16
[core/ui] Load image sequence from node's output in SequencePlayer

Just-Kiel:
- Image sequence is detected because the semantic of the output is defined as "sequence" in the description of the node
- SequencePlayer is enabled when double-clicked on the node

- Need to set the path for the image sequence is : `/path/to/define/<FRAMEID>.extension`

- Unsync of Viewers and SequencePlayer when displayed output is sequence output

### DIFF
--- a/meshroom/core/node.py
+++ b/meshroom/core/node.py
@@ -1289,7 +1289,7 @@ class BaseNode(BaseObject):
         Return True if at least one attribute has the 'sequence' semantic (and can thus be loaded in the 2D Viewer), False otherwise.
         """
         for attr in self._attributes:
-            if attr.enabled and attr.isOutput and attr.desc.semantic == "sequence":
+            if attr.enabled and attr.isOutput and (attr.desc.semantic == "sequence" or attr.desc.semantic == "imageList"):
                 return True
         return False
 
--- a/meshroom/core/node.py
+++ b/meshroom/core/node.py
@@ -1284,6 +1284,15 @@ class BaseNode(BaseObject):
                 return True
         return False
 
+    def hasSequenceOutputAttribute(self):
+        """
+        Return True if at least one attribute has the 'sequence' semantic (and can thus be loaded in the 2D Viewer), False otherwise.
+        """
+        for attr in self._attributes:
+            if attr.enabled and attr.isOutput and attr.desc.semantic == "sequence":
+                return True
+        return False
+
     def has3DOutputAttribute(self):
         """
         Return True if at least one attribute is a File that can be loaded in the 3D Viewer, False otherwise.
@@ -1345,6 +1354,7 @@ class BaseNode(BaseObject):
 
     outputAttrEnabledChanged = Signal()
     hasImageOutput = Property(bool, hasImageOutputAttribute, notify=outputAttrEnabledChanged)
+    hasSequenceOutput = Property(bool, hasSequenceOutputAttribute, notify=outputAttrEnabledChanged)
     has3DOutput = Property(bool, has3DOutputAttribute, notify=outputAttrEnabledChanged)
 
 class Node(BaseNode):
--- a/meshroom/ui/components/filepath.py
+++ b/meshroom/ui/components/filepath.py
@@ -4,6 +4,7 @@ from PySide2.QtCore import QUrl, QFileInfo
 from PySide2.QtCore import QObject, Slot
 
 import os
+import glob
 
 
 class FilepathHelper(QObject):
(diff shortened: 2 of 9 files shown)


meshroom/core/node.py
```diff
@@ -1220,6 +1220,7 @@ def has3DOutputAttribute(self):
 
     outputAttrEnabledChanged = Signal()
     hasImageOutput = Property(bool, hasImageOutputAttribute, notify=outputAttrEnabledChanged)
+    hasSequenceOutput = Property(bool, lambda self: any(attr.enabled and attr.isOutput and attr.desc.semantic == "sequence" for attr in self._attributes), notify=outputAttrEnabledChanged)
```
cbentejac, line 1223: Define a function (even if it's simple) for the getter instead of using a lambda in the property definition.

meshroom/ui/components/filepath.py
```diff
@@ -98,3 +99,17 @@ def globFirst(self, path):
     def fileSizeMB(self, path):
         """ Returns the file size in MB. """
         return QFileInfo(self.asStr(path)).size() / (1024*1024)
+
+    @Slot(str, result="QVariantList")
+    @Slot(str, str, result="QVariantList")
+    def getFilenamesFromFolder(self, folder_path: str, extension: str = None):
```
cbentejac, line 105:
```suggestion
    def getFilenamesFromFolder(self, folderPath: str, extension: str = None):
```

`folder_path` respects snake_case which is usually what's expected in Python... but no what we're using here, starting with the method's name.

cbentejac (dismissed):

meshroom/ui/qml/Viewer/Viewer2D.qml
```diff
@@ -317,12 +343,21 @@ FocusScope {
                 if (attr.isOutput && attr.desc.semantic === "image" && attr.enabled) {
                     names.push(attr.name)
                 }
+
+                if (attr.isOutput && attr.desc.semantic === "sequence" && attr.enabled) {
+                    names.push(attr.name)
+                }
```
fabiencastan, line 349: Test "image" or "sequence" in the same if.

meshroom/ui/qml/Viewer/Viewer2D.qml
```diff
@@ -225,25 +228,60 @@ FocusScope {
...
-        }
+        let attr = getAttributeByName(displayedNode, outputAttribute.name)
 
-        let resolved = path;
-        for (let key in replacements) {
-            resolved = resolved.replace(key, replacements[key])
-        }
+        let replacements = {}
+        if (vp == null) {
+            replacements = Filepath.getFilenamesFromFolder(Filepath.dirname(path), Filepath.extension(path))
```
fabiencastan, line 235: we should not re-parse the folder at each frame

meshroom/ui/qml/Viewer/Viewer2D.qml
```diff
@@ -225,25 +228,60 @@ FocusScope {
...
+                    let splitAfter = splitBefore[1].split(")")
+                    let id = replacements[i].match("["+splitBefore[0]+"]("+splitAfter[0]+")["+splitAfter[1]+"]")
+                    if (id) {
+                        // put id in replacements as key of replacements[i]
+                        ids.push(id[1])
+                    }
+                }
+                ids.sort()
+                frameRange.min = ids[0]
+                frameRange.max = ids[ids.length-1]
```
fabiencastan, line 253: should be done on the python side

meshroom/ui/qml/Viewer/Viewer2D.qml
```diff
@@ -539,7 +513,7 @@ FocusScope {
                                 'canBeHovered': false,
                                 'idView': Qt.binding(function() { return ((root.displayedNode && !root.displayedNode.hasSequenceOutput && _reconstruction) ? _reconstruction.selectedViewId : -1) }),
                                 'cropFisheye': false,
-                                'sequence': Qt.binding(function() { return ((root.enableSequencePlayer && (_reconstruction || (root.displayedNode && root.displayedNode.hasSequenceOutput))) ? getSequence() : []) }),
+                                'sequence': Qt.binding(function() { root.sequence = getSequence(); return ((root.enableSequencePlayer && (_reconstruction || (root.displayedNode && root.displayedNode.hasSequenceOutput))) ? getSequence() : []) }),
```
fabiencastan, line 516: you are not supposed to assign a variable here (root.sequence)

meshroom/ui/components/filepath.py
```diff
@@ -103,18 +106,54 @@ def fileSizeMB(self, path):
...
+    @Slot(str, result="QVariantList")
+    def resolveSequence(self, path):
+        """
+        Get id of each file in the sequence.
+        """
+        # use of pyseq to get the sequences
+        seq = pyseq.get_sequences(self.asStr(path))
+
+        ids = [[s.start(), s.end()] for s in seq]
+        resolved = [[os.path.join(self.dirname(path), s.head()+str(s.format("%p")%frame)+s.tail()) for frame in s.frames()] for s in seq]
```
fabiencastan, line 158:
```suggestion
        folder = self.dirname(path)
        resolved = [[os.path.join(folder, s.head() + str(s.format("%p")%frame) + s.tail()) for frame in s.frames()] for s in seq]
```

fabiencastan:
Need to update requirements with the new dependency.

fabiencastan (approved):

PR #2377, opened 2024-04-18
[Viewer] fix lens distortion viewer status when switching between projects

elyasbny:
## Description

When opening an empty pipeline while the lens distortion viewer was enabled, we ended up in a state where the distortion viewer remained enabled and could not be disabled anymore as its button was greyed out.

### DIFF
--- a/meshroom/ui/qml/Viewer/Viewer2D.qml
+++ b/meshroom/ui/qml/Viewer/Viewer2D.qml
@@ -1096,6 +1096,9 @@ FocusScope {
                         }
                         MaterialToolButton {
                             id: displayLensDistortionViewer
+
+                            property int numberChanges: null
+                            property bool previousChecked: null
                             property var activeNode: root.aliceVisionPluginAvailable && _reconstruction ? _reconstruction.activeNodes.get('sfmData').node : null
                             property bool isComputed: {
                                 if (!activeNode)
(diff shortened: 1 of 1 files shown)


meshroom/ui/qml/Viewer/Viewer2D.qml
```diff
@@ -1096,6 +1096,9 @@ FocusScope {
                         }
                         MaterialToolButton {
                             id: displayLensDistortionViewer
+
+                            property int number_changes: null
+                            property bool previous_checked: null
```
cbentejac, line 1101: Variable names should ideally respect the case (here, camelCase) that's used across the file.

```suggestion
                            property int numberChanges: null
                            property bool previousChecked: null
```

meshroom/ui/qml/Viewer/Viewer2D.qml
```diff
@@ -1127,6 +1130,28 @@ FocusScope {
...
+                            onEnabledChanged: {
+                                if (!enabled) {
+                                    previous_checked = checked
+                                    checked = false
+                                    number_changes = 0
+            
+                                }
+
+                                if (enabled && (number_changes == 1) && previous_checked) {
+                                    console.warn("activeNode changed to " + number_changes)
```
cbentejac, line 1150: Debugging logs should be removed here.

cbentejac (changes requested):

cbentejac (approved):

PR #2378, opened 2024-04-19
[ui] Fix drag and drop of heavy number of frames

Just-Kiel:
## Description
Fix drag and drop of heavy number of frames, now should display the overlay saying if drop is possible faster.


servantftechnicolor (approved):
OK

PR #2382, opened 2024-04-24
[core/ui] Add support for PushButton attribute

Just-Kiel:
## Description
Push Button is a new type of attribute for nodes. 
- They are described as PushButtonParam and do not need value. Value is set to "" to not have DescriptionConflict and be consistent with other types of attribute.
- Definition of `on{Attribute}Clicked` function available in the description of the node to have a custom behavior.

### DIFF
--- a/meshroom/core/attribute.py
+++ b/meshroom/core/attribute.py
@@ -6,6 +6,7 @@ import re
 import weakref
 import types
 import logging
+import inspect
 
 from collections.abc import Iterable, Sequence
 from string import Template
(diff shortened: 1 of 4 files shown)


meshroom/core/attribute.py
```diff
@@ -398,6 +425,21 @@ def wrapper(attr, *args, **kwargs):
...
+    def __init__(self, node, attributeDesc, isOutput, root=None, parent=None):
+        super(PushButtonParam, self).__init__(node, attributeDesc, isOutput, root, parent)
+    
+    def getExportValue(self):
+        """ Return the function defined of the PushButtonParam """
+        paramName = self.name[:1].upper() + self.name[1:]
+        methodName = f'on{paramName}Clicked'
+        if hasattr(self.node.nodeDesc, methodName):
+            return inspect.getsource(getattr(self.node.nodeDesc, methodName))
+        return f'def {methodName}:\n pass'
```
fabiencastan, line 438:
```suggestion
    def getExportValue(self):
        """ Nothing to export for a PushButton """
        return ""
```

meshroom/core/node.py
```diff
@@ -1691,6 +1705,13 @@ def nodeFactory(nodeDict, name=None, template=False, uidConflict=False):
 
             # verify that all inputs match their descriptions
             for attrName, value in inputs.items():
+                paramName = attrName[:1].upper() + attrName[1:]
+                functionName = f'on{paramName}Clicked'
+                if hasattr(nodeDesc, functionName):
+                    m = inspect.getsource(getattr(nodeDesc, functionName))
+                    if value != m:
+                        compatibilityIssue = CompatibilityIssue.DescriptionConflict
+                        break
```
fabiencastan, line 1714: Nothing to check regarding compatibility.
```suggestion
```

fabiencastan (approved):

PR #2385, opened 2024-04-25
Viewer3D: Adjust bounding-box by moving faces

elyasbny:
## Description
Implemented this feature, one can now adjust the bounding-box by clicking on an moving the points which are in the center of the surfaces.

## Implementation remarks
I implemented two different entities, one for surfaces that are located in the direction of the axis, and one for the other surfaces. I did it, as it seemed to be a simple way to tackle the different behaviour of these surfaces when I wanted to move them. There is maybe a better way to do this.

### DIFF
--- a/meshroom/ui/qml/Viewer3D/TransformGizmo.qml
+++ b/meshroom/ui/qml/Viewer3D/TransformGizmo.qml
@@ -57,6 +57,11 @@ Entity {
         Z
     }
 
+    enum Direction {
+        Forward,
+        Backward
+    }
+
     enum Type {
         TRANSLATION,
         ROTATION,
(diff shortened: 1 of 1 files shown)


meshroom/ui/qml/Viewer3D/TransformGizmo.qml
```diff
@@ -541,6 +550,131 @@ Entity {
                 }
             }
 
+            // MOVE SURFACE ENTITY (surfaces on axis direction)
+            Entity {
+                id: surface1MoveEntity
```
servantftechnicolor, line 555: Can we rename both entities ids to something more explicit ? Also, can you add a line or two explaining what is a "surface on axis direction". It may be clear in your head, but not for the next reader.

servantftechnicolor (approved):

PR #2391, opened 2024-04-26
[ui] Bounding Box are usable in other nodes, not only Meshing

Just-Kiel:
## Description
Bounding boxes can be used in nodes, not only Meshing. It only needs to have an attribute with useBoundingBox name.


servantftechnicolor:
I asked this for lidarMeshing.

servantftechnicolor (approved):

PR #2395, opened 2024-04-29
[ui] Remove intrinsic if not used by any viewpoint

Just-Kiel:
## Description
When deleting image from the ImageGallery, if it is the last one using a specific intrinsic, it now removes the intrinsic.

### DIFF
--- a/meshroom/ui/qml/ImageGallery/ImageGallery.qml
+++ b/meshroom/ui/qml/ImageGallery/ImageGallery.qml
@@ -292,6 +292,24 @@ Panel {
 
                     function sendRemoveRequest() {
                         if (!readOnly) {
+                            // look if the viewpoint's intrinsic is used by another viewpoint
+                            var intrinsicId = object.value.get('intrinsicId').value
+                            var intrinsicUsed = false
+                            for (var i = 0; i < m.viewpoints.count; ++i) {
+                                if (m.viewpoints.at(i) != object && m.viewpoints.at(i).value.get("intrinsicId").value === intrinsicId) {
+                                    intrinsicUsed = true
+                                    break
+                                }
+                            }
+                            if (!intrinsicUsed) {
+                                //find the intrinsic and remove it
+                                for (var i = 0; i < m.intrinsics.count; ++i) {
+                                    if (m.intrinsics.at(i).value.get("intrinsicId").value === intrinsicId) {
+                                        _reconstruction.removeAttribute(m.intrinsics.at(i))
+                                        break
+                                    }
+                                }
+                            }
                             removeImageRequest(object)
                             // If the last image has been removed, make sure the viewpoints and intrinsics are reset
                             if (m.viewpoints.count === 0)


meshroom/ui/qml/ImageGallery/ImageGallery.qml
```diff
@@ -292,6 +292,24 @@ Panel {
 
                     function sendRemoveRequest() {
                         if (!readOnly) {
```
fabiencastan, line 294:
```
if(readOnly) {
    return
}
...
```

meshroom/ui/qml/ImageGallery/ImageGallery.qml
```diff
@@ -292,6 +292,24 @@ Panel {
...
+                            // look if the viewpoint's intrinsic is used by another viewpoint
```
fabiencastan, line 295: Should be done after the removeImageRequest, isn't it?

fabiencastan: Maybe better to do it on the python side.
Instead of a generic call to reconstruction.removeAttribute(attribute), here:
https://github.com/alicevision/Meshroom/blob/7b691a05dd73c1aa1ee913aa11af84ba81d227b5/meshroom/ui/qml/WorkspaceView.qml#L81C17-L81C80

Add a removeImage and do all the checks in it (including the job done in allViewpointsCleared / removeAllImages).
https://github.com/alicevision/Meshroom/blob/7b691a05dd73c1aa1ee913aa11af84ba81d227b5/meshroom/ui/graph.py#L820C9-L820C24

meshroom/ui/graph.py
```diff
@@ -816,6 +816,29 @@ def appendAttribute(self, attribute, value=QJsonValue()):
     def removeAttribute(self, attribute):
         self.push(commands.ListAttributeRemoveCommand(self._graph, attribute))
 
+    @Slot(Attribute)
+    def removeImage(self, image):
```
fabiencastan, line 820:
```suggestion
    def removeImage(self, image):
        with self.groupedGraphModification("Remove Image"):
```

meshroom/ui/graph.py
```diff
@@ -816,6 +816,29 @@ def appendAttribute(self, attribute, value=QJsonValue()):
...
+        intrinsicUsed = False
+        for intrinsic in self.cameraInit.attribute("viewpoints").getExportValue():
+            if image.getExportValue() != intrinsic and intrinsic['intrinsicId'] == intrinsicId:
+                intrinsicUsed = True
+                break
+
+        if not intrinsicUsed:
+            #find the intrinsic and remove it
+            for x in range(len(self.cameraInit.attribute("intrinsics"))):
+                intrinsic = self.cameraInit.attribute("intrinsics").at(x)
```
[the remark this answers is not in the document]
Just-Kiel: It is not possible to do so because ListAttribute is not iterable. That is why I decided to do with the range and then init the intrinsic variable.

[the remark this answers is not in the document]
Just-Kiel: Accessor implemented for ListAttribute so now it works better

meshroom/ui/qml/WorkspaceView.qml
```diff
@@ -78,7 +78,7 @@ Item {
                 cameraInit: reconstruction ? reconstruction.cameraInit : null
                 tempCameraInit: reconstruction ? reconstruction.tempCameraInit : null
                 cameraInitIndex: reconstruction ? reconstruction.cameraInitIndex : -1
-                onRemoveImageRequest: reconstruction.removeAttribute(attribute)
+                onRemoveImageRequest: reconstruction.removeImage(attribute)
                 onAllViewpointsCleared: { reconstruction.removeAllImages(); reconstruction.selectedViewId = "-1" }
```
fabiencastan, line 82: We should now remove the call to removeAllImages here, as it creates an empty element in the undo stack.

PR #2396, opened 2024-04-30
[ui/core] Fix get latest SfM node for previz

Just-Kiel:
## Description
Latest SfM node was not always the same because the edges of nodes weren't always in the same order. To solve that we decided to order the edges and then we obtain the correct node for previz.


fabiencastan:
Better to keep self.edges ordered when we modify it:
https://github.com/alicevision/Meshroom/blob/4e8c0c2cf891b1338fc08635e7a3c5036d20a3d9/meshroom/core/graph.py#L879
Instead of doing it at each call of the getter.
Also better to keep the order of the node attributes instead of the alphabetical order.

fabiencastan (approved):

PR #2399, opened 2024-05-06
[ui] Cut option available in GraphEditor

Just-Kiel:
## Description
Cut option on nodes available in Graph Editor. Shortcut CTRL+X also added for cut option.

### DIFF
--- a/meshroom/ui/qml/main.qml
+++ b/meshroom/ui/qml/main.qml
@@ -559,6 +559,38 @@ ApplicationWindow {
     }
 
     Action {
+        id: cutAction
+
+        property string tooltip: {
+            var s = "Copy selected node"
+            s += (_reconstruction && _reconstruction.selectedNodes.count > 1 ? "s (" : " (") + getSelectedNodesName()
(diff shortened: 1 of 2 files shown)


fabiencastan:
I have not tested it, but looking at the code, it will generate 2 elements in the undo stack, isn't it? One action should generate only one.

meshroom/ui/qml/main.qml
```diff
@@ -558,6 +558,38 @@ ApplicationWindow {
...
+            return s
+        }
+        text: "Cut Node" + (_reconstruction && _reconstruction.selectedNodes.count > 1 ? "s " : " ")
+        enabled: _reconstruction ? _reconstruction.selectedNodes.count > 0 : false
+        onTriggered: {
+            graphEditor.copyNodes()
+            graphEditor.uigraph.removeNodes(graphEditor.uigraph.selectedNodes)
+        }
+
+        function getSelectedNodesName()
```
fabiencastan, line 577: Same function than for copy. You can move it outside of the Action and use the same function.

fabiencastan (approved):

PR #2401, opened 2024-05-14
[ui] Selected node header set to base color

Just-Kiel:
## Description
As we would like maybe to have color for categories of nodes, it is better if we don't hide those colors while selecting nodes. We still have the border but the header keeps its baseColor.


fabiencastan:
The mouse hover needs to use another color, to avoid the confusion with the selection (as the title bar does not make the difference anymore).

fabiencastan (approved):

PR #2407, opened 2024-05-22
Plugin System

mh0g:
## Description

This feature adds a plugin system to meshroom.
It also allows to install and run new nodes using containerization tools, such as  conda and docker.

## Features list
v1
- [X] Install button in the ui
- [X] Install from local folder
- [x] Install from git 
- [x] Camelify code
- [x] Add support for pipelines
- [x] New 'build status' for nodes that are building their environment
- [x] .json to customize plugin install
- [x] Support multiple env per plugin
- [x] Documentation to create plugins
- [x] Node uninstallation
- [x] Catalog logic 
- [x] PipNode to install pip packages in meshroom's python
- [x] PythonVirtualEnvNode taking requirements.txt 
--------------
v2
- [x] Changing the logic from env class to env attribute 
- [x] Modify process() in node to run normally if no or pip env, or conda, venv, docker, rez via command line meshroom_compute, if specified
- [x] Added badge and warning to build
- [x] Access to meshroom within envs

-------------
v3
- [x] support for rez packages
- [x] autorefresh

## Implementation remarks

The logic is in meshroom/core/plugin.py .
It defines a new node classes to run python code in environments/containers using meshroom_compute.
Added bandit file to not test the shell call errors in codefactor.
Uses the CI en var to skip the tests on CI.

-------------
Future needed work: ui part to list/uninstall catalog, multiples envs per node, button to delete env


meshroom/ui/qml/main.qml
Just-Kiel, line 1: Maybe should use Platform.FileDialog instead of Dialog to prevent from slow and laggy behavior

Just-Kiel:
After some tests, it seems that the logging system with the log file in MeshroomCache is not working anymore. I think it is not normal and should be fixed.

Just-Kiel:
When submitting a node (for the moment is failing at the end), when status is BUILD, the node is unlocked and we can Compute and Submit in the UI.

Just-Kiel:
Warn that when trying to compute you need to save the file (even if had previous save, needs to save the latest modifs), because it throws an error without being explicit.

mh0g:
> After some tests, it seems that the logging system with the log file in MeshroomCache is not working anymore. I think it is not normal and should be fixed. Seems to be related because while doing processes with os.system and not popen, log is lost.

solved!

Just-Kiel:
Could it be great to have "Delete Environment" button ? Like the "Delete Data"

mh0g:
> When submitting a node (for the moment is failing at the end), when status is BUILD, the node is unlocked and we can Compute and Submit in the UI. Same error if status is FIRST_RUN (and/or computting) we can launch process again.

solved

mh0g:
Todo: add a flag to the CI so we dont run the unit tests in it

mh0g:
> Could it be great to have "Delete Environment" button ? Like the "Delete Data"

added in todos for next version. Could be added to the "installed plugins" page

mh0g:
Issues with meshroom not being found in virtualenv node

Just-Kiel:
After few tests "Stop Computation" seems to have disappeared from all kind of node (which is not normal)

meshroom/core/__init__.py
```diff
@@ -30,12 +30,19 @@
...
 pipelineTemplates = {}
 
+#meshroom paths
+meshroomFolder = os.path.dirname(os.path.dirname(__file__))
+cacheFolderName = 'MeshroomCache'
+defaultCacheFolder = os.environ.get('MESHROOM_CACHE', os.path.join(tempfile.gettempdir(), cacheFolderName))
+
+#plugin paths
+pluginsNodesFolder = os.path.join(meshroomFolder, "plugins")
+pluginsPipelinesFolder = os.path.join(meshroomFolder, "pipelines")
```
Just-Kiel, line 44: It doesn't access to the "pipelines" folder included in a plugin and so you can't never access the custom pipelines.

meshroom/core/desc.py
```diff
@@ -658,9 +658,53 @@ class Node(object):
...
+    def isPlugin(self):
+        """
+        Tests if the node is a valid plugin node
+        """
+        return self._isPlugin
+
+    @property
+    def isBuilt(self):
+        """
+        Tests if the envirronement is built
```
fabiencastan, line 704: typo

tests/test_plugin_nodes.py
```diff
@@ -0,0 +1,17 @@
+import logging
+import os
+
+from meshroom.core.graph import Graph
+
+logging = logging.getLogger(__name__)
+
+def test_pluginNodes():
+    if "CI" in os.environ:
+        return
```
fabiencastan, line 10: a comment to explain why it's disabled in the CI

almarouk:
Would it be possible to add support for non-default location for conda environments? i.e. using "--prefix PATH" instead of "--name NAME".
In that case, checks for the existence of the environment need to be adapted correspondingly, notably in the following:
https://github.com/alicevision/Meshroom/blob/24c4585878bf6682a022f8db5a3baee77d22bdf5/meshroom/core/plugin.py#L90
would become: `cmd = condaBin+" list --prefix "+envName`, or alternatively by checking the result of `conda env list`.

fabiencastan:
New implementation divided into several stages starting with https://github.com/alicevision/Meshroom/pull/2703 and https://github.com/alicevision/Meshroom/pull/2733

PR #2412, opened 2024-05-24
[ui] Conversion of relative paths to absolute ones

Just-Kiel:
## Description
Project, import and import recursive in meshroom command are now absolute paths.
Prevent from having some files gone while reopening the scene not from the same folder.

### DIFF
--- a/meshroom/multiview.py
+++ b/meshroom/multiview.py
@@ -130,6 +130,7 @@ def findFilesByTypeInFolder(folder, recursive=False):
 
     output = FilesByType()
     for currentFolder in inputFolders:
+        currentFolder = os.path.abspath(currentFolder)
         if os.path.isfile(currentFolder):
             output.addFile(currentFolder)
             continue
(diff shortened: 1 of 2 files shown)


meshroom/ui/reconstruction.py
```diff
@@ -836,6 +836,7 @@ def importImagesFromFolder(self, path, recursive=False):
             recursive: List files in folders recursively.
 
         """
+        path = [os.path.abspath(p) for p in path]
         logging.debug("importImagesFromFolder: " + str(path))
         filesByType = multiview.findFilesByTypeInFolder(path, recursive)
```
fabiencastan, line 841: path could be a folder, file or list of files, so this does not work.

fabiencastan: should be done inside findFilesByTypeInFolder

fabiencastan (approved):

PR #2428, opened 2024-06-07
[nodes] Fix outputImages if no export in ExportAnimatedCamera

Just-Kiel:
## Description
Fix outputImages if no export in ExportAnimatedCamera

If exportUndistortedImages is false, we do not want outputImages attribute.

### DIFF
--- a/meshroom/core/node.py
+++ b/meshroom/core/node.py
@@ -750,18 +750,20 @@ class BaseNode(BaseObject):
             # Apply expressions for File attributes
             if attr.attributeDesc.isExpression:
                 defaultValue = ""
-                try:
-                    defaultValue = attr.defaultValue()
-                except AttributeError as e:
-                    # If we load an old scene, the lambda associated to the 'value' could try to access other params that could not exist yet
-                    logging.warning('Invalid lambda evaluation for "{nodeName}.{attrName}"'.format(nodeName=self.name, attrName=attr.name))
-                try:
-                    attr.value = defaultValue.format(**self._cmdVars)
-                    attr._invalidationValue = defaultValue.format(**cmdVarsNoCache)
-                except KeyError as e:
-                    logging.warning('Invalid expression with missing key on "{nodeName}.{attrName}" with value "{defaultValue}".\nError: {err}'.format(nodeName=self.name, attrName=attr.name, defaultValue=defaultValue, err=str(e)))
-                except ValueError as e:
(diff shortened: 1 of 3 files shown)


meshroom/nodes/aliceVision/ExportAnimatedCamera.py
```diff
@@ -117,9 +117,10 @@ class ExportAnimatedCamera(desc.AVCommandLineNode):
             name="outputImages",
             label="Undistorted Images",
             description="Output undistorted images.",
-            value=desc.Node.internalFolder + "undistort/" + "<INTRINSIC_ID>_<FILESTEM>.{undistortedImageTypeValue}",
+            value=lambda attr: desc.Node.internalFolder + "undistort/" + "<INTRINSIC_ID>_<FILESTEM>.{undistortedImageTypeValue}" if attr.node.exportUndistortedImages.value else "",
```
fabiencastan, line 120: adding "enabled" should be enough, we don't need to change the "value", it can stay static.

meshroom/core/node.py
```diff
@@ -750,18 +750,20 @@ def _buildAttributeCmdVars(cmdVars, name, attr):
...
-                    # If we load an old scene, the lambda associated to the 'value' could try to access other params that could not exist yet
-                    logging.warning('Invalid lambda evaluation for "{nodeName}.{attrName}"'.format(nodeName=self.name, attrName=attr.name))
-                try:
-                    attr.value = defaultValue.format(**self._cmdVars)
-                    attr._invalidationValue = defaultValue.format(**cmdVarsNoCache)
-                except KeyError as e:
-                    logging.warning('Invalid expression with missing key on "{nodeName}.{attrName}" with value "{defaultValue}".\nError: {err}'.format(nodeName=self.name, attrName=attr.name, defaultValue=defaultValue, err=str(e)))
-                except ValueError as e:
-                    logging.warning('Invalid expression value on "{nodeName}.{attrName}" with value "{defaultValue}".\nError: {err}'.format(nodeName=self.name, attrName=attr.name, defaultValue=defaultValue, err=str(e)))
+                # Do not evaluate expression for disabled attributes (the expression may refer to other attributes that are not yet defined)
```
fabiencastan, line 753:
```suggestion
                # Do not evaluate expression for disabled attributes (the expression may refer to other attributes that are not defined)
```

fabiencastan (approved):

PR #2431, opened 2024-06-10
[ui] Auto Update CameraInit when displaying node

Just-Kiel:
## Description
When displaying node not related with the current Camera Init, try to switch to the correct CameraInit to have everything working.

### DIFF
--- a/meshroom/ui/reconstruction.py
+++ b/meshroom/ui/reconstruction.py
@@ -999,12 +999,25 @@ class Reconstruction(UIGraph):
         for category, nodeTypes in self.activeNodeCategories.items():
             if node.nodeType in nodeTypes:
                 self.activeNodes.get(category).node = node
+
                 if category == 'sfm':
                     self.setSfm(node)
 
                 # if the active node is a CameraInit node, update the camera init index
                 if node.nodeType == "CameraInit":
                     self.setCameraInitIndex(self._cameraInits.indexOf(node))
+
+        # get all nodes that are input nodes of the active node
+        inputNodes = node.getInputNodes(recursive=True, dependenciesOnly=True)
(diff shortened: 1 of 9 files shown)


meshroom/ui/reconstruction.py
```diff
@@ -998,13 +998,25 @@ def setActiveNode(self, node):
...
             if node.nodeType in nodeTypes:
+                # get all nodes that are input nodes of the active node
+                inputNodes = node.getInputNodes(recursive=True, dependenciesOnly=True)
+                # get the first CameraInit node in the input nodes
+                cameraInitNode = next((n for n in inputNodes if n.nodeType == "CameraInit"), None)
+                # if the current camera init node is not the same as the camera init node of the active node
+                if cameraInitNode and cameraInitNode != self.cameraInit:
+                    # set the camera init node of the active node as the current camera init node
+                    self.setCameraInitIndex(self._cameraInits.indexOf(cameraInitNode))
+                    self.setSelectedViewId(self.viewpoints[0].viewId.value)
```
fabiencastan, line 1009: Not tested
```suggestion
                inputCameraInitNodes = [n for n in inputNodes if n.nodeType == "CameraInit")]
                # if the current camera init node is not the same as the camera init node of the active node
                if inputCameraInitNodes and self.cameraInit not in inputCameraInitNodes:
                    # set the camera init node of the active node as the current camera init node
                    # if multiple CameraInit, select one arbitrarily (the one with more viewpoints)
                    inputCameraInitNodes.sort(key=lambda n: len(n.viewpoints.value))
                    cameraInitNode = inputCameraInitNodes[-1]  # keep the last/largest one
                    self.setCameraInitIndex(self._cameraInits.indexOf(cameraInitNode))
                    self.setSelectedViewId(self.viewpoints[0].viewId.value)
```

fabiencastan (approved):

PR #2435, opened 2024-06-13
Prevent updates of the latest SfM node when the graph's topology is dirty

Just-Kiel:
## Description

When graph is in dirty topology, we do not want to update the last SfM node in the graph.


fabiencastan:
Upgrade CameraInit fixed in https://github.com/alicevision/Meshroom/pull/2436/commits/54f84056c30bbe0294dd329509df14baa99bd8b1

fabiencastan (changes requested):
The problem does not come from the CameraInit itself, see https://github.com/alicevision/Meshroom/pull/2436/commits/54f84056c30bbe0294dd329509df14baa99bd8b1

cbentejac (approved):

PR #2439, opened 2024-06-16
[ui] do not use native dialog

fabiencastan:
Avoid display troubles with native dialogs.


cbentejac:
Tested on both Linux and Windows, works as expected.

PR #2443, opened 2024-06-19
Provide access to the current frame from the graph

cbentejac:
## Description

This PR adds a `viewpoint` property to the `Graph` object, containing the viewpoint attribute corresponding to the image that is currently selected in the Image Gallery.

This property is updated every time the selected viewpoint changes, and can thus be accessed from any object that has access to the `Graph` object itself (e.g. from inside a node, with `node.graph.viewpoint`).

### DIFF
--- a/meshroom/core/graph.py
+++ b/meshroom/core/graph.py
@@ -226,6 +226,8 @@ class Graph(BaseObject):
         self._fileDateVersion = 0
         self.header = {}
 
+        self._selectedViewpoint = None
+
     def clear(self):
         self.header.clear()
         self._compatibilityNodes.clear()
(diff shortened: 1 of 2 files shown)


meshroom/core/graph.py
```diff
@@ -1564,6 +1566,10 @@ def cacheDir(self, value):
         self.updateStatusFromCache(force=True)
         self.cacheDirChanged.emit()
 
+    @property
+    def viewpoint(self):
```
fabiencastan, line 1570: selectedViewpoint/activeViewpoint/currentViewpoint?

fabiencastan: a docstring to explain too

fabiencastan (approved):

PR #2452, opened 2024-06-28
First Homepage

Just-Kiel:
## Description
First implementation of Homepage. The aim is to have something at the opening of Meshroom.

- [x] Sponsors section with redirection to websites
- [x] Grid view for recent projects
    - [x] Thumbnail for each project if possible
    - [x] Filestem of project
    - [x] Tooltip with the path of the project

- [x] Links to start section


cbentejac (changes requested):

cbentejac (changes requested):

cbentejac (approved):

meshroom/ui/qml/Homepage.qml
```diff
@@ -286,6 +286,33 @@ Page {
                         width: gridView.cellWidth
                         height: gridView.cellHeight
 
+                        property var source: modelData["thumbnail"] ? Filepath.stringToUrl(modelData["thumbnail"]) : ""
+
+                        // Update grid item when corresponding thumbnail is computed
+                        Connections {
```
fabiencastan, line 292: The "Connections" may be moved at the gridView level instead of each delegate.

PR #2456, opened 2024-07-03
Add a Python ScriptEditor in the GraphEditor tab

cbentejac:
## Description

This PR adds a new QML component to the GraphEditor tab which allows to easily write and execute Python code within Meshroom. Python scripts can be exported from or imported in the script editor, and the history of executed scripts is saved.

The ScriptEditor's input zone is set by default with a minimal script that lists the nodes that are part of the current graph, thus showing an example on how to access Meshroom elements in Python.

## Features list

- [x] Save and load Python scripts
- [x] Execute Python scripts
- [x] Navigate through the history of executed Python scripts during the Meshroom session

### DIFF
--- a/meshroom/__init__.py
+++ b/meshroom/__init__.py
@@ -150,4 +150,5 @@ def setupEnvironment(backend=Backend.STANDALONE):
 
 
 os.environ["QML_XHR_ALLOW_FILE_READ"] = '1'
+os.environ["QML_XHR_ALLOW_FILE_WRITE"] = '1'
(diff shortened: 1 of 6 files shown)


Just-Kiel:
As feature improvements :
- [ ] "Input" and "Output" labels are not centered
- [ ] If there is error maybe display it in red in the Output
- [ ] If there is error do not delete the script from the Input
- [ ] Have a stack of all Previous Scripts to choose which one you want to access

Just-Kiel (approved):

PR #2487, opened 2024-08-01
Initialize `core` plugins at different moments

cbentejac:
## Description

This PR changes the order of initialization of the `core` plugins both in the main app and `meshroom_batch`.

- For the main app, the pipelines are now initialized prior to the parsing of the command line to ensure that the names of all the available templates can be provided to the "help" message. The nodes and submitters remain initialized at the same point.
- For `meshroom_batch`, the pipelines are also initialized prior to the parsing of the command line (for the same reasons), the nodes are initialized at the same point, and the submitters are initialized if and only if the `--submit` argument has been provided.

### DIFF
--- a/bin/meshroom_batch
+++ b/bin/meshroom_batch
@@ -14,7 +14,7 @@ from meshroom import multiview
 from meshroom.core.desc import InitNode
 import logging
 
-meshroom.core.initPlugins()
+meshroom.core.initPipelines()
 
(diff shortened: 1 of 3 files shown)


Just-Kiel:
Maybe check that all commands within bin folder are correctly working, because nodes or submitters doesn't seem to be initialised.

fabiencastan (approved):

PR #2489, opened 2024-08-01
[core] Prevent from having undefined attribute and connect onChanged after first init

Just-Kiel:
## Description
To remove all errors raised in Image Gallery, we need to first set the value of the attribute before connecting it to onChanged. Plus, we remove the self._value = None to be sure that the UI never fails to display something.


cbentejac:
Covered by https://github.com/alicevision/Meshroom/pull/2586.

PR #2493, opened 2024-08-07
[core] Reference the attribute's instance type in its description

cbentejac:
## Description

This PR adds a new property called `instanceType` in the attribute description classes that contains the class of the corresponding attribute object.

This property is then used to instantiate directly attribute objects, rather than performing tests on the type of the description class to determine which constructor to use.

This solves an issue where any description that would inherit from an existing one would lead to instantiating an attribute of the parent class, e.g. having a description class inheriting `ChoiceParam` would always lead to a `ChoiceParam` attribute being created, even if a dedicated attribute class corresponding to that description existed. 

To set `instanceType` in a description class, the `getInstanceType` method can be overriden. By default, `instanceType` will always be set to `Attribute`.

Below is an example of custom description and attribute classes that inherit `ChoiceParam`, and a node using this new attribute:
```python

from meshroom.common import Property, Variant
from meshroom.core import attribute, desc

class TestAttribute(attribute.ChoiceParam):
    # custom implementation
    pass

class TestDescription(desc.ChoiceParam):
    def getInstanceType(self):
        return TestAttribute

    instanceType = Property(Variant, getInstanceType, constant=True)
```

```python

__version__ = "1.0"

from meshroom.core import desc, custom

class TestAttributeType(desc.InputNode):
    documentation = """ """
    category = 'Test'

    # Inputs to the node
    inputs = [
        custom.TestDescription(
            name="testAttribute",
            label="Test Attribute",
            description="Custom attribute inheriting ChoiceParam.",
            value="a",
            values=["a", "b", "c", "d"],
            exclusive=True,
            uid=[0],
        )
    ]
```


Just-Kiel (approved):

meshroom/core/desc.py
```diff
@@ -130,6 +138,11 @@ def matchDescription(self, value, strict=True):
             return self._elementDesc.matchDescription(value[0], strict)
         return True
 
+    elementDesc = Property(Attribute, lambda self: self._elementDesc, constant=True)
+    uid = Property(Variant, lambda self: self.elementDesc.uid, constant=True)
+    joinChar = Property(str, lambda self: self._joinChar, constant=True)
+    instanceType = Property(Variant, getInstanceType, constant=True)
```
fabiencastan, line 144:
```suggestion
    instanceType = Property(Variant, lambda self: self.getInstanceType(), constant=True)
```
And then we can remove it from all subclasses.

meshroom/core/desc.py
```diff
@@ -99,9 +107,9 @@ def __init__(self, elementDesc, name, label, description, group='allParams', adv
         self._joinChar = joinChar
         super(ListAttribute, self).__init__(name=name, label=label, description=description, value=[], uid=(), group=group, advanced=advanced, semantic=semantic, enabled=enabled, visible=visible)
 
-    elementDesc = Property(Attribute, lambda self: self._elementDesc, constant=True)
-    uid = Property(Variant, lambda self: self.elementDesc.uid, constant=True)
-    joinChar = Property(str, lambda self: self._joinChar, constant=True)
+    def getInstanceType(self):
+        from meshroom.core.attribute import ListAttribute
```
fabiencastan, line 111:
```suggestion
        # Import within the method to prevent cyclic import
        from meshroom.core.attribute import ListAttribute
```

meshroom/core/desc.py
cbentejac: Done.

meshroom/core/desc.py
cbentejac: Done.

fabiencastan (approved):

PR #2500, opened 2024-08-14
[core] Catch exception for calls to optional descriptor method on node creation

cbentejac:
## Description

If a node's description contains invalid operations in its `onNodeCreated` method, exceptions are thrown without being caught and cause Meshroom to crash.

This PR fixes this issue by safely trying optional calls to the node's descriptor, and handling any encountered exception.

### DIFF
--- a/meshroom/core/node.py
+++ b/meshroom/core/node.py
@@ -1417,12 +1417,30 @@ class Node(BaseNode):
 
         self.optionalCallOnDescriptor("onNodeCreated")
 
+
     def optionalCallOnDescriptor(self, methodName, *args, **kwargs):
-        """ Call of optional method defined in the descriptor. By now there is the onNodeCreated existing. """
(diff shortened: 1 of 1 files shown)


meshroom/core/node.py
```diff
@@ -1415,10 +1415,19 @@ def __init__(self, nodeType, position=None, parent=None, uids=None, **kwargs):
             for uidIndex in attr.attributeDesc.uid:
                 self.attributesPerUid[uidIndex].add(attr)
 
-        self.optionalCallOnDescriptor("onNodeCreated")
+        try:
+            self.optionalCallOnDescriptor("onNodeCreated")
```
fabiencastan, line 1419: Maybe better to directly try/except in optionalCallOnDescriptor.

fabiencastan (approved):

PR #2503, opened 2024-08-16
[Controls] Prevent the filter text field from being out of the window

cbentejac:
## Description

This PR fixes a UI issue that occurred for exclusive `ChoiceParam` attributes with large lists: when the node editor's height is bigger than the default one, the popup with the list of values is positioned at the top of the attribute's field, and its filtering text field was positioned outside of the application window.

If the top-left corner of the popup is to be positioned outside of the application window (or too close to its border), it is now automatically replaced at a fixed distance from the top of the window.


cbentejac:
https://github.com/alicevision/Meshroom/pull/2656 replaces the use for this PR.

PR #2506, opened 2024-08-20
Fix Several Compatibility Nodes Operations

Just-Kiel:
## Description
- Recompute and Re-Submit not allowed for Compatibility Nodes because they are not computable nodes

### DIFF
--- a/meshroom/ui/qml/GraphEditor/GraphEditor.qml
+++ b/meshroom/ui/qml/GraphEditor/GraphEditor.qml
@@ -555,7 +555,7 @@ Item {
             Menu {
                 id: nodeMenu
                 property var currentNode: null
-                property bool canComputeNode: currentNode != null && uigraph.graph.canCompute(currentNode)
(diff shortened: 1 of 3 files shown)


meshroom/ui/qml/GraphEditor/GraphEditor.qml
```diff
@@ -565,14 +577,30 @@ Item {
...
-                    enabled: true
+                    enabled: {
+                        var canPaste = false
+                        for (var i = 0; i < uigraph.selectedNodes.count; ++i) {
+                            if (!uigraph.selectedNodes.at(i).isCompatibilityNode) {
+                                canPaste = true
+                            }
+                        }
+                        return canPaste
+                    }
```
fabiencastan, line 603: Paste is independent from the selection.

fabiencastan (changes requested):

fabiencastan (approved):

fabiencastan (approved):

PR #2519, opened 2024-09-02
[ui] Fix function evaluations in invalid QML context and minor fixes

cbentejac:
## Description

This PR fixes the following QML warning:
`QQmlVMEMetaObject: Internal error - attempted to evaluate a function in an invalid context`
which was introduced in #1925 and occurred when edges were being destroyed while the QML context was invalid (meaning the application had entered its closing phase).

The status of the application's window is monitored to detect when it's entering its closing phase (which will lead to the QML context being invalidated). If the window is closing, during the destruction of edges components, no pin operation to re-enable pins will be performed. 

Additionally, this PR also fixes minor QML warnings:
- calls to `hasAttribute` when `currentNode` is `null`
- assignments of `undefined` to edge-related properties when the graph is being updated

## Features list

- [x] Add a property that monitors when the QML context might become invalid;
- [x] Do not call functions while components are being destroyed if the QML context has become invalid;
- [x] Do not call functions on non-existing objects
- [x] Prevent assigning `undefined` to typed properties

### DIFF
--- a/meshroom/ui/qml/GraphEditor/GraphEditor.qml
+++ b/meshroom/ui/qml/GraphEditor/GraphEditor.qml
@@ -542,11 +542,15 @@ Item {
 
                     Component.onDestruction: {
                         // Handles the case where the edge is destroyed while hidden because it is replaced: the pins should be re-enabled
-                        if (src && src !== undefined)
-                            src.updatePin(true, true)  // isSrc = true, isVisible = true
-                        if (dst && dst !== undefined)
-                            dst.updatePin(false, true)  // isSrc = false, isVisible = true
+                        if (!_window.isClosing) {
+                            // When the window is closing, the QML context becomes invalid, which causes function calls to result in errors
+                            if (src && src !== undefined)
+                                src.updatePin(true, true)  // isSrc = true, isVisible = true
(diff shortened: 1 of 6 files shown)


Just-Kiel (approved):

meshroom/ui/qml/GraphEditor/GraphEditor.qml
```diff
@@ -443,11 +443,15 @@ Item {
 
                     Component.onDestruction: {
                         // Handles the case where the edge is destroyed while hidden because it is replaced: the pins should be re-enabled
-                        if (src && src !== undefined)
-                            src.updatePin(true, true)  // isSrc = true, isVisible = true
-                        if (dst && dst !== undefined)
-                            dst.updatePin(false, true)  // isSrc = false, isVisible = true
+                        if (!_window.closing) {
+                            // If the window is closing, the QML context is rendered invalid, so function calls will generate errors
```
fabiencastan, line 447:
```suggestion
                            // When the window is closing, the QML context becomes invalid, which causes function calls to result in errors.
```

cbentejac: Done

PR #2521, opened 2024-09-03
[ui] Add proper distinction between the main window and the application

cbentejac:
## Description

This PR finishes to separate the content of the application (in opposition to the content of the homepage) from the content of the main window. 
- All the actions, file dialogs and message dialogs that are specific to the application are removed from `main.qml` and placed instead in `Application.qml`.
- Only elements and functions related to the main window remain in `main.qml`. As such, the `ensureSaved` and `ensureNotComputing` functions, used when closing the main window, remain in `main.qml`.
- Message dialogs that may be opened by these functions however are moved to `Application.qml` as they are application-specific and can't be opened on the homepage.
- Application dialogs that may need to be called when triggering the closure of the main window are aliased from `Application.qml` to remain callable through the `StackView`.

This also fixes an issue that stemed from the initial separation between `main` and `Application`, where some pop-ups where opened from `main` while calling actions from `Application` that they had lost all reference to.

The use of the `DontUseNativeDialog` option for `FileDialogs` has been fixed to make sure all the file dialogs are always correctly displayed.

## Features list

- [x] Fully separate contents of `main` and `Application`;
- [x] Clean-up QML imports in `main.qml`;
- [x] Use correct `DontUseNativeDialog` option across all file dialogs;
- [x] Rename general and UI settings components to respect `camelCase`.

### DIFF
--- a/meshroom/ui/qml/Application.qml
+++ b/meshroom/ui/qml/Application.qml
@@ -40,9 +40,10 @@ Page {
         if (_reconstruction.graph && _reconstruction.graph.filepath) {
             folder = Filepath.stringToUrl(Filepath.dirname(_reconstruction.graph.filepath))
         } else {
-            var projects = MeshroomApp.recentProjectFiles;
-            if (projects.length > 0 && Filepath.exists(projects[0])) {
-                folder = Filepath.stringToUrl(Filepath.dirname(projects[0]))
+            var projects = MeshroomApp.recentProjectFiles
+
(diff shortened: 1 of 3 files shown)


servantftechnicolor (approved):
Remove the detected bugs

Just-Kiel (changes requested):
When going back to the homepage from the application if the project is not saved, the popup immediately closes without waiting.

Just-Kiel (approved):

PR #2527, opened 2024-09-10
[AttributePin] Add tooltip to display type of attribute

Just-Kiel:
## Description
If we need to know the type of attribute shortly we now have a tooltip displaying it.


meshroom/ui/qml/GraphEditor/AttributeItemDelegate.qml
```diff
@@ -33,7 +33,7 @@ RowLayout {
             var tooltip = ""
             if (!attribute.validValue && attribute.desc.errorMessage !== "")
                 tooltip += "<i><b>Error: </b>" + Format.plainToHtml(attribute.desc.errorMessage) + "</i><br><br>"
-            tooltip += "<b> " + attribute.desc.name + "</b><br>" + Format.plainToHtml(attribute.desc.description)
+            tooltip += "<b> " + attribute.desc.name + ": " + attribute.type + "</b><br>" + Format.plainToHtml(attribute.desc.description)
```
fabiencastan, line 36: maybe the attribute type does not need to be bold.

fabiencastan (approved):

PR #2528, opened 2024-09-10
[core/ui] "Exposed" property added to attributeDesc

Just-Kiel:
## Description
`exposed` is useful to display the attribute in the upper part of the node in the Graph Editor. Prior to this PR, all outputs attributes were displayed in that part, but only input `File` attributes were as well. Any other input attributes was displayed in the lower part of the node. 

`exposed` is thus useless for output attributes, and is only taken into account for input attributes. By default, only `File` attributes are exposed while all the other attributes are not, in order to preserve the default behaviour we had until now.


fabiencastan:
Need to expose "Features/Matches Folders" in AliceVision nodes.
It would be good to also expose "Describer Types".

PR #2539, opened 2024-09-17
Added Pre and Post process functions on the Base Node

fabiencastan:
Added Pre and Post process functions on the Node Description which can be utilised by the nodes.

### DIFF
--- a/meshroom/core/node.py
+++ b/meshroom/core/node.py
@@ -1030,13 +1030,16 @@ class BaseNode(BaseObject):
         self._chunks[iteration].process()
 
     def preprocess(self):
-        pass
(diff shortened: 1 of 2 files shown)


meshroom/core/node.py
```diff
@@ -1030,13 +1030,16 @@ def processIteration(self, iteration):
         self._chunks[iteration].process()
 
     def preprocess(self):
-        pass
+        # Invoke the Node Description's pre-process for the Client Node to prep it's processing
```
fabiencastan, line 1033:
```suggestion
        # Invoke the Node Description's pre-process for the Client Node to prep its processing
```

PR #2544, opened 2024-09-20
Add display of attributes that are part of `GroupAttributes`

cbentejac:
## Description

This PR is based on #2538 and must be merged once #2538 itself has been integrated into `develop`.


cbentejac (commented):
Need to check if the generation of the attributes within a group in the node cannot be performed in the same way as in the node editor.

nicolas-lambert-tc:
Screenshot_2025-06-17_09-02-56

I think it would be easier to read if the GroupAttribute that have at least one sub attribute connected to be visible

PR #2566, opened 2024-10-09
[qt6]|Viewer3D] Fix mouse for camera controller

servantftechnicolor:
For whatever reason, the logicaldevice does not work on qt 6.7 with the mouse. 
This PR will avoid the problem by using another approach


fabiencastan (changes requested):
Based on commits from another branch.

cbentejac (approved):

PR #2579, opened 2024-10-24
[nodes] `ExportDistortion`: Add animated distortion option

servantftechnicolor:
In ExportDistortion, add an option to export a nuke file with per frame information about distortion


cbentejac (changes requested):
The version of the node needs to be updated as the list of parameters has changed.

cbentejac (approved):

PR #2581, opened 2024-10-24
[ui] Improve Search Bar component

waaake:
### Search bar Updates
* Search bar can now be toggled with the Search Button.
* The GraphEditor had an obstructive search before and now it is clubbed with the bottom panel options.
* The Search Bar now supports "Return Key Press" to move on the next filtered item in the Graph Editor.
* Search Bar also gets a clear text action when text is entered in the field.
* Updated SearchBar in ImageGallery, Node Editor, 3D Inspector to now be toggled.

### DIFF
--- a/meshroom/ui/qml/Controls/SearchBar.qml
+++ b/meshroom/ui/qml/Controls/SearchBar.qml
@@ -9,8 +9,23 @@ import MaterialIcons 2.2
  */
 
 FocusScope {
+    id: root
     property alias textField: field
     property alias text: field.text
+
+    // Enables hiding and showing of the text field on Search button click
+    property bool toggle: false
+    property bool isVisible: false
+
+    // Size properties
+    property int maxWidth: 150
+    property int minWidth: 30
+
+    // The default width is computed based on whether toggling is enabled and if the visibility is true
+    width: toggle && isVisible ? maxWidth : minWidth
+
+    // Keyboard interaction related signals
+    signal accepted()
 
     implicitHeight: childrenRect.height
     Keys.forwardTo: [field]
@@ -24,10 +39,17 @@ FocusScope {
     }
 
     RowLayout {
+        spacing: 0
         width: parent.width
 
-        MaterialLabel {
+        MaterialToolButton {
             text: MaterialIcons.search
+
+            onClicked: {
+                isVisible = !root.isVisible
+                // Set Focus on the Text Field
+                field.focus = field.visible
+            }
         }
 
         TextField {
@@ -36,9 +58,45 @@ FocusScope {
             Layout.fillWidth: true
             selectByMouse: true
 
+            rightPadding: clear.width
+
+            // The text field is visible either when toggle is not activated or the visible property is set
+            visible: root.toggle ? root.isVisible : true
+
             // Ensure the field has focus when the text is modified
             onTextChanged: {
                 forceActiveFocus()
+            }
+
+            // Handle enter Key press and forward it to the parent
+            Keys.onPressed: (event)=> {
+                if ((event.key == Qt.Key_Return || event.key == Qt.Key_Enter)) {
+                    event.accepted = true
+                    root.accepted()
+                }
+            }
+
+            MaterialToolButton {
+                id: clear
+
+                // Anchors
+                anchors.right: parent.right
+                anchors.rightMargin: 2  // Leave a tiny bit of space so that its highlight does not overlap with the boundary of the parent
+                anchors.verticalCenter: parent.verticalCenter
+
+                // Style
+                font.pointSize: 8
+                text: MaterialIcons.close
+                ToolTip.text: "Clears text."
+
+                // States
+                visible: field.text
+
+                // Signals -> Slots
+                onClicked: {
+                    field.text = ""
+                    parent.focus = true
+                }
             }
         }
     }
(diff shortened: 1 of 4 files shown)


meshroom/ui/qml/Viewer3D/Inspector3D.qml
```diff
@@ -298,10 +298,11 @@ FloatingPane {
 
                 SearchBar {
                     id: searchBar
-                    Layout.minimumWidth: 150
-                    Layout.fillWidth: true
-                    Layout.rightMargin: 10
-                    Layout.leftMargin: 10
+
+                    toggle: true // enable toggling the actual text field by the search button
```
cbentejac, line 302:
```suggestion
                    toggle: true  // enable toggling the actual text field by the search button
                    isVisible: true
```

I think the search bar should already be expanded in the case of the 3D inspector. Otherwise, you just have the magnifier icon on the top-left that blocks the entire row while the loaded elements are displayed on the entire row below, and it creates a major imbalance visually speaking.

meshroom/ui/qml/Controls/SearchBar.qml
```diff
@@ -35,10 +58,45 @@ FocusScope {
...
+                    root.returnPressed()
+                }
+            }
+
+            MaterialToolButton {
+                id: clear
+
+                // Anchors
+                anchors.right: parent.right
+                anchors.verticalCenter: parent.verticalCenter
```
cbentejac, line 84: I think some minor right-padding should be added: when the "clear" button appears, the search bar generally has the focus (so is "highlighted") and when hovering over the button, it goes over the highlighting, which is not very pleasing.
Screenshot_2024-11-12_10-28-03

meshroom/ui/qml/Viewer3D/Inspector3D.qml
cbentejac: I also think it'd be better to have some left- and right-margins in this case: the search bar takes up 100% of the horizontal space, which is not the case of the loaded 3D elements, and it also creates an imbalance.

Screenshot_2024-11-12_10-33-26

cbentejac (changes requested):

meshroom/ui/qml/Viewer3D/Inspector3D.qml
```diff
@@ -298,10 +298,11 @@ FloatingPane {
...
+                    Layout.minimumWidth: searchBar.width
+                    maxWidth: parent.width
```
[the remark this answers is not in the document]
waaake: Inspector 3d gets the default search bar, so this warning doesn't appear anymore.

meshroom/ui/qml/Controls/SearchBar.qml
waaake: Yes, this is now fixed with the change here, https://github.com/alicevision/Meshroom/blob/1a0d2cf905903091012db3384c985a1d198490c8/meshroom/ui/qml/Controls/SearchBar.qml#L83
This change ensures we're leaving sufficient space on the right side as to not overlap with the parent's bounds.
Hoping this makes it better 🙂

cbentejac (approved):

yann-lty (changes requested):
Works well overall (only minor changes required)!
Side note, the handling of bound iteration is acting weird when we don't have a match in the Graph Editor (but outside the scope of you changes). 
image

meshroom/ui/qml/GraphEditor/GraphEditor.qml
```diff
@@ -1101,32 +1101,34 @@ Item {
...
-            model: SortFilterDelegateModel {
-                model: root.graph ? root.graph.nodes : undefined
-                sortRole: "label"
-                filters: [{role: "label", value: graphSearchBar.text}]
-                delegate: Item {
-                    property var index_: index
-                }
-                function modelData(item, roleName_) {
-                    return item.model.object[roleName_]
+            Item {
```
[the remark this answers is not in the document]
waaake: I started facing the issue with Qt6 rebase where the width of the panel reacts to the model contents in the repeater, as the filteredNodes repeater has `root.graph.nodes` as the model, a new node getting add to the graph was increasing the width of the panel and same when the node gets removed, the width would get decreased. Prior to Qt6 this was working as expected without need for the additional Item which is hidden. 

https://github.com/user-attachments/assets/dc619e1a-f0eb-4c37-90c9-73ea576ae6bb

meshroom/ui/qml/Controls/SearchBar.qml
```diff
@@ -36,10 +58,46 @@ FocusScope {
...
+            // The text field is visible either when toggle is not activated or the visible property is set
+            visible: root.toggle ? root.isVisible : true
+
             // Ensure the field has focus when the text is modified
             onTextChanged: {
                 forceActiveFocus()
             }
+
+            // Handle enter Key press and forward it to the parent
+            Keys.onPressed: (event)=> {
```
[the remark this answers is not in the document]
waaake: The core signal from the class is updated to be called `accepted` but using `onAccepted` instead of `Keys.onPressed` currently is emitting the signal twice so keeping it same at the moment, 
Have updated the event to consider both Return and Number Pad Enter key though.

waaake:
> Works well overall (only minor changes required)! Side note, the handling of bound iteration is acting weird when we don't have a match in the Graph Editor (but outside the scope of you changes). image

Thanks for this feedback, I had noticed it earlier, but left it as is.
Now with the changes to have a separate function `navigateForwards` and `navigateBackwards`, have added a condition to ensure we have something filtered to be able to change the index of the search. That resolves this issue 🙂

PR #2587, opened 2024-10-28
[ui] Python Script Editor Improvements

waaake:
## Features list
* Script Editor now retains history of the code which is executed across Meshroom sessions.
* Script Editor also highlights the code Syntax for Python
* The output clearly distinguishes between executed code and the result of the code with different colors.

### DIFF
--- a/meshroom/ui/qml/GraphEditor/ScriptEditor.qml
+++ b/meshroom/ui/qml/GraphEditor/ScriptEditor.qml
@@ -9,6 +9,8 @@ import Utils 1.0
 
 import Qt.labs.platform 1.0 as Platform
 
+import ScriptEditor 1.0
+
 Item {
     id: root
 
@@ -338,6 +340,13 @@ Item {
                         lineNumbers.contentY = contentY
                     }
                 }
+            }
+
+            // Syntax Highlights for the Input Area for Python Based Syntax
+            PySyntaxHighlighter {
+                id: syntaxHighlighter
+                // The document to highlight
+                textDocument: input.textDocument
             }
         }
     }
--- a/meshroom/ui/qml/GraphEditor/ScriptEditor.qml
+++ b/meshroom/ui/qml/GraphEditor/ScriptEditor.qml
@@ -211,50 +211,16 @@ Item {
             }
         }
 
-        MSplitView {
-            id: topBottomSplit
+        RowLayout {
             Layout.fillHeight: true
             Layout.fillWidth: true
-
-            orientation: Qt.Vertical
-        
-            // Output Text Area -- Shows the output for the executed script(s)
-            Rectangle {
-                id: outputArea
-
-                // Has a minimum height
-                SplitView.minimumHeight: 80
-
-                color: palette.base
-
-                Flickable {
-                    width: parent.width
-                    height: parent.height
-                    contentWidth: width
-                    contentHeight: ( output.lineCount + 5 ) * output.font.pixelSize // + 5 lines for buffer to be scrolled and visibility
-
-                    ScrollBar.vertical: MScrollBar {}
-
-                    TextArea.flickable: TextArea {
-                        id: output
-
-                        readOnly: true
-                        selectByMouse: true
-                        padding: 0
-                        Layout.fillHeight: true
-                        Layout.fillWidth: true
-                        wrapMode: Text.WordWrap
-
-                        textFormat: Text.RichText
-                    }
-                }
-            }
+            width: root.width
 
             // Input Text Area -- Holds the input scripts to be executed
             Rectangle {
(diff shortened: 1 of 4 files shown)


meshroom/ui/qml/GraphEditor/ScriptEditor.qml
```diff
@@ -182,31 +210,50 @@ Item {
...
         }
 
-        RowLayout {
-            Label {
-                text: "Input"
-                font.bold: true
-                horizontalAlignment: Text.AlignHCenter
-                Layout.fillWidth: true
-            }
+        MSplitView {
```
cbentejac, line 213: This needs to be discussed but I wonder if it wouldn't be better to have the split view be horizontal instead of vertical... As such, it's is pretty much unusable unless we expand the Script Editor's tab, which may not be ideal.

waaake: Thanks for your feedback @cbentejac, I do agree with how the widgets are placed in Meshroom at the moment, the best layout is a RowLayout. 
Have reverted back in terms of Layout and I guess we can keep it till the time we have a better location/placement of Script Editor.

waaake:
Thanks for the suggestions @yann-lty 
I have added a confirmation dialog for clearing the history, let me know how these new icons feel.
MeshroomScriptEditorConfirmation

meshroom/ui/qml/GraphEditor/ScriptEditor.qml
```diff
@@ -9,22 +9,81 @@ import Utils 1.0
...
+        return "<font color=#868686>" + replace(text, "\n", "<br>") + "</font><br><br>"
+    }
+
+    function formatOutput(text) {
+        /**
+         * Formats the text to be displayed as the result of the script executed
+         */
+
+        // Replace the text to be RichText Supportive
+        return "<font color=#49a1f3>" + "Result: " + replace(text, "\n", "<br>") + "</font><br><br>"
```
fabiencastan, line 63:
```suggestion
        return "<font color=#49a1f3>" + "> Result:\n" + replace(text, "\n", "<br>") + "</font><br>"
```

meshroom/ui/qml/GraphEditor/ScriptEditor.qml
```diff
@@ -9,22 +9,81 @@ import Utils 1.0
...
+        return lines.join(replacement)
+    }
+
+    function formatInput(text) {
+        /**
+         * Formats the text to be displayed as the input script executed
+         */
+
+        // Replace the text to be RichText Supportive
+        return "<font color=#868686>" + replace(text, "\n", "<br>") + "</font><br><br>"
```
fabiencastan, line 54:
```suggestion
        return "<font color=#868686>" + replace(text, "\n", "<br>") + "</font><br>"
```

fabiencastan (commented):
We cannot scroll to see the end of the text:
image

fabiencastan:
font size for icons is too big:
image

fabiencastan:
I would propose to update the order of the icons:
image
Groups all the actions related to the script, center the execution, and align to the right the clear.

Or:
image

fabiencastan:
Could it be a SplitView between the Editor and the Output.

fabiencastan:
Maybe also disable the button if there is no action to perform on the "Get Previous/Next Script".

waaake:
> The first click on "Get Previous Script" is not working. You need to click a second time to do the action.

It feels like that actually, but it is actually working, 
If I executed the below example:
`print("A")`

Then I executed 
`print("B")`
Clicked on Get Previous Script, 
it gets back `print("B")`

but I would not notice any difference because my current text in script editor input is still `print("B")` so I have to click on it two times to get it back to `print("A")`

**I think in the first version of script editor we used to clear the input as soon as it was executed so this made it work as expected but now that we're not clearing the script contents, it feels like this is not working....** 
🙁

fabiencastan:
It would be good to scroll the output to the end on execution.

fabiencastan:
It looks surprising to me that clearing the history is not only clearing the history but also the current code in the script editor.

meshroom/ui/qml/GraphEditor/ScriptEditor.qml
```diff
@@ -168,46 +198,57 @@ Item {
...
+                }
+            }
         }
 
-        RowLayout {
-            Layout.fillWidth: true
-            Layout.fillHeight: true
-            width: root.width
+        MSplitView {
+            id: scriptSplitView;
```
fabiencastan, line 243: I think we do not put the extra ";" in general.

fabiencastan (approved):

PR #2602, opened 2024-11-22
[BugFix] File save dialog now requires a valid filename

waaake:
Validating the filename to ensure that the file does not gets saved with just the extension

## Description
* Added a Validation to the file and template save process to ensure we have a valid filename to save to and not just the extension in the saved filename.

## Implementation remarks
* The validation process involves showing up a dialog box prompting user to provide a correct filename to save. This dialog box is recurring till the time a valid filename is entered for saving the file.

### DIFF
--- a/meshroom/ui/qml/Application.qml
+++ b/meshroom/ui/qml/Application.qml
@@ -59,6 +59,47 @@ Page {
         return ""
     }
 
+    Component {
+        id: invalidFilepathDialog
+
+        MessageDialog {
+            title: "Invalid Filepath"
+
+            required property string filepath
+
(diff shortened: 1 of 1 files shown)


waaake:
Thanks @yann-lty for the wonderful suggestions on the PR, this is working as expected with function and reused implementation of MessageDialog as a warning. 

The only annoyance that I get is of this error at the moment,
`Created Graphical Object was not placed in the Graphics Scene`
With myself trying to pass a parent which comes from the main.qml (parent.parent) or (root.parent) instead of root still somehow causes this to be raised.

yann-lty (approved):

PR #2612, opened 2024-12-02
Refactor Graph de/serialization

yann-lty:
## Description
Refactoring graph de/serialization for code-base simplification, fixing issues and better testability.

This PR introduces a major refactoring of the graph serialization/deserialization logic (aka graph IO).
With those changes, we address some of the known issues in this area, most notably when copy/pasting nodes.

Overall, the main concept is that any operation that needs to operate on serialized Graph data uses deserialization first, to work with Graph API, instead of manipulating a raw dictionary of serialized data.
It also creates a clear distinction in the API between the loading of a Meshroom file, the initialization of a Graph from a template and importing the content of another Graph.

## Features
- [X] Fix copy pasting issues that was failling to re-connect nodes correctly on paste.
- [X] Deal with advanced copy/pasting scenarios, eg. handling compatibility nodes.
- [X] Improved graph loading performance.

## Implementation details
The goal of this PR is to move away from a monolothic Graph class that mixes a lot of responsibilities, to make the code more manageable and testable.

### NodeFactory
The nodeFactory function has been refactored to rely on an underlying class that manages the compatibility checks. While preserving the existing behavior, this class helps understanding the different steps that occurs during that that process.

#### Deserialization
The current implementation exposes a unique entry point, as a single `load` function with several booleans to manage the different cases,  to deal with:
- loading a Meshroom file
- initializing a Graph instance from a template file
- importing another graph's content

```mermaid
classDiagram
class Graph {
  load(filepath, setupProjectFile=True, importProject=False, publishOutputs=False)
}
```
This makes it complex to manage the possible combinations of options, as all of them are not necessarily meaningful.

In the new approach, those are decoupled into as many entry points as required to handle the distinct logic those operations have:

```mermaid
classDiagram

class Graph {
 + load(filepath: PathLike)
 + initFromTemplate(filepath: PathLike, publishOutputs: bool)
 + importGraphContentFromFile(filepath: PathLike)
 + importGraphContent(graph: Graph)
}
```

#### UID conflict check
The current implementation creates all nodes twice to deal with UID compatibility issues, which increases load times and may keep false-positives depending on the order in which nodes are handled.

In the new approach, we only consider nodes that actually have a UID conflict, and solve them by depth ordering.
Indeed, UID conflicts are contagious: an upstream node can impact the downstream ones, as UID flow through connections.
By replacing upstream nodes first with CompatibilityNodes, we restore their serialized UID, which may solve the "false-positive" conflicts downstream, that were only due to the contagion by this node.

#### Importing another graph content
This area benefits the most from the refactoring, which solves errors we may currently encounter in the copy-pasting of nodes.
Instead of manipulating raw dictionaries to solve conflicting node names w.r.t the current content of the Graph, it now relies on: 
- the Graph API to make changes to node names in a separate Graph instance
- node serialization, to get the correctly updated link expressions (serialized Edges)
- node deserialization, to import nodes in the target Graph instance

#### Serialization
Introducing new GraphSerializer classes, following a Template Method pattern.
This makes it easier to define several node serialization logic, while inheriting from the standard serialization structure.


meshroom/core/graphIO.py
```diff
@@ -54,3 +59,110 @@ def getFeaturesForVersion(fileVersion: Union[str, Version]) -> tuple["GraphIO.Fe
...
+        inputKeys = list(nodeData["inputs"].keys())
+
+        internalInputKeys = []
+        internalInputs = nodeData.get("internalInputs", None)
+        if internalInputs:
+            internalInputKeys = list(internalInputs.keys())
+
+        for attrName in inputKeys:
+            attribute = node.attribute(attrName)
+            # check that attribute is not a link for choice attributes
```
fabiencastan, line 149: I don't understand this comment. Same for internalAttributes. Why are we talking about Choice attributes here?

yann-lty: Good point - I've just moved that template serialization code within this class as-is, but it's true that I did not rework the internals (comments included).
I'll try to understand what it means, improve the comment and write a test around that.

meshroom/ui/qml/GraphEditor/GraphEditor.qml
```diff
@@ -82,25 +82,18 @@ Item {
 
     /// Paste content of clipboard to graph editor and create new node if valid
     function pasteNodes() {
-        var finalPosition = undefined
-        var centerPosition = false
+        let finalPosition = undefined;
```
fabiencastan, line 85: In many other places, we don't use ";".

meshroom/core/graph.py
```diff
@@ -663,25 +663,49 @@ def upgradeNode(self, nodeName):
...
+            self.addNode(newNode, nodeName)
+            self._restoreOutEdges(outEdges, outListAttributes)
+    
+    def _restoreOutEdges(self, outEdges: dict[str, str], outListAttributes):
+        """Restore output edges that were removed during a call to "removeNode".
+        
+        Args:
+            outEdges: a dictionary containing the outgoing edges removed by a call to "removeNode".
+                {dstAttr.getFullNameToNode(), srcAttr.getFullNameToNode()}
+            outListAttributes: a dictionary containing the values, indices and keys of attributes that were connected
```
fabiencastan, line 697: Not introduced by this PR, but it may be good to create a structured type instead of this dictionnary.

meshroom/core/graph.py
```diff
@@ -643,7 +643,7 @@ def _createUniqueNodeName(self, inputName: str, existingNames: Optional[set[str]
     def node(self, nodeName):
         return self._nodes.get(nodeName)
 
-    def upgradeNode(self, nodeName):
+    def upgradeNode(self, nodeName) -> Node:
```
fabiencastan, line 655: That's much better!

fabiencastan (approved):

cbentejac (approved):

PR #2613, opened 2024-12-03
[GraphEditor] Edge: Correctly update the `EdgeMouseArea` when moving nodes

cbentejac:
## Description

This pull request correctly updates the mouse area over edges when nodes these edges are connected to have been moved. 

Prior to it, the mouse area of each edge was set once and never updated afterwards, meaning the edge remained selectable if and only if it did not move too much from its original position. The curve scale of the mouse area is now recomputed every single time the size of its edge has changed (i.e. whenever a node it is connected to has been moved).

### DIFF
--- a/meshroom/ui/qml/GraphEditor/Edge.qml
+++ b/meshroom/ui/qml/GraphEditor/Edge.qml
@@ -148,8 +148,8 @@ Item {
         Component.onCompleted: {
             /* The curve scale must be set only once the component has been fully created, so
              * that all the events following the update of the curve scale can be taken into
-             * account */
-            curveScale = cubic.ctrlPtDist / root.width  // Normalize by width
+             * account. */
+            curveScale = Qt.binding(() => cubic.ctrlPtDist / root.width)  // Normalize by width
         }
     }
 }


meshroom/ui/qml/GraphEditor/Edge.qml
```diff
@@ -148,8 +148,21 @@ Item {
         Component.onCompleted: {
             /* The curve scale must be set only once the component has been fully created, so
              * that all the events following the update of the curve scale can be taken into
-             * account */
+             * account. */
             curveScale = cubic.ctrlPtDist / root.width  // Normalize by width
         }
+
+        Connections {
```
yann-lty, line 155: If I understand the problem well, Qt6 broke the declarative binding we used to have for the `curveScale`, because the order in which some events are triggered has changed.
If so, I would recommend keeping the logic in `Component.onCompleted`, but create that binding imperatively using `Qt.binding()`. Instead of having a first initialization there, and the updates handled in a Connections component.

Something like:
```
Component.onCompleted: {
    /* ... */
    curveScale = Qt.binding(() => cubic.ctrlPtDist / root.width)  // Normalize by width
}
```

cbentejac: Thank you for this feedback! All done :)

yann-lty (approved):

PR #2617, opened 2024-12-12
Export Maya .mel Script 

servantftechnicolor:
Export a maya mel script such that the scene will contains

- The mesh
- The cameras correctly setup'ed
- The background images


mh0g:
Screenshot 2025-02-06 114534
tested on photogrametry pipeline, works and can open the scene in maya

mh0g (approved):

PR #2624, opened 2024-12-23
[ui] NodeEditor: Addressed Tab Retention when switching Node selection

waaake:
## Description
This PR addresses an issue with the Node Editor, where the selected tab was not getting retained when switching between selected nodes.

## Features list
- [x] Fixed Node Editor Switching to ensure the current tab retains when it can, depending on the node type.
- [x] Fixed issue with StatViewer causing the polish-loop during updates.
~- Introduced the `currentProjectChanged` signal which is meant to trigger update on various components when the scene is changed~
~- Currently the `onCurrentProjectChanged` slot is linked to refreshing the NodeEditor tabs.~

## Implementation remarks
Updated the event trigger to be `onVisibleChanged` rather than `onIsComputableChanged` as the latter does not get triggered when there is no node selected and then an incomputable node gets selected but the value of isComputable remains the same.

Visibility acts as a better trigger in my view, getting triggered when a node gets selected or deselected.
The condition is also updated to check if the current tab is not valid for the incomputable node before resetting it to 0.

This ensures that when documentation or such tab is selected, we don't end up resetting it back to 0 as it can exist in both the node types.

The available indices can be updated as we progress further with different types of nodes in the future.

### DIFF
--- a/meshroom/ui/qml/Application.qml
+++ b/meshroom/ui/qml/Application.qml
@@ -20,6 +20,16 @@ Page {
     property alias unsavedDialog: unsavedDialog
     property alias workspaceView: workspaceView
 
+    readonly property var scenefile: _reconstruction ? _reconstruction.graph.filepath : "";
+
+    onScenefileChanged: {
+        // Check if we're not currently saving and emit the currentProjectChanged signal
+        if (! _reconstruction.graph.isSaving) {
+            // Refresh the NodeEditor
+            nodeEditor.refresh();
+        }
+    }
+
     Settings {
         id: settingsUILayout
         category: "UILayout"
--- a/meshroom/core/graph.py
+++ b/meshroom/core/graph.py
@@ -255,7 +255,7 @@ class Graph(BaseObject):
     
     @property
     def isSaving(self):
-        """ Return True if the graph is currently being loaded. """
+        """ Return True if the graph is currently being saved. """
         return self._saving
 
     @Slot(str)
--- a/meshroom/core/graph.py
+++ b/meshroom/core/graph.py
@@ -215,6 +215,7 @@ class Graph(BaseObject):
         super(Graph, self).__init__(parent)
         self.name = name
         self._loading = False
+        self._saving = False
         self._updateEnabled = True
         self._updateRequested = False
         self.dirtyTopology = False
@@ -251,6 +252,11 @@ class Graph(BaseObject):
     def isLoading(self):
         """ Return True if the graph is currently being loaded. """
         return self._loading
+    
+    @property
+    def isSaving(self):
+        """ Return True if the graph is currently being loaded. """
+        return self._saving
 
     @Slot(str)
     def load(self, filepath, setupProjectFile=True, importProject=False, publishOutputs=False):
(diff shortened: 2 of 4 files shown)


cbentejac (changes requested):
This PR works as intended but there's an unexpected (and unwanted) side effect to it: when the selected tab is the "Statistics" one, moving from one node to another triggers a polish loop:
`meshroom/ui/qml/GraphEditor/StatViewer.qml:305:9: QML ColumnLayout: Layout polish loop detected for QQuickColumnLayout(0x2055aed0, parent=0x2044f590, geometry=0,0 438x39). Aborting after two iterations.`
(`MESHROOM_OUTPUT_QML_WARNINGS` needs to be set to 1 for the warning to appear) 

There are likely some minor changes to do on the "Statistics" component so it is loaded correctly when it appears as the first component.

I also think the tab index should be reset when we are changing the active project. Currently, it is only reset if we select a non-computable node or if we go back to the homepage, but it never is if we keep on loading projects from the application and selecting computable nodes. What do you think?

meshroom/ui/qml/Application.qml
```diff
@@ -20,6 +20,23 @@ Page {
...
+    property var scenefile: _reconstruction ? _reconstruction.graph.filepath : "";
+
+    // Signal to denote the current meshroom project has been changed
+    signal currentProjectChanged();
+
+    onScenefileChanged: {
+        // Check if we're not currently saving and emit the currentProjectChanged signal
+        if (! _reconstruction.graph.isSaving) {
+            currentProjectChanged();
+        }
```
cbentejac, line 32: Couldn't we do
```suggestion
        if (! _reconstruction.graph.isSaving) {
            nodeEditor.refresh();
        }
```
?

That would prevent us from having one more signal/slot for something that can be done in a single operation.

cbentejac: `currentProjectChanged` could thus be completely removed.

meshroom/ui/qml/Application.qml
```diff
@@ -20,6 +20,23 @@ Page {
     property alias unsavedDialog: unsavedDialog
     property alias workspaceView: workspaceView
 
+    property var scenefile: _reconstruction ? _reconstruction.graph.filepath : "";
```
cbentejac, line 23:
```suggestion
    readonly property var scenefile: _reconstruction ? _reconstruction.graph.filepath : "";
```
Since there's no reason to ever edit this property from the QML side, we can probably set it as read-only.

cbentejac (changes requested):
The polish loop issue has been fixed and everything else is working as expected, but I think there are a couple of simplifications that can be done. 

We're almost there! 😄

meshroom/ui/qml/Application.qml
waaake: Nice catch !!, Thank you 🙂

meshroom/ui/qml/Application.qml
waaake: The intention was to have a callback and then use that callback for this and other purposes in the future. I do agree with you since we don't have anything right now, let's use that itself and we can review this later when it will be needed.

cbentejac (approved):

PR #2626, opened 2024-12-23
[docker] Add Dockerfiles for Rocky 9 and handle Qt 6 installation

cbentejac:
## Description

This PR updates the Dockerfiles for Meshroom as follows:
- The Dockerfiles for CentOS 7 are replaced with Dockerfiles for Rocky 9
- The version of Ubuntu and CUDA is upgraded for the Ubuntu Dockerfiles (from 20.04 to 22.04)
- The installation of Qt is updated to correctly use the installer for Qt 6

## Features list

- [X] Remove CentOS 7 Dockerfiles and scripts
- [X] Update Ubuntu Dockerfiles
- [X] Add new Dockerfiles and scripts for Rocky 9


sebastienmascha:
Thank you for this great PR.

I am new to Meshroom, I would like to know more about the Docker infrastructure design and eventually contribute to the project as well.

For AliceVision and Meshroom, why are we using a `dl` folder and use `wget --no-check-certificate "https://download.qt.io/official_releases/online_installers/qt-unified-linux-x64-online.run` in the `build-*.sh` scripts?

Why not directly include those calls in the `Dockerfile`? It would improve the isolation and avoid unnecessary Build Context.

I am happy to create a PR if needed.

hiaselhans:
@cbentejac since alicevision uses the 22.04 image now and these changes seem due, do you think these will be merged in the near future?

fabiencastan (approved):

PR #2629, opened 2024-12-26
Add relative paths to nodes as variables

cbentejac:
## Description

**This PR will invalidate all pre-existing graphs.**

**Failures of the CI will be solved when the following PR will be merged: https://github.com/alicevision/meshroomNodes/pull/12.**

This PR adds new variables that can be used in the description of nodes to write relative paths. For each `Node` object, there are currently two variables that are available by default:
- `{nodeCacheFolder}`, which points towards the cache folder for this node (and is effectively the same as `desc.Node.internalFolder`);
- `{nodeSourceCodeFolder}`, which points towards the folder containing the node's description.

These variables are resolved with the `cmdVars` dictionary, which contains essentially the {"key": "value"} mapping for all the attributes in a node's command line. `desc.Node.internalFolder` is removed as its purpose is fully replaced by `{nodeCacheFolder}`, and `{nodeSourceCodeFolder}` replaces the absolute path for the Blender script in the `ScenePreview` node.

## Features

- [x] Add `{nodeCacheFolder}` and `{nodeSourceCodeFolder}` to the list of command-line variables
- [x] Fix "Open File" links in the Node Editor

## Implementation remarks

For output attributes, we implement a change that will invalidate almost all pre-existing graphs: when computing an output attribute, we want "/path/to/something" to evaluate to the same thing as "/path/to/something/". "/" on the right are thus stripped from invalidation values when evaluating the UID for an output attribute. This causes all the nodes which had outputs finishing with a "/" to evaluate to a new UID, hence the invalidation of all pre-existing graphs.


meshroom/core/node.py
```diff
@@ -1621,6 +1638,12 @@ def __init__(self, nodeType, nodeDict, position=None, issue=CompatibilityIssue.U
         self.splitCount = self.parallelization.get("split", 1)
         self.setSize(self.parallelization.get("size", 1))
 
+        # Only keep the internal folder in the list of root folders, as the source code folder is not stored
+        # in the node dictionary.
+        self.rootFolder = {
+            "NODE_CACHE_FOLDER": self._internalFolder,
+        }
```
fabiencastan, line 1645:
```suggestion
            "NODE_CACHE_FOLDER": self._internalFolder,
            "NODE_SOURCECODE_FOLDER": "UndefinedPath",
        }
```

fabiencastan: to avoid errors

fabiencastan (changes requested):

meshroom/core/node.py
```diff
@@ -503,6 +503,9 @@ def __init__(self, nodeType, position=None, parent=None, uid=None, **kwargs):
         self._internalFolder = ""
         self._sourceCodeFolder = ""
 
+        # Root folders to write relative paths with
+        self.rootFolder = dict()
```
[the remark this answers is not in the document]
cbentejac: Done

meshroom/core/node.py
cbentejac: Done

meshroom/core/node.py
```diff
@@ -792,6 +793,8 @@ def _buildAttributeCmdVars(cmdVars, name, attr):
                                         format(nodeName=self.name, attrName=attr.name))
                     if defaultValue is not None:
                         try:
+                            # If there is a root folder variable, substitute it before resolving the value
```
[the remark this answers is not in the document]
cbentejac: Done

meshroom/core/desc/node.py
```diff
@@ -13,7 +13,7 @@
 class Node(object):
     """
     """
-    internalFolder = '{cache}/{nodeType}/{uid}/'
+    internalFolder = '{cache}/{nodeType}/{uid}'
```
fabiencastan, line 16: Do we still need it?

meshroom/core/node.py
```diff
@@ -503,6 +503,9 @@ def __init__(self, nodeType, position=None, parent=None, uid=None, **kwargs):
         self._internalFolder = ""
         self._sourceCodeFolder = ""
 
+        # Dynamic environment variables: defines variables that will be replaced with set values when used in attributes
+        self.dynamicEnvVars = dict()
+
```
fabiencastan: On second thought, instead of adding a new concept, these 2 variables should probably just be part of the cmdVars and used as ‘{nodeCacheFolder}’ and ‘{nodeSourceCodeFolder}’. So, we don't mix the dynamic variables and the env variables (that are static by definition).

fabiencastan (changes requested):
Sorry, I should have seen that in the first review...

meshroom/core/desc/node.py
cbentejac: Removed

meshroom/core/node.py

fabiencastan (approved):

fabiencastan:
Need update to fix the CI.

PR #2643, opened 2025-01-13
[ui][fix] Application Menubar: Ensure Menubar occupies all of available height of the layout

waaake:
## Description
This PR modifies the `Menubar` component to ensure it fills the layout in terms of height.
The change addresses an inconsistency where the MenuBar did not align with the Parent's dimensions.

## Details of the change
* Parenting the Menubar inside the Rectangle matching the colour of the background header ensures the additional area is filled correctly.
* Aligning the Menubar at the bottom, leaving a pixel to match the placement of the HomeButton.
* HomeButton and ExecutionButtons have reduced padding to reduce the overall area of the button while retaining the icon size.

The above changes ensure the overall alignment of the Header bar elements, Custom buttons and MenuBar.

## Appearence after the changes
image

### DIFF
--- a/meshroom/ui/qml/Application.qml
+++ b/meshroom/ui/qml/Application.qml
@@ -599,400 +599,422 @@ Page {
         }
     }
 
-    header: RowLayout {
-        spacing: 0
-        MaterialToolButton {
-            id: homeButton
-            text: MaterialIcons.home
+    header: Rectangle {
+        color: Qt.darker(activePalette.window, 1.15)
+        height: 26
 
-            font.pointSize: 18
+        RowLayout {
+            anchors.fill: parent
+            spacing: 0
+            MaterialToolButton {
+                id: homeButton
+                text: MaterialIcons.home
 
-            background: Rectangle {
-                color: homeButton.hovered ? activePalette.highlight : Qt.darker(activePalette.window, 1.15)
-                border.color: Qt.darker(activePalette.window, 1.15)
-            }
+                topPadding: 0
+                bottomPadding: 0
 
-            onClicked: {
-                if (!ensureNotComputing())
-                    return
-                ensureSaved(function() {
-                    _reconstruction.clear()
-                    if (mainStack.depth == 1)
-                        mainStack.replace("Homepage.qml")
-                    else
-                        mainStack.pop()
-                })
-            }
-        }
-        MenuBar {
-            palette.window: Qt.darker(activePalette.window, 1.15)
-            Menu {
-                title: "File"
-                Action {
-                    id: newAction
-                    text: "New"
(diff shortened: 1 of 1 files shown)


cbentejac:
The main issue I have with this change is that, although the menu bar now takes up the whole space as it should, the menus themselves are not vertically centered with the homepage icon. The menu bar starts slightly higher than the homepage button but finishes slightly higher as well, and it's very clear when hovering.

It'd be great if the button and the menu bar could be aligned.

meshroom/ui/qml/Application.qml
```diff
@@ -624,320 +626,327 @@ Page {
...
+             * occupying all of the additional space which the MenuBar is not going to get
+             */
+            Layout.fillWidth: true
+            Layout.fillHeight: true
+            color: Qt.darker(activePalette.window, 1.15)
+
+            MenuBar {
+                palette.window: Qt.darker(activePalette.window, 1.15)
+                anchors.bottom: parent.bottom
+                anchors.bottomMargin: 1     /// this margin counters the offset vertically when lining up against with the homeButton
```
fabiencastan, line 641: Looks better to me without this offset.
```suggestion
```

meshroom/ui/qml/Application.qml
```diff
@@ -624,320 +626,327 @@ Page {
...
+             * This Rectangle serves as the parent of the menubar
+             * occupying all of the additional space which the MenuBar is not going to get
+             */
+            Layout.fillWidth: true
+            Layout.fillHeight: true
+            color: Qt.darker(activePalette.window, 1.15)
+
+            MenuBar {
+                palette.window: Qt.darker(activePalette.window, 1.15)
+                anchors.bottom: parent.bottom
```
fabiencastan, line 640: if we could avoid anchors that's ideal.

meshroom/ui/qml/Application.qml
```diff
@@ -624,320 +626,327 @@ Page {
...
-                }
+        Rectangle {
+            /**
+             * This Rectangle serves as the parent of the menubar
+             * occupying all of the additional space which the MenuBar is not going to get
+             */
+            Layout.fillWidth: true
+            Layout.fillHeight: true
+            color: Qt.darker(activePalette.window, 1.15)
+
```
fabiencastan, line 637: Just using "height: parent.height" on the MenuBar works for me, without adding a Rectangle into the hierarchy.

waaake: This rectangle already existed to cover up additional space horizontally, but I could try using height property to set this, last I can recall, there was a very minor pixel worth of offset against the Custom buttons. Let me try the changes, thanks.

PR #2645, opened 2025-01-14
Extract more metadata using exifTool

demoulinv:
## Description
This PR adds a pure python node dedicated to extract metadata from image files using exifTool. This node can be connected to the cameraInit node and for all images referenced in the "cameraInit.sfm" file, if exifTool is available, a file containing all metadata is generated. The file format can be selected amongst "txt", "xmp", "xml". Be aware that the "xmp" format cannot store all metadata contained in the image file.

## Implementation remarks
The invocated exifTool command depends on the requested file format in the following way:

- "txt": `exiftool [args] imageFile > metadataFile.txt`
- "xml": `exiftool -X [args] imageFile > metadataFile.xml`
- "xmp": `exiftool -tagsfromfile imageFile [args] metadataFile.xmp`

Optional arguments 'args' can be defined through a string parameter of the meshroom node.
The boolean parameter "keepfilename" allows to keep the image file name for the metadata file but changing the file extension.

### DIFF
--- a/meshroom/nodes/aliceVision/ExtractMetadata.py
+++ b/meshroom/nodes/aliceVision/ExtractMetadata.py
@@ -91,9 +91,8 @@ Using exifTool, this node extracts metadata of all images referenced in a sfmDat
             outFiles = self.resolvedPaths(chunk.node.input.value, chunk.node.output.value, chunk.node.keepFilename.value, chunk.node.extension.value)
 
             if not outFiles:
-                error = 'ExtractMetadata: input files listed, but no metadata to extract'
+                error = 'ExtractMetadata: No input files! Check that a sfmData is connected as input.'
                 chunk.logger.error(error)
-                chunk.logger.info('Listed input files: {}'.format([i.value for i in chunk.node.inputFiles.value]))
                 raise RuntimeError(error)
 
             if not os.path.exists(chunk.node.output.value):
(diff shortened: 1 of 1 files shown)


meshroom/nodes/aliceVision/ExtractMetadata.py
```diff
@@ -0,0 +1,124 @@
...
+
+            for iFile, oFile in outFiles.items():
+                if chunk.node.extension.value == 'txt':
+                    cmd = 'exiftool ' + chunk.node.arguments.value.strip() + ' ' + iFile + ' > ' + oFile
+                elif chunk.node.extension.value == 'xml':
+                    cmd = 'exiftool -X ' + chunk.node.arguments.value.strip() + ' ' + iFile + ' > ' + oFile
+                else: #xmp
+                    cmd = 'exiftool -tagsfromfile ' + iFile + ' ' + chunk.node.arguments.value.strip() + ' ' + oFile
+                chunk.logger.debug(cmd)
+                os.system(cmd)
```
servantftechnicolor, line 110: No checks, no return code retrieval, etc

servantftechnicolor (commented):
- I got non explicit errors if nothing is connected to the node
- Nothing appears in the log if exiftool is not found
- On Error, the node does not fail.

meshroom/nodes/aliceVision/ExtractMetadata.py
demoulinv: Messages added

servantftechnicolor (approved):

PR #2650, opened 2025-01-20
[ui][fix] Edge: Fixing an issue with mouse event on Custom EdgeMouseArea causing Crash

waaake:
~~Event::button() was sometimes causing a crash, copying the event and using event.buttons() from the copied event seems to be stable before calling event.button() eventually~~
## Description
This PR addresses a known crash issue with the EdgeMouseArea (QQuickItem)'s mousePressEvent.
The crash occurs when querying button() from the event.
The event is not null and is able to provide other data correctly before ending up crashing when queried for button.

## Features list
- [X] Fix UpdateShape to get invoked on geometryChange
- [X] Fix Alt + Mouse Click to Remove an Edge due to a change in data type for Keyboard Modifiers from int to enum.Flag
- [ ] Fix Crash on clicking egde multiple times

## Implementation remarks
~~Current Fix is based on copying the provided event and calling event.buttons() to check if the buttons are supported and then calling button() is the custom MouseEvent constructor.~~
This seems to have stopped the crashes for now, but still testing on larger scale is pending before we confirm anything.

### DIFF
--- a/meshroom/ui/qml/GraphEditor/Edge.qml
+++ b/meshroom/ui/qml/GraphEditor/Edge.qml
@@ -138,6 +138,7 @@ Item {
         anchors.fill: parent
         acceptedButtons: Qt.LeftButton | Qt.RightButton
         thickness: root.thickness + 4
+        curveScale: cubic.ctrlPtDist / root.width  // Normalize by width
         onPressed: function(event) {
             root.pressed(event)
         }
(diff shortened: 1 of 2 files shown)


meshroom/ui/qml/GraphEditor/Edge.qml
```diff
@@ -149,7 +149,7 @@ Item {
             /* The curve scale must be set only once the component has been fully created, so
              * that all the events following the update of the curve scale can be taken into
              * account. */
-            curveScale = Qt.binding(() => cubic.ctrlPtDist / root.width)  // Normalize by width
+            curveScale = cubic.ctrlPtDist / root.width  // Normalize by width
```
yann-lty, line 152: That can be moved back to a normal binding in the edgeArea (we don't need the Component.onCompleted anymore).
```js
curveScale: cubic.ctrlPtDist / root.width
```

waaake: This is updated.

yann-lty (approved):

PR #2658, opened 2025-01-27
Mask processing node

servantftechnicolor:
See https://github.com/alicevision/AliceVision/pull/1814

### DIFF
--- a/meshroom/nodes/aliceVision/MaskProcessing.py
+++ b/meshroom/nodes/aliceVision/MaskProcessing.py
@@ -1,0 +1,85 @@
+__version__ = "1.0"
+
+from meshroom.core import desc
+from meshroom.core.utils import DESCRIBER_TYPES, VERBOSE_LEVEL
+
+import os.path
+
+class MaskProcessingNodeSize(desc.DynamicNodeSize):
+    """
+    MaskProcessingNodeSize expresses a dependency to multiple input attributess to define
+    the size of a Node in terms of individual tasks for parallelization.
+    """
+    def __init__(self, param):
+        self._params = param
+
+    def computeSize(self, node):
+        
+        size = 0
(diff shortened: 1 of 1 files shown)


meshroom/nodes/aliceVision/MaskProcessing.py
```diff
@@ -0,0 +1,73 @@
...
+        self._params = param
+
+    def computeSize(self, node):
+        
+        size = 0
+
+        for input in node.attribute(self._params).value:
+            paramName = input.getFullName()
+            param = node.attribute(paramName)
+            size = size + param.getLinkParam().node.size
```
fabiencastan, line 23: As we are not accumulating, we should just take the max?
size = max(size, param.getLinkParam().node.size)

meshroom/nodes/aliceVision/MaskProcessing.py
```diff
@@ -0,0 +1,74 @@
...
+        )
+    ]
+
+    outputs = [
+        desc.File(
+            name="output",
+            label="Output",
+            description="Path to the output directory.",
+            value=desc.Node.internalFolder,
+        )
```
fabiencastan, line 73: Need to expose the output masks to the Viewer.
Here is an example: https://github.com/alicevision/Meshroom/blob/develop/meshroom/nodes/aliceVision/ImageSegmentation.py#L86

servantftechnicolor: > Need to expose the output masks to the Viewer. Here is an example: https://github.com/alicevision/Meshroom/blob/develop/meshroom/nodes/aliceVision/ImageSegmentation.py#L86

How do i do that ? I have no clue about the masks names as it may come from various sources.

PR #2663, opened 2025-02-03
[ui] Edge: Workaround on the Crash due to Edge Connections

waaake:
## Description
This PR addresses a crash issue by working around the issue of querying buttons for the MouseEvent.
**MouseEvent::button()** or **MouseEvent::buttons()** on the `mousePressEvent` or in the `Custom MouseEvent` was causing crashes  some times when an event provided has issues.
While the underlying issues at this point are not clear, it is evident that the crash is happening due to the PySide binding.
A similar implementation was tested on the C++ side which seems to be working fine.

## Features list
[X] Workaround for edge crashing by implementing a Mouse Area which checks the pressed point to exist on the edge before emitting a pressed/released event.

## Implementation remarks
* This PR introduces a MouseArea which follows the edge, any clicks on the mouse area checks if the point exists on the edgeArea, if so a relevant pressed/released event gets triggered, else the event is ignored.

### DIFF
--- a/meshroom/ui/components/edge.py
+++ b/meshroom/ui/components/edge.py
@@ -1,4 +1,4 @@
-from PySide6.QtCore import Signal, Property, QPointF, Qt, QObject
+from PySide6.QtCore import Signal, Slot, Property, QPointF, Qt, QObject
 from PySide6.QtGui import QPainterPath, QVector2D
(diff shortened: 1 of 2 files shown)


fabiencastan:
Resolved with qt >= 6.7

PR #2664, opened 2025-02-03
Create new pipeline for testing modular sfm

servantftechnicolor:
- Create experimental pipelines templates for 

1. tracking
2. tracking without calibration
3. photogrammetry
4. photogrammetry with tracking

- Update meshroom to consider the new sfm Nodes as if it was the previous StructureFromMotion node

Requires https://github.com/alicevision/AliceVision/pull/1818


mh0g (approved):
Tested the camera tracking pipeline vs the original camera tracking on sequ 035 of anna.
After aligning the two set of poses, quantitatively the difference was extremely small for the average center distance and rotation angles:
image

PR #2666, opened 2025-02-06
[ui] Fix: Improve UX for search by allowing Search focus to be removed when clicking outside box

waaake:
## Description
This PR addresses an issue where Search Text box focus was not getting stolen when clicking on few components in the Active UI.

## Features list
- [X] Mouse Area for Node Editor
- [X] Mouse Area for Inspector 3D List View
- [X] Mouse Area for Top Menu Bar
- [X] Mouse Area for Bottom Status Bar

## Implementation remarks
Added MouseArea(s) to components allowing clicks to force focus on the parent components thereby stealing focus from anything that may have it, including the Search Input Text.

### DIFF
--- a/meshroom/ui/qml/Application.qml
+++ b/meshroom/ui/qml/Application.qml
@@ -927,6 +927,13 @@ Page {
         }
 
         Rectangle {
+            id: menuSpacer
+            MouseArea {
+                anchors.fill: parent
+                onClicked: {
+                    menuSpacer.forceActiveFocus();
+                }
+            }
             Layout.fillWidth: true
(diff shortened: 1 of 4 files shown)


nicolas-lambert-tc:
I propose an other approach on this [PR](https://github.com/alicevision/Meshroom/pull/2740)
It should avoid future maintanability issues. _(having to handle MouseArea on all the existing and future component that break the classical focus behavior)_
It uses the `EventFilter` and can have some performance cost, but because it directly filtered event by type() _(not isinstance)_ , it should be imperceptible.

PR #2672, opened 2025-02-14
[ui] Graph: In minimal refresh, do not poll files for chunks run locally

cbentejac:
## Description

When the minimal refresh mode is on, only chunks that are running, submitted or in error while their execution mode is external should be polled.

Chunks that have the same status but with a local execution mode will be correctly updated as their status changes will be notified.

The behaviour for the automatic refresh mode remains the same: all the chunks are polled, independently from their status and their execution mode.

### DIFF
--- a/meshroom/ui/graph.py
+++ b/meshroom/ui/graph.py
@@ -38,7 +38,7 @@ from meshroom.ui.utils import makeProperty
 class PollerRefreshStatus(Enum):
     AUTO_ENABLED = 0  # The file watcher polls every single status file periodically
     DISABLED = 1  # The file watcher is disabled and never polls any file
-    MINIMAL_ENABLED = 2  # The file watcher only polls status files for chunks that are either submitted or running
+    MINIMAL_ENABLED = 2  # The file watcher only polls status files for chunks that are either submitted or running externally
 
 
 class FilesModTimePollerThread(QObject):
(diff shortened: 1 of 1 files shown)


meshroom/ui/graph.py
```diff
@@ -183,10 +183,12 @@ def watchedStatusFiles(self):
             return self.statusFiles, self.monitorableChunks
         elif self.filePollerRefresh is PollerRefreshStatus.MINIMAL_ENABLED.value:
             for c in self.monitorableChunks:
-                # When a chunk's status is ERROR, it may be externally re-submitted and it should thus still be monitored
-                if c._status.status is Status.SUBMITTED or c._status.status is Status.RUNNING or c._status.status is Status.ERROR:
-                    files.append(c.statusFile)
-                    chunks.append(c)
+                # Only chunks that are run externally should be monitored; when run locally, status changes are already notified
+                if c._status.execMode is ExecMode.EXTERN:
```
yann-lty, line 187: That solves the conflict issue for monitoring chunks being computed in the current meshroom instance :+1:.

But could we even use `c.isExtern()` here? 
To handle both remote computation and local computation happening within another Meshroom instance (the 2 cases were monitoring via polling makes sense)?

yann-lty:
:+1:

meshroom/ui/graph.py
cbentejac: You're totally right, I'll make that change!

cbentejac: Done!

yann-lty (approved):

PR #2674, opened 2025-02-17
Add maya scene export

servantftechnicolor:
- ExportMaya node was exporting a mel script
- This is now also using maya_batch to export a mb scene file


mh0g (approved):
tested on a dummy scene, the export creates a .mg that can be opened in maya direclty

PR #2676, opened 2025-02-19
[ui] app: Register components to QML before instantiating the engine

cbentejac:
## Description

This PR moves the QML registration of components before the instantiation of the QML engine. 

Functionally speaking, this does not change anything as Qt 6.6.x supports the registration of all types, including singleton, even after the instantiation of the engine, but this will not be the case anymore with Qt 6.7.x and above.


yann-lty (approved):
Great find, works like a charm on 6.8 :).

PR #2699, opened 2025-03-18
[ui] NodeEditor: Don't allow to reduce split view size below 350

cbentejac:
## Description

This prevents reducing the Node Editor up to the point where components within it (especially in its header) are disappearing from screen.


Alxiice (approved):
Okay after rebase, but maybe also increase the `minimumWidth` for the `TabPanel`

cbentejac:
> Okay after rebase, but maybe also increase the `minimumWidth` for the `TabPanel`

Done!

PR #2722, opened 2025-05-13
[ui] AttributeEditor: Allow displaying attibute in corresponding viewport 

nicolas-lambert-tc:
## Description

In the `NodeEditor`:
- an eye icon is added after the attribute's label when the current attribute is displayed in the corresponding viewport. (2d or 3d)
- `Double click` on an attribute display it in the corresponding viewport if it's possible
- `Show in viewport`menu ittem has been added in the context menu when the attribute can be displayed 

### Icon
attribute-viewer-2dfile
attribute-viewer-3dfile

### Context menu
attribute-viewer-contextmenu

Features list

- [X] an eye icon is added after the attribute's label when the current attribute is displayed in the corresponding viewport. (2d or 3d)
- [X] `Double click` on an attribute display it in the corresponding viewport if it's possible
- [X] `Show in viewport`menu ittem has been added in the context menu when the attribute can be displayed 

## Implementation remarks

Valid displayable 3d extensions and attribute descritpiton semantics are now in Attribute static attribute.
How about move it in a config file?


fabiencastan:
It would be good to always display the eye icon when the attribute can be displayed, maybe something similar to the 3D Viewer:
image

fabiencastan:
Use Upper Camel Case in menus, so "Show in Viewport", instead of "Show in viewport"
image
In the title of the widgets, we use "Image Viewer" and "3D Viewer", so maybe "Show in Viewer".

meshroom/ui/reconstruction.py
```diff
@@ -474,6 +474,9 @@ def __init__(self, undoStack: commands.UndoStack, taskManager: TaskManager, defa
         self._activeNodes = meshroom.common.DictModel(keyAttrName="nodeType")
         self.initActiveNodes()
 
+        # initialize activeAttriubtes (attributes currently visible in some viewports)
```
fabiencastan, line 477: typo

meshroom/ui/qml/GraphEditor/AttributeItemDelegate.qml
```diff
@@ -180,7 +181,8 @@ RowLayout {
                 color: palette.mid
                 font.pointSize: 8
                 padding: 4
-            }
+            }
```
fabiencastan, line 198: empty spaces

meshroom/ui/qml/GraphEditor/AttributeItemDelegate.qml
```diff
@@ -162,6 +163,19 @@ RowLayout {
                         forceActiveFocus()
                         if (mouse.button == Qt.RightButton) {
                             var menu = menuComp.createObject(parameterLabel)
+                            
+                            // Could not found a way to do edit declaratively
```
fabiencastan, line 167: Is it not possible to create a static Menu and open it from here?

nicolas-lambert-tc: I've tired but i have some weird behaviors in the context menu, the item is displaying and disappearing some times ?

cbentejac:
It's not introduced by this PR but it is highlighted by it (since there are more things to display on a single line): image

The button doesn't fill the space vertically (and potentially isn't centered vertically either, I haven't checked the code) which creates a strange offset between the icons.

meshroom/ui/qml/GraphEditor/AttributeItemDelegate.qml
```diff
@@ -180,29 +182,53 @@ RowLayout {
...
 
                 property bool shouldBeVisible: (attribute != undefined && attribute.hasOutputConnections)
 
-                text: shouldBeVisible ? MaterialIcons.logout : " "
+                text: MaterialIcons.logout
                 font.pointSize: 8
                 enabled: shouldBeVisible
                 Layout.alignment: Qt.AlignTop | Qt.AlignRight
                 topPadding: 7
+                opacity: shouldBeVisible ? 1 : 0
```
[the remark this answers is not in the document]
nicolas-lambert-tc: It's a relicate of the way the icon was displayed. I'll apply your suggestion

meshroom/ui/qml/GraphEditor/AttributeItemDelegate.qml
nicolas-lambert-tc: Thanks to @cbentejac, now the item is created in a declarative way.

meshroom/ui/qml/GraphEditor/AttributeItemDelegate.qml
```diff
@@ -193,16 +207,29 @@ RowLayout {
...
 
+            MaterialLabel {
+                property bool isDisplayable: attribute.isOutput && (attribute.is2D || attribute.is3D)
+                property bool isDisplayed: attribute === _reconstruction.displayedAttr2D || _reconstruction.displayedAttrs3D.count && _reconstruction.displayedAttrs3D.contains(attribute)
+                text: isDisplayed ? MaterialIcons.visibility : MaterialIcons.visibility_off
+                enabled: isDisplayed
+                visible: isDisplayable
+
+                Layout.alignment: Qt.AlignRight
+                rightPadding: 5
```
cbentejac, line 218: The alignment on the right is not needed here since the label cannot move freely, so I'd remove it. Same for the top padding, as the label seems to be vertically centered by default. 

However, the "advanced" button and the navigation tool buttons have a global padding of 4 pixels, so for the sake of harmony, this one should also be set to 4.

```suggestion
                padding: 4
```

meshroom/core/attribute.py
```diff
@@ -446,6 +449,28 @@ def updateInternals(self):
         # Emit if the enable status has changed
         self.setEnabled(self.getEnabled())
 
+    def _is3D(self) -> bool:
+        """ Return True if the current attribute is considered as a 3d file """
+
+        # List of supported extensions, taken from Viewer3DSettings
```
cbentejac, line 455: Leftover comment, to be removed since the list of supported extensions has been moved upwards.
```suggestion

```

meshroom/ui/qml/Application.qml
```diff
@@ -1105,22 +1105,45 @@ Page {
                     for (var i = 0; i < node.attributes.count; i++) {
                         var attr = node.attributes.at(i)
                         if (attr.isOutput && attr.desc.semantic !== "image")
-                            if (!alreadyDisplay || attr.desc.semantic == "3D")
+                            if (!alreadyDisplay || attr.desc.semantic == "3D") {
```
cbentejac, line 1108: Not introduced by this PR but since the line has been modified... isn't there a typo here? Shouldn't the check on the semantic be for `"3d"` instead of `"3D"`? (or maybe I'm missing something else)
```suggestion
                            if (!alreadyDisplay || attr.desc.semantic == "3d") {
```

cbentejac (changes requested):
Apart from the code comments, it might be a good idea to add a tooltip on the "eye" icon (as well as the tool buttons) explaining what it means. 

Otherwise everything is working perfectly fine, I couldn't find a single issue or a QML warning :)

cbentejac (approved):

PR #2723, opened 2025-05-14
[ui] Viewer2D: Add the pixel (x,y) values in the toolbar (editable)

nicolas-lambert-tc:
## Description
Add the ability to display and edit `x, y` value of the current `rgba color` displayed in the viewer2D.

viewer2d-xy

## Features list

- [X] Add x and y text fields to `viewer2D` to display the pixel's coordinates 
- [X] x and y can be edited by user to select a specific pixel

### DIFF
--- a/meshroom/ui/qml/Viewer/HdrImageToolbar.qml
+++ b/meshroom/ui/qml/Viewer/HdrImageToolbar.qml
@@ -23,8 +23,12 @@ FloatingPane {
     property real gammaValue: Math.pow(gammaCtrl.value, slidersPowerValue)
     property alias channelModeValue: channelsCtrl.value
     property variant colorRGBA: null
+    property variant mousePosition: {return {x:0, y:0} }
 
     property bool colorPickerVisible: true
+
+    property int pixelX: 0
+    property int pixelY: 0
 
     background: Rectangle { color: root.palette.window }
 
@@ -136,6 +140,33 @@ FloatingPane {
             }
         }
 
+        RowLayout {
+
+            Label {
+                text: "x"
+            }
+            TextField {
+                id: xPixel
+                text: root.mousePosition.x
+                Layout.preferredWidth: 50
+                onTextEdited: {
+                    pixelX = parseInt(xPixel.text)
+                }
+            }
+            Label {
+                text: "y"
+            }
+            TextField {
+                id: yPixel
+                text: root.mousePosition.y
+                Layout.preferredWidth: 50
+                onTextEdited: {
+                    pixelY = parseInt(yPixel.text)
+                }
+            }
+
+        }
+
         Rectangle {
             visible: colorPickerVisible
             Layout.preferredWidth: 20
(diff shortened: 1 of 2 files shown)


meshroom/ui/qml/Viewer/HdrImageToolbar.qml
```diff
@@ -23,9 +23,13 @@ FloatingPane {
     property real gammaValue: Math.pow(gammaCtrl.value, slidersPowerValue)
     property alias channelModeValue: channelsCtrl.value
     property variant colorRGBA: null
+    property variant mousePosition: {return {x:0, y:0} }
```
cbentejac, line 26:
```suggestion
    property variant mousePosition: ({x: 0, y: 0})
```
The initialization of `x` and `y` remains the same, but this is more QML-like (as a reference: https://doc.qt.io/qt-6.5/qml-var.html#property-value-initialization-semantics)

meshroom/ui/qml/Viewer/Viewer2D.qml
```diff
@@ -427,17 +427,30 @@ FocusScope {
...
+
+            mousePosition: {
+                return {
+                    x: Math.floor(floatImageViewerLoader.item.mouseX), 
+                    y: Math.floor(floatImageViewerLoader.item.mouseY)
                 }
-                var pix = floatImageViewerLoader.item.pixelValueAt(Math.floor(floatImageViewerLoader.item.mouseX),
-                                                                   Math.floor(floatImageViewerLoader.item.mouseY))
-                return pix
             }
```
cbentejac, line 452: Same here, since there's no condition and the property is simply reset, we can get rid of the `return` and set it with `({x: ..., y: ...})`.

cbentejac (changes requested):
Everything works well as long as the mouse is hovering over the image. 

If it is hovering in the 2D Viewer but outside of the image, the (x, y) coordinates display the last "recorded" position of the mouse on the image, but these displayed values do not correspond to the actual values of `mousePosition` and so the displayed RGB values are off as well. What seems to happen is that `x` and `y` are reset to 0 as soon as the mouse goes out of the image, the RGB values are updated to match the colors of pixel (0, 0), but the text fields for `x` and `y` still display the previous values. Modifying the values works, but for example, if I only modify `x`, `y` will still display the last recorded position while its value will actually still be 0.

Prior to this PR, the behaviour for RGB values was to display "--" when the mouse position was outside of the image. This can be discussed but I think we might want the same for the (x, y) coordinates: when hovering outside of the image, the values become "unset", and should the user choose to set them, only then would we display the corresponding RGB values.

cbentejac:
There's a remaining "issue" when the mouse is hovering outside of the picture: RGB values are still displayed while (x, y) are empty:
Screenshot_2025-05-19_15-02-09

Also, I think it'd be visually better if the (x, y) coordinates displayed "--" instead of nothing when the mouse isn't hovering over any pixel.

cbentejac (changes requested):
Everything's working functionally, just a few minor changes and we'll be good to go!

cbentejac (approved):

PR #2724, opened 2025-05-15
[ui] Homepage: Project can be removed by right click

nicolas-lambert-tc:
## Description

Add a context menu on the project to allow the deletion of a specific project

image

Add a 'Project ..." item in the main menu to allow navigation to the project page

image

## Features list

- [X] Homepage: ContextMenu to open/delete a project
- [X] MainMenu: Add `Projects ...` to have the possibility to list the projects when a project is already open

### DIFF
--- a/meshroom/ui/qml/Homepage.qml
+++ b/meshroom/ui/qml/Homepage.qml
@@ -9,6 +9,14 @@ import Controls 1.0
 Page {
     id: root
 
+    function setCurrentTab(tabName) {
+        const tabIndex = tabPanel.tabs.indexOf(tabName)
+
+        if (tabIndex) {
+            tabPanel.currentTab = tabIndex
+        }
+    }
+
     onVisibleChanged: {
         logo.playing = false
         if (visible) {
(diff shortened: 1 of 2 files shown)


cbentejac (changes requested):
The "Projects..." action is not necessary as we have the "Home" icon next to the "File" menu, so that part from the second commit can be dropped. 

I also don't think there's a need to make a distinction between the simple and the double click: if the user left-clicks (once) on a projet, we should try to open it as is currently the case; if the user right-clicks, then the context menu should be opened, with the "Open" and "Delete" options.

meshroom/ui/qml/Homepage.qml
```diff
@@ -9,6 +9,14 @@ import Controls 1.0
...
     id: root
 
+    function setCurrentTab(tabName) {
+        const tabIndex = tabPanel.tabs.indexOf(tabName)
+
+        if (tabIndex) {
+            tabPanel.currentTab = tabIndex
+        }
+    }
+
```
cbentejac, line 19: Is this function still necessary with the recent changes? I don't see it used anywhere.

nicolas-lambert-tc: Thx, you're right, it's an oversight.

PR #2725, opened 2025-05-22
[ui] Attribute / Node: Refacto of the attribute validation paradigm

nicolas-lambert-tc:
## Description

[Alicevision related PR](https://github.com/alicevision/AliceVision/pull/1875)

Add a standard way to validate the attribute value.
It add a way to apply validators to an attribute description and to represent in the UI when these attributes are not valid.

Screenshot_2025-05-23_17-35-17

image

#### Validators
```python
class AttributeValidator(object):

    def __call__(self, node: "Node", attribute: "Attribute") -> tuple[bool, list[str]]: raise NotImplementedError()


class NotEmptyValidator(AttributeValidator):

    def __call__(self, node: "Node", attribute: "Attribute") -> tuple[bool, list[str]]:

        if attribute.value is None or attribute.value == "":
            return error("Empty value are not allowed")
        
        return success()
```

#### Validators usage (can be a lambda)

```python
desc.StringParam(
            name="input",
            label="Test input",
            description="",
            value="",
            validators=[
                NotEmptyValidator(),
                lambda node, attribute: success() if attribute == "test" else error("Attribute value should be 'test'")
                ]
        ),
```

## Features list
- [X] Add the notion of reusable validators
- [X] Validator embed the error messages
- [X] Validator can be a lambda
- [X] An icon with error message appear when an attribute value is invalid (at least one validator return error messages)
- [X] A node containing an invalid attribute is tagged with an orange icon and the selection rectangle is orange
- [X] Attribute value is checked every time the attribute value is set (in the Command) 
- [X] A Warning Dialog is displayed before trying to submit a graph where a node is in warning state


meshroom/ui/qml/Application.qml
```diff
@@ -399,6 +417,26 @@ Page {
...
 
+        MessageDialog {
+            id: submitWithWarningDialog
+
+            canCopy: false
+            icon.text: MaterialIcons.warning
+            parent: Overlay.overlay
+            preset: "Warning"
+            title: "Nodes have warnings"
+            text: "Some nodes have warnings, are you sure you want to submit ?"
```
fabiencastan, line 428:
```suggestion
            text: "Some nodes contain warnings. Are you sure you want to submit?"
```

meshroom/ui/qml/Application.qml
```diff
@@ -399,6 +417,26 @@ Page {
...
         }
 
+        MessageDialog {
+            id: submitWithWarningDialog
+
+            canCopy: false
+            icon.text: MaterialIcons.warning
+            parent: Overlay.overlay
+            preset: "Warning"
+            title: "Nodes have warnings"
```
fabiencastan, line 427:
```suggestion
            title: "Nodes Containing Warnings"
```

meshroom/ui/qml/GraphEditor/Node.qml
```diff
@@ -272,6 +278,17 @@ Item {
                                 }
                             }
 
+                            // Attribute warnings
+                            MaterialLabel {
+                                visible: hasWarnings
+                                text: MaterialIcons.fmd_bad
+                                color: "orange"
```
fabiencastan, line 285: It would be good to specify a "warning" color in the Colors file.

PR #2733, opened 2025-05-27
New Plugin Architecture for Node Registration

cbentejac:
## Description

This PR adds the support for plugins in Meshroom, building on top of https://github.com/alicevision/Meshroom/pull/2703.

A `Plugin` is a collection of at least one `NodePlugin`, which contains the description of a node, and can be associated to templates. `Plugins` can be loaded and unloaded with the `NodePluginManager`, which is in charge of the registration and unregistration of all `NodePlugins` (a node cannot be instantiated until its `NodePlugin` has been registered).

All collections of nodes are now considered as plugins, may they be provided through the `MESHROOM_NODES_PATH` or through the `MESHROOM_PLUGINS_PATH` environment variables. In the case of `MESHROOM_NODES_PATH`, a Python module containing node descriptions is expected. For `MESHROOM_PLUGINS_PATH`, it is expected that the nodes are contained within a Python module that is itself located in a `meshroom` folder. 

## Features list

- [x] Add a "plugins" `core` module containing new classes: `Plugin`, `NodePlugin`, `NodePluginManager`;
- [x] Fully manage the loading and registration of nodes through the `NodePluginManager`;
- [x] Add unit tests for the newly added classes.

## Implementations remarks

Currently, the whole process of going through folders in search of Python packages remains in the `__init__.py` of the `core` module. This is because the method that is used to find files containing nodes is the same as the one used to discover submitters. Submitters are, for now, not included in the "plugins" module and are handled as they were previously.

### DIFF
--- a/tests/test_invalidation.py
+++ b/tests/test_invalidation.py
@@ -7,8 +7,10 @@ from meshroom.core import desc, registerNodeType
 class SampleNode(desc.Node):
     """ Sample Node for unit testing """
     inputs = [
-        desc.File(name='input', label='Input', description='', value='',),
-        desc.StringParam(name='paramA', label='ParamA', description='', value='', invalidate=False)  # No impact on UID
+        desc.File(name="input", label="Input", description="", value="",),
+        desc.StringParam(name="paramA", label="ParamA",
+                         description="", value="",
+                         invalidate=False)  # No impact on UID
     ]
     outputs = [
         desc.File(name='output', label='Output', description='', value="{nodeCacheFolder}")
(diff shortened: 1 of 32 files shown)


tests/test_invalidation.py
```diff
@@ -1,28 +1,30 @@
...
     ]
     outputs = [
         desc.File(name='output', label='Output', description='', value="{nodeCacheFolder}")
     ]
 
-
-registerNodeType(SampleNode)
-
+nodePlugin = NodePlugin(SampleNode)
+pluginManager.registerNode(nodePlugin)  # register standalone NodePlugin
```
fabiencastan, line 21: Use a new dedicated method `pluginManager.registerNodeDesc(SampleNode)`
and later unregisterNodeDesc()

fabiencastan (approved):

PR #2740, opened 2025-06-05
[ui] [fix]: The focus on text components is now released on outside click

nicolas-lambert-tc:
### Bug: 
Clicking outside of a text component (a search filter in the example below), didn't release the focus.
Screenshot_2025-06-05_14-21-39

Now the focus is stolen by the main.MouseArea component

## Related PR
This is an other approach of the PR provided by @waaake: https://github.com/alicevision/Meshroom/pull/2666
It avoid creating `MouseArea` on existing and future components.

## Implementation Details

It uses the eventFilter to only catch the mousePress events and clear focus each time an item have a focus at this step.

/!\ Having a python condition on each event can be time consuming.
It uses the `EventFilter` and can have some performance cost, but because it directly filtered event by type() _(not isinstance)_ , it should be imperceptible.

### DIFF
--- a/meshroom/ui/__main__.py
+++ b/meshroom/ui/__main__.py
@@ -10,4 +10,6 @@ import meshroom.ui
 import meshroom.ui.app
 
 meshroom.ui.uiInstance = meshroom.ui.app.MeshroomApp(sys.argv)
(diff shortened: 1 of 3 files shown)


fabiencastan (changes requested):
Problem on copy/paste

PR #2745, opened 2025-06-11
Dev/connect group attributes and avoid cyclic dependency

nicolas-lambert-tc:
## Description
This PR is a rebase of this PR from @yann-lty : https://github.com/alicevision/Meshroom/pull/2684 with some adjustment.

## Features list

- [X] Support connection between GroupAttributes (API, serialization, UI).
- [X] [ui] Make connected GroupAttribute readonly.

- [X] GroupAttributeConnection do not replace the attributes, but copy value only
- [X] Only GroupAttribute with same structure can be connected. (Ordered sub attributes with same types)
- [X] Nested GroupAttributes are handled recursively

- [X] Cyclic dependency is now forbidden

## Implementation remarks

- A new DAGVisitor has been added and used as default visitor to avoid the cyclic dependency
- Attributes are now responsible of the connection validation

### DIFF
--- a/meshroom/core/attribute.py
+++ b/meshroom/core/attribute.py
@@ -881,10 +881,21 @@ class GroupAttribute(Attribute):
         if not linkedParam:
             return self._value
         
+        def linkAttributesValues(srcAttr, dstAttr):
+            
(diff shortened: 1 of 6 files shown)


cbentejac:
Integrated in #2965.

PR #2747, opened 2025-06-12
[core] Runtime-specific environments support

cbentejac:
## Description

This PR builds on top of #2703 and #2733, which respectively added the notion of local isolated computations for nodes, and the concept of plugins. This pull request adds the notion of runtime-specific environment, allowing nodes to be executed in a dedicated environment that contains their required dependencies independently from Meshroom's. 

Two types of execution environments are added in this PR:

1. **DirTreeProcessEnv**

Based on the root folder of the plugin, it sets up the environment (`PYTHONPATH`, `PATH` and `LD_LIBRARY_PATH` environment variables). It assumes the plugin has a standard hierarchy:
```
mrPlugin/
|-- bin
|-- lib
|   `-- pythonX.X
|       `-- site-packages
|           |-- anotherModule
|           `-- myCustomModule
`-- meshroom
    |-- myNodes
    |   |-- NodeA.py
    |   |-- NodeB.py
    |   |-- NodeC.py
    |   `-- __init__.py
    |-- myPipelineA.mg
    `-- myPipelineB.mg
```
To declare a plugin using the DirTree, set the env variable `MESHROOM_PLUGINS_PATH`.

2. **RezProcessEnv**

Based on a [Rez](https://rez.readthedocs.io) package that's provided through a dedicated environment variable, it gathers all the requirements to execute the process within that Rez environment.

To declare plugins using Rez for their execution, use the `MESHROOM_REZ_PLUGINS` environment variable.

Rez plugins need to be provided to `MESHROOM_REZ_PLUGINS` with the following syntax: `packageName=/path/to/root`. Their `package.py` also need to follow a certain structure. An example is given below for a package named "mrExample":
```
name="mrExample"
version="1.0"

plugin_for=["meshroom"]
requires = []

def commands():
    subRequires = [
        'python-3',
        'dependency1-1.0',
        'dependency2-1.0',
        'dependency3'
    ]
    env.MESHROOM_REZ_PLUGINS.append(this.name + "={root}/folder")
    env.MREXAMPLE_SUBREQUIRES = ":".join(subRequires)
``` 

## Features list

- [x] Add a new `MESHROOM_REZ_PLUGINS` environment variable to load Rez plugins;
- [x] Use an isolated and node-specific runtime environment for all the nodes. 

## Implementation remarks

For Rez plugins, a few things are to be noted:
1. Meshroom's ability to detect and set up the runtime environment for Rez plugins relies on the "name=path" pair being correctly provided to `MESHROOM_REZ_PLUGINS`.
2. Setting `env.MRNAME_SUBREQUIRES` in the Rez package's `package.py` is necessary, as it is used by Meshroom to list the dependencies.
5. Providing versions of packages for the subrequires is not mandatory: if a version is provided, it will strictly be respected; otherwise, Meshroom will attempt to match the package's version with its own, if it exists in its environment.
6. When the node is executed in its dedicated Rez environment, both Meshroom's path and the Rez package's are injected into the `PYTHONPATH` environment variable, strictly within the Rez environment.

### DIFF
--- a/meshroom/core/desc/node.py
+++ b/meshroom/core/desc/node.py
@@ -260,8 +260,8 @@ class Node(BaseNode):
             meshroomComputeCmd += f" --iteration {chunk.range.iteration}"
 
         runtimeEnv = chunk.node.nodeDesc.plugin.runtimeEnv
-        cmdPrefix = chunk.node.nodeDesc.plugin.processEnv.getCommandPrefix()
-        cmdSuffix = chunk.node.nodeDesc.plugin.processEnv.getCommandSuffix()
+        cmdPrefix = chunk.node.nodeDesc.plugin.commandPrefix
+        cmdSuffix = chunk.node.nodeDesc.plugin.commandSuffix
         self.executeChunkCommandLine(chunk, cmdPrefix + meshroomComputeCmd + cmdSuffix, env=runtimeEnv)
 
 
@@ -279,8 +279,8 @@ class CommandLineNode(BaseNode):
         return MrNodeType.COMMANDLINE
 
     def buildCommandLine(self, chunk) -> str:
-        cmdPrefix = chunk.node.nodeDesc.plugin.processEnv.getCommandPrefix()
-        cmdSuffix = chunk.node.nodeDesc.plugin.processEnv.getCommandSuffix()
+        cmdPrefix = chunk.node.nodeDesc.plugin.commandPrefix
+        cmdSuffix = chunk.node.nodeDesc.plugin.commandSuffix
         if chunk.node.isParallelized and chunk.node.size > 1:
             cmdSuffix = cmdSuffix + " " + self.commandLineRange.format(**chunk.range.toDict())
 
--- a/meshroom/core/desc/node.py
+++ b/meshroom/core/desc/node.py
@@ -78,8 +78,9 @@ class BaseNode(object):
     outputs = []
     size = StaticNodeSize(1)
     parallelization = None
-    documentation = ''
-    category = 'Other'
+    documentation = ""
+    category = "Other"
+    plugin = None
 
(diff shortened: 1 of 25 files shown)


meshroom/core/plugins.py
```diff
@@ -46,16 +47,123 @@ def validateNodeDesc(nodeDesc: desc.Node) -> list:
...
+    def __init__(self, folder: str):
+        super().__init__(folder, ProcessEnvType.DIRTREE)
+
+        self.binPaths: list = [str(Path(folder, "bin"))]
+        self.libPaths: list = [str(Path(folder, "lib")), str(Path(folder, "lib64"))]
+        self.pythonPaths: list = [str(Path(folder))] + self.binPaths
+
+    def getEnvDict(self) -> dict:
+        env = os.environ.copy()
+        env["PYTHONPATH"] = f"{_MESHROOM_ROOT}{os.pathsep}{os.pathsep.join(self.pythonPaths)}{os.pathsep}{os.getenv('PYTHONPATH', '')}"
```
fabiencastan, line 97: This looks easier to read to me (avoiding the repetition of pathsep):
```suggestion
        env["PYTHONPATH"] = os.pathsep.join([f"{_MESHROOM_ROOT}"] + self.pythonPaths + [os.getenv('PYTHONPATH', '')])
```

meshroom/core/desc/node.py
```diff
@@ -277,17 +279,17 @@ def getMrNodeType(self):
         return MrNodeType.COMMANDLINE
 
     def buildCommandLine(self, chunk):
```
fabiencastan, line 281:
```suggestion
    def buildCommandLine(self, chunk) -> str:
```

fabiencastan (approved):

PR #2778, opened 2025-07-24
[core] plugins: Load plugin's configuration file upon its initialisation

cbentejac:
## Description

This PR introduces a new feature to load and process a plugin's configuration file, if it has one.

When initialising a plugin, Meshroom now looks for a `config.json` file that contains the environment variables that need to be set for it to work properly:
- If such a file is found, the environment variables are automatically set. In particular, if variables such as paths to files or folders need to be set, they are checked for and converted to absolute paths.
- If no such file exists, nothing happens.

A plugin's configuration file is always expected to be named `config.json`, placed at the-top level of the plugin (next to its templates), and be written as follows:
```json
[
    {
        "key": "key1",
        "type": "path",
        "value": "../value1"
    },
    {
        "key": "key2",
        "type": "string",
        "value": "value2"
    },
    {
        "key": "key3",
        "value": "value3"
    },
    {
        "key": "key4",
        "type": "path",
        "value": "/path/to/value4"
    }
]
```
The "type" key may or may not be provided, but in case it is not, Meshroom will assume the variable is of type "string". Variables that provide paths need to be types as paths: this way, the value is correctly resolved before being written, meaning both absolute and relative paths can be used.

### DIFF
--- a/meshroom/core/plugins.py
+++ b/meshroom/core/plugins.py
@@ -76,10 +76,11 @@ class ProcessEnv(BaseObject):
         self._configEnv: dict[str: str] = configEnv
         self._processEnvType: ProcessEnvType = envType
         self.uri: str = uri
+        self._env: dict = None
 
     def getEnvDict(self) -> dict:
         """ Return the environment dictionary if it has been modified, None otherwise. """
-        return None
+        return self._env
 
     def getCommandPrefix(self) -> str:
         """ Return the prefix to the command line that will be executed by the process. """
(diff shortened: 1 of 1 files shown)


meshroom/core/plugins.py
```diff
@@ -285,6 +287,42 @@ def loadTemplates(self):
...
+                        if t == "path":
+                            resolvedPath = Path(os.path.join(self.path, val)).resolve()
+                            if resolvedPath.exists():
+                                val = resolvedPath.as_posix()
+
+                        # Do not overwrite env variables: if the user has already defined them by
+                        # hand, then that value should take precedence over the one in the config
+                        # file.
+                        if not os.getenv(k, None):
+                            os.environ[k] = val
```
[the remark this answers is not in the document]
cbentejac: Done

meshroom/core/plugins.py
```diff
@@ -126,6 +126,14 @@ def getEnvDict(self) -> dict:
...
         env["PATH"] = f"{os.pathsep.join(self.binPaths)}{os.pathsep}{os.getenv('PATH', '')}"
 
+        for k, val in self._configEnv.items():
+            # Preserve user-defined environment variables:
+            # manually set environment variable values take precedence over config file defaults.
+            if k in env:
+                continue
+
+            env[k] = val
+
```
fabiencastan, line 136: Is it still needed to modify the process env, as we now evaluate the value using the config in the evalValue?

cbentejac: Yes, because some environment variables are not paths to objects but need to be set at runtime. Example here: https://github.com/meshroomHub/mrSegmentation/blob/main/meshroom/config.json#L28

fabiencastan (approved):

PR #2800, opened 2025-08-12
[core/submitters] Freeze rez plugins dependencies versions in subprocess and farm submitter

Alxiice:
Force using specific versions on subprocess and simpleFarm jobs.

> [!NOTE]  
> Now, when we submit/compute a node, it uses the current env to select versions that are going to be used.
> However there are several case where we can have a mismatch.
> For example, if `pkg-v` and `pkg-v.1` are deployed and we have `pkg==v` in the current env, then `REZ_RESOLVE` will have resolved `pkg-v` (which is correct) however if we resolve `pkg-v` we won't have the version we want.

Therefore in this PR we propose to fix the packages whose version is taken from the current env and to fix it in the subprocess by using the `==` delimiter.
Also we address some other issues :
- package versions that are explicitely excluded (prefixed by `!`)
- packages with version delimiters different than `-`
- packages with `-` in the version string (because of the split on this symbol some issues could happen if a version is `pkg-version-0.1` for example, while it's a valid way to name packages)

### DIFF
--- a/meshroom/submitters/simpleFarmSubmitter.py
+++ b/meshroom/submitters/simpleFarmSubmitter.py
@@ -43,8 +43,9 @@ class SimpleFarmSubmitter(BaseSubmitter):
             for p in packages:
                 if p.startswith('~'):
                     continue
-                v = p.split('-')
-                self.reqPackages.append('-'.join([v[0], resolvedVersions[v[0]]]))
+                delimiter = "==" if "==" in p else "-"
+                v = p.split(delimiter)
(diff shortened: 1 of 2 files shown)


meshroom/submitters/simpleFarmSubmitter.py
```diff
@@ -38,13 +40,20 @@ def __init__(self, parent=None):
                 # logging.info('REZ: {}'.format(str(r)))
                 v = r.split('-')
```
fabiencastan: Force splitting only once, at most:
```suggestion
                v = r.split('-', 1)
```

fabiencastan (approved):

PR #2802, opened 2025-08-13
[ui] Add additional node information in the documentation tab

Alxiice:
## Description

This PR adds additional information on the Documentation tab of nodes. 
Here are examples :
```py
import os

__version__ = "1.0"
__author__ = "moduleauthor"
__license__ = "MIT"

class TestCLNode(desc.CommandLineNode):
    category = 'TestPlugin'
    commandLine = 'command'
    
    documentation = """# TestCLNode
## Node info
here are info on the node

## How to use it
Here are more info
1. First step
2. Second step
"""
```
image

Also we can override parameters at the node level :
- "author" and "version" can be directly overrided
- It is possible to put any additional information in a `__nodeInfo__` list
```py
class TestCLNode2(desc.CommandLineNode):
    """# TestCLNode

## Node info
here are info on the node

## How to use it
Here are more info
1. First step
2. Second step
"""

    category = 'TestPlugin'
    author = "Alice"
    version = "2.0.0"
    __nodeInfo__ = [
        ("model link", "https://huggingface.co/alicevision/great-model")
    ]
```
image

Finally I also added the possibility to read the documentation from the class `__doc__` parameterr

### DIFF
--- a/meshroom/ui/qml/GraphEditor/NodeDocumentation.qml
+++ b/meshroom/ui/qml/GraphEditor/NodeDocumentation.qml
@@ -104,9 +104,9 @@ FocusScope {
 
             TextEdit {
                 id: documentationText
-                anchors.top: nodeInfoListView.bottom
                 padding: 8
-                topPadding: 15
+                topPadding: 20
+                Layout.alignment: Qt.AlignTop | Qt.AlignLeft
                 Layout.preferredWidth: width
                 width: parent.parent.parent.width
                 textFormat: TextEdit.MarkdownText
--- a/meshroom/ui/qml/GraphEditor/NodeDocumentation.qml
+++ b/meshroom/ui/qml/GraphEditor/NodeDocumentation.qml
@@ -38,7 +38,6 @@ FocusScope {
                             id: nodeInfoKey
                             anchors.margins: 2
                             color: Qt.darker(activePalette.window, 1.1)
-                            //Layout.preferredWidth: sizeHandle.x
                             Layout.preferredWidth: nodeDocColumnLayout.keyColumnWidth
                             Layout.minimumWidth: 0.2 * parent.width
                             Layout.maximumWidth: 0.8 * parent.width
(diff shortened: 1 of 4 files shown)


fabiencastan:
I would suggest a simplication like this (with a fallback on the module):
```
class TestCLNode2(desc.CommandLineNode):
    """# TestCLNode

## Node info
here are info on the node

## How to use it
Here are more info
1. First step
2. Second step
"""

    category = 'TestPlugin'
    commandLine = 'python ' + exe_path
    author = "Alice"
    version = "2.0.0"

```

meshroom/ui/qml/GraphEditor/NodeDocumentation.qml
```diff
@@ -22,17 +22,100 @@ FocusScope {
...
+                            id: nodeInfoKey
+                            anchors.margins: 2
+                            color: Qt.darker(activePalette.window, 1.1)
+                            Layout.preferredWidth: nodeDocColumnLayout.keyColumnWidth
+                            Layout.minimumWidth: 0.2 * parent.width
+                            Layout.maximumWidth: 0.8 * parent.width
+                            Layout.fillWidth: false
+                            Layout.fillHeight: true
+                            Label {
+                                text: modelData.key
```
cbentejac, line 47: The labels for the keys should be capitalized. Generally speaking, we try to avoid lowercase fields. So we should have "Author", "License", and so on, instead of using directly the keys.

cbentejac (changes requested):
The documentation keys should be capitalized in the display but apart from that, everything is working well and ready to be merged!

meshroom/ui/qml/GraphEditor/NodeDocumentation.qml
Alxiice: I added this with a font capitalization in the QML. Is it okay or do you think it would be better to do it in the python side ?

cbentejac: All good for me :)

cbentejac (approved):

PR #2836, opened 2025-08-18
[ui] Eval expression directly in specific TextFields

Alxiice:
- Adding `ExpressionTextField` component that is evaluating and replacing the text field when entered math expressions
expre1
- Error cases are handled with NaN values and a visual indicator
expre2
- Possibility to specify if the field is int or float (int will automatically cast to an integer)

For now this is used :
- On slider attributes (Used for IntParam and FloatParam)
- On the `HdrImageToolbar` component and on the gamma and gain settings

Some additional examples :
Peek 2025-09-09 13-03


meshroom/ui/qml/Controls/ExpressionTextField.qml
```diff
@@ -0,0 +1,52 @@
...
+        radius: 4
+        border.color: "red"
+        color: "transparent"
+        visible: root.hasExprError || root.text == "NaN"
+        z: 1
+    }
+
+    function evalExpression() {
+        try {
+            if (root.text == "NaN") {
```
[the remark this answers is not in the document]
Alxiice: I put a string for display reason mainly, if I use NaN then it displays the string as "nan" for some reason. Also in the end it doesn't really change anything because we control what we display on the field

meshroom/ui/qml/Viewer/HdrImageToolbar.qml
```diff
@@ -154,7 +153,7 @@ FloatingPane {
                     gammaCtrl.value = gammaDefaultValue;
```
cbentejac, line 167: As a result of the changes, this doesn't fully work as expected anymore: the value is correctly updated, but the label is not. 

This is also true for the gain's `ExpressionTextField`.

meshroom/ui/qml/Controls/mathEvaluator.js
cbentejac, line 1: I think this would be better if it was placed in qml/Utils/ instead of qml/Controls/.

meshroom/ui/qml/Controls/ExpressionTextField.qml
cbentejac, line 1: It's not a blocker for the PR, but with this component, we lose the text field's history when an expression is used (if we type in values without expressions before or after, we can go back to them with Ctrl-Z, but the ones that result from the evaluation of expressions are just lost). I don't know if it's something that's easy to get back but if so, it'd be really nice (otherwise, it's not a big deal).

meshroom/ui/qml/Controls/ExpressionTextField.qml
```diff
@@ -0,0 +1,52 @@
...
+
+    // Overlay for error state
+    Rectangle {
+        anchors.fill: parent
+        radius: 4
+        border.color: "red"
+        color: "transparent"
+        visible: root.hasExprError || root.text == "NaN"
+        z: 1
+    }
```
cbentejac: There's a corner case with this that's not yet handled: if you input a value that triggers the error state, and then right-click on the attribute and reset its default value (which is then valid), the "visible" state is not re-evaluated and the text/red border are not updated.

The only way to get back to something normal is to select another node and then get back on this one.

PR #2845, opened 2025-08-21
[core] Setup logger to be able to use logging in plugins

Alxiice:
## Description

- Adding `prepareLogger` and `restoreLogger` on the node, that can be used to setup and restore the root logger
- Update the `LogManager` to handle more than only the chunks
- Add trace log level to logging with an inferior level than debug (debug is 10 by default so trace would be 5 with this PR)
- Add new internal input "logLevel" to setup default logging level for the node

## Now here's how we can use it

1. Using `chunk.logger` (previous way)
```py
class TestNode(desc.Node):
        chunk.logManager.start(chunk.node.verboseLevel.value)
        chunk.logger.info   (f"Message with info    level ({chunk.logger})")
        chunk.logger.warning(f"Message with warning level ({chunk.logger})")
        chunk.logger.error  (f"Message with error   level ({chunk.logger})")
        chunk.logManager.end()
```
image

2. Using `logging`
```py
import logging
logger = logging.getLogger("TestNode2")

def print_info():
    logger.info   (f"Message with info    level ({logger})")
    logger.warning(f"Message with warning level ({logger})")
    logger.error  (f"Message with error   level ({logger})")

class TestNode2(desc.Node):
        logger.setLevel(chunk.node.verboseLevel.value.upper())
        print_info()
```
image

### DIFF
--- a/meshroom/core/desc/node.py
+++ b/meshroom/core/desc/node.py
@@ -13,6 +13,7 @@ from .attribute import StringParam, ColorParam, ChoiceParam
 
 import meshroom
 from meshroom.core import cgroup
+from meshroom.core.utils import VERBOSE_LEVEL
 
 _MESHROOM_ROOT = Path(meshroom.__file__).parent.parent.as_posix()
 _MESHROOM_COMPUTE = (Path(_MESHROOM_ROOT) / "bin" / "meshroom_compute").as_posix()
@@ -68,11 +69,11 @@ class BaseNode(object):
             invalidate=False,
         ),
         ChoiceParam(
-            name='nodeDefaultLogLevel',
-            label='Default logging level',
-            description='''Default logging level for the node (critical, error, warning, info, debug).''',
-            value='info',
-            values=['critical', 'error', 'warning', 'info', 'debug', 'trace'],
+            name="nodeDefaultLogLevel",
+            label="Default Logging Level",
+            description="Default logging level for the node (critical, error, warning, info, debug).",
+            value="info",
+            values=VERBOSE_LEVEL,
             invalidate=False,
         ),
         ColorParam(
--- a/meshroom/core/desc/node.py
+++ b/meshroom/core/desc/node.py
@@ -9,7 +9,7 @@ import shutil
 import sys
 
 from .computation import Level, StaticNodeSize
-from .attribute import StringParam, ColorParam
+from .attribute import StringParam, ColorParam, ChoiceParam
 
 import meshroom
 from meshroom.core import cgroup
@@ -65,6 +65,14 @@ class BaseNode(object):
             description="Customize the default label (to replace the technical name of the node "
                         "instance).",
             value="",
+            invalidate=False,
+        ),
+        ChoiceParam(
+            name='nodeDefaultLogLevel',
+            label='Default logging level',
+            description='''Default logging level for the node (critical, error, warning, info, debug).''',
+            value='info',
+            values=['critical', 'error', 'warning', 'info', 'debug', 'trace'],
             invalidate=False,
         ),
         ColorParam(
--- a/meshroom/core/node.py
+++ b/meshroom/core/node.py
@@ -14,14 +14,14 @@ import types
 import uuid
 from collections import namedtuple, OrderedDict
 from enum import Enum, auto
-from typing import Callable, Optional
-
+from typing import Callable, Optional, List
 
 import meshroom
 from meshroom.common import Signal, Variant, Property, BaseObject, Slot, ListModel, DictModel
 from meshroom.core import desc, plugins, stats, hashValue, nodeVersion, Version, MrNodeType
 from meshroom.core.attribute import attributeFactory, ListAttribute, GroupAttribute, Attribute
 from meshroom.core.exception import NodeUpgradeError, UnknownNodeTypeError
+from meshroom.core.mtyping import PathLike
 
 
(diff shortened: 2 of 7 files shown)


fabiencastan:
I would like to simply it a bit more, like this:

```py
import logging

def print_info():
    logging.info   (f"Message with info    level ({logger})")
    logging.warning(f"Message with warning level ({logger})")
    logging.error  (f"Message with error   level ({logger})")

class TestNode2(desc.Node):
        logging.getLogger().setLevel(chunk.node.verboseLevel.value.upper())
        print_info()
```

So if we call any 3rd party python code, the logging behavior will be correct.

And maybe in another PR, move the verbose param as a default param, so we can configure it by default.

Alxiice:
@fabiencastan 

> So if we call any 3rd party python code, the logging behavior will be correct.

Actually I tested and with the code you provided it works already

> And maybe in another PR, move the verbose param as a default param, so we can configure it by default.

You mean as an [internalInput](https://github.com/alicevision/Meshroom/blob/develop/meshroom/core/desc/node.py#L42) ?

fabiencastan:
> You mean as an internalInput ?

Yes

meshroom/core/desc/node.py
```diff
@@ -71,6 +71,14 @@ class BaseNode(object):
             value="",
             invalidate=False,
         ),
+        ChoiceParam(
+            name='nodeDefaultLogLevel',
+            label='Default logging level',
```
cbentejac:
```suggestion
            label='Default Logging Level',
```
For labels, we're trying to always use PascalCase.

meshroom/core/desc/node.py
```diff
@@ -71,6 +71,14 @@ class BaseNode(object):
...
+            description='''Default logging level for the node (critical, error, warning, info, debug).''',
+            value='info',
+            values=['critical', 'error', 'warning', 'info', 'debug', 'trace'],
```
cbentejac: There's a constant for this as it is reused in pretty much every node:
```python
from meshroom.core.utils import VERBOSE_LEVEL
...
ChoiceParam(
    ...,
    values=VERBOSE_LEVEL,
),
```

meshroom/core/desc/node.py
```diff
@@ -71,6 +71,14 @@ class BaseNode(object):
             value="",
             invalidate=False,
         ),
+        ChoiceParam(
```
cbentejac, line 71: Generally, we try to harmonize the use of single and double quotes across the files. In this case, most of the file is using double quotes, so it'd be better to use them here.

meshroom/core/node.py
```diff
@@ -220,30 +220,42 @@ def fromDict(self, d):
...
+        self._previousHandlers = []
+        self._previousLevel = 0
+    
     class Formatter(logging.Formatter):
         def format(self, record):
             # Make level name lower case
             record.levelname = record.levelname.lower()
             return logging.Formatter.format(self, record)
-
+
```
cbentejac: Trailing whitespaces

meshroom/core/node.py
```diff
@@ -220,30 +220,42 @@ def fromDict(self, d):
...
     dateTimeFormatting = '%H:%M:%S'
 
-    def __init__(self, chunk):
-        self.chunk = chunk
-        self.logger = logging.getLogger(chunk.node.getName())
-
+    def __init__(self, logger, logFile):
+        self.logger = logger
+        self.logFile = logFile
+        self._previousHandlers = []
```
cbentejac: It would be nice to add some typing here if possible.

meshroom/core/nodeFactory.py
```diff
@@ -174,13 +174,15 @@ def _checkAttributesCompatibility(
 
     def _createNode(self) -> Node:
         logging.info(f"Creating node '{self.name}'")
-        # TODO: user inputs/outputs may conflicts with internal names (like position, uid)
+        # TODO: user inputs/outputs may conflicts with internal names (like logLevel, position, uid)
+        # The line below can cause UI issues but at least prevent crashes
+        internalInputs = {k: v for k, v in self.internalInputs.items() if k not in self.inputs.keys()}
```
cbentejac, line 179: Is this necessary, since the internal input for the logger doesn't have (and isn't likely to ever have) the same name as an input parameter?

cbentejac (changes requested):
Everything works as expected, the requested changes are mostly cosmetic.

tests/test_compute.py
```diff
@@ -0,0 +1,120 @@
+# coding:utf-8
+
+"""
+In this test we test the code that is usually launched directly from the meshroom_compute script
+
+TODO : Directly test by launching the executable
+- We can get the path with `desc.node._MESHROOM_COMPUTE_EXE`
+- However this is not implemented yet because it requires to create a plugin that meshroom could discover
+  and I didn't want to create a plugin just for what I'm doing here right now (which is testing the logger)
```
cbentejac: This is unnecessary to keep, the rest of the TODO is clear enough as is

meshroom/core/nodeFactory.py
Alxiice: It shouldn't be necessary, but I still think this is better to have this protection, because having twice the same keyword arg in the call of `Node(...)` will always raise an error.

cbentejac (approved):

PR #2847, opened 2025-08-21
[desc] Add node process function to use on single-chunk nodes

Alxiice:
> [!WARNING]
> It would be better to merge this PR after #2845 because the logger is not available in `process` for now (as it's provided by `chunk`). This PR mentioned fixes this by setuping `logging` so that we can use it directly

## Description

- Add a `process` function
- If there is no parallelization, then if `processChunk` has not been overrided, then `process` will be called

## Examples
Because we don't have access to the chunk we can't use `chunk.logger`
```py
import logging
logger = logging.getLogger("TestNode2")

def print_info():
    logger.info(f"Message with info    level ({logger})")
    logger.warning(f"Message with warning level ({logger})")
    logger.error(f"Message with error   level ({logger})")

class TestNode3(desc.Node):    
    inputs = [
        desc.ChoiceParam(
            name='verboseLevel',
            label='Verbose Level',
            description='''verbosity level (fatal, error, warning, info, debug, trace).''',
            value='info',
            values=['fatal', 'error', 'warning', 'info', 'debug', 'trace'],
            exclusive=True,
        ),
    ]

    outputs = [
        desc.File(
            name='outputFolder',
            label='outputFolder',
            description='outputFolder',
            value="{nodeCacheFolder}",
        ),
    ]

    def process(self, node):
        logger.setLevel(node.verboseLevel.value.upper())
        logger.info(f"> verboseLevel : {node.verboseLevel.value}")
        logger.info(f"> outputFolder : {node.outputFolder.value}")
        print_info()

```
image

However, after #2845 this won't be an issue because we could use logging normally :
image

### DIFF
--- a/tests/test_compute.py
+++ b/tests/test_compute.py
@@ -42,76 +42,114 @@ def executeChunks(node, tmpPath, size):
     return logFiles
 
 
+_INPUTS = [
+    desc.IntParam(
+        name="input",
+        label="Input",
+        description="input",
+        value=0,
+    ),
+]
+_OUTPUTS = [
+    desc.IntParam(
+        name="output",
+        label="Output",
+        description="Output",
+        value=None,
+    ),
+]
+
 class TestNodeA(desc.BaseNode):
+    """
+    Test process with chunks
+    """
     __test__ = False
-    
+    _size = 2
     size = desc.StaticNodeSize(2)
     parallelization = desc.Parallelization(blockSize=1)
-
-    inputs = [
-        desc.IntParam(
-            name="input",
-            label="Input",
-            description="input",
-            value=0,
-        ),
-    ]
-    outputs = [
-        desc.IntParam(
-            name="output",
-            label="Output",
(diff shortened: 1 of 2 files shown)


tests/test_compute.py
```diff
@@ -0,0 +1,127 @@
...
+            chunk = node.chunks[chunkIndex]
+            chunk.process(True, True)
+        else:
+            node.process(True, True)
+        node.postprocess()
+    return logFiles
+
+
+class TestNodeA(desc.BaseNode):
+    __test__ = False
```
cbentejac, line 66: FYI: without the `__test__ = False` line, pytest tries to load the `TestNodeA` and `TestNodeB` classes because of the "Test" in their name, and fails as they have `__init__` methods.

An alternative to using `__test__ = False` would be to rename the classes to not use the word "Test".

cbentejac (approved):
Approved but waiting for #2845 to be merged first.

PR #2850, opened 2025-08-21
[ui] Keep projects in recent list even if there is a load error

fabiencastan:


cbentejac:
Wouldn't it be better to distinguish the case where the load error occurs because the file does not exist anymore (in which case we have no reason to keep it in the list at all) from the case where there's a "real" loading error (in which case it does make sense to keep the project in the list)?

fabiencastan:
Yes, you're right!

cbentejac (approved):

cbentejac (changes requested):
Please handle the case where the file is missing separately :)

Alxiice:
So I did the update this way by disabling items
> [!NOTE]
> I think this is better because maybe users want to be aware that one of their projects is missing, because it could be due to folder renaming / disk that is unplugged
image
On the home page we can still remove the entry :
image

cbentejac (approved):

PR #2867, opened 2025-08-29
[bin] `meshroom_batch`: Add support for setting multiple `Publish` nodes' output folders 

### LINKED ISSUE #1141: [request] meshroom_photogrammetry: deal with multiple Publish nodes

cbentejac:
## Description

This PR adds the support for multiple `Publish` nodes with different output folders in `meshroom_batch`. Prior to it, a graph could have several `Publish` nodes, but only one path could be provided to the `--output` option, meaning all the nodes were writing their outputs in the same folder.

With this PR, specific `Publish` nodes can have their output folder set using the syntax `NAME=FOLDER` (e.g `-o Publish_1=/path/to/output/folder`). Different cases are covered:
1. All the `Publish` nodes are specified individually (`-o Publish_1=/path/to/folder1 Publish_2=/path/to/folder2`): they will be set with their provided output folders.
2. None of the `Publish` node has a specified path (`-o /path/to/output/folder`): all the `Publish` nodes will be set with the same provided value (this was the behaviour up until this PR).
3. Some `Publish` nodes are specified individually, but some are not (`-o Publish_1=/path/to/folder1`): all the `Publish` nodes that do not have a specified output folder will use the first specified node's (here, `/path/to/folder1`).
4. Some `Publish` nodes are specified individually, and a path that is not associated to any node is also provided (`-o Publish_1=/path/to/folder1 /path/to/output`): the specified nodes will be set with their own output folders, all the other ones will use the remaining path.
5. None of the `Publish` nodes have a specified path, but several paths are provided (`-o /path/to/folder1 /path/to/folder2`): all the `Publish` nodes will be set with the last path that was provided (here `/path/to/folder2`).

Additionally, the `Publish` node is slightly modified to replace `os.mkdir` with `os.makedirs` when the provided output folder does not exist and needs to be created. With that change, if the intermediate folders also do not exist, they will be created instead of an error being raised.

### DIFF
--- a/bin/meshroom_batch
+++ b/bin/meshroom_batch
@@ -62,8 +62,11 @@ general_group.add_argument(
     )
 
 general_group.add_argument(
-    '-o', '--output', metavar='FOLDER', type=str, required=False,
+    '-o', '--output', metavar='FOLDER PUBLISH_INSTANCE=FOLDER',
+    type=str, required=False, nargs='*',
     help='Output folder where results should be copied to. '
+         'If the output folder is provided without specifiying the instance of a Publish node, '
+         'all the Publish nodes in the scene will be set with the same ouput folder value. '
          'If not set, results will have to be retrieved directly from the cache folder.')
 
 general_group.add_argument(
@@ -129,7 +132,6 @@ advanced_group.add_argument(
     help='Execute job with a specific submitter.')
 
 args = parser.parse_args()
-
 
 logStringToPython = {
     'fatal': logging.FATAL,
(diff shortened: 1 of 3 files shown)


bin/meshroom_batch
```diff
@@ -221,15 +223,48 @@ with meshroom.core.graph.GraphModification(graph):
...
+                globalPublishPath = p
+                continue
+
+            node, value = result.groups()
+            for i, n in enumerate(publishNodes):  # Find the correct Publish node in the list
+                if n.name == node:  # If found, set the value, and remove it from the list
+                    n.output.value = value
+                    publishNodes.pop(i)
+                    if globalPublishPath == "":  # Fallback in case some nodes would have no path
+                        globalPublishPath = value
```
fabiencastan, line 264: I would remove these 2 lines for clarity/simplicity of the rules.

cbentejac: If some nodes don't have a specified path and there's no "global path", should we then throw an error?

Alxiice (approved):

PR #2868, opened 2025-08-29
[nodes] Rename the `Publish` node to `CopyFiles`

cbentejac:
## Description

This PR renames the `Publish` node, whose name may be ambiguous depending on the context, to `CopyFiles`.

This should preferably be merged after https://github.com/alicevision/Meshroom/pull/2867 as it contains changes that are based on the `Publish` node.

> [!NOTE]
> The following PRs should be merged along with this one:
> - https://github.com/alicevision/AliceVision/pull/1962
> - https://github.com/meshroomHub/mrSegmentation/pull/29

### DIFF
--- a/bin/meshroom_batch
+++ b/bin/meshroom_batch
@@ -57,17 +57,17 @@ general_group.add_argument(
     default=os.environ.get('MESHROOM_DEFAULT_PIPELINE', 'photogrammetry'),
     help='Template pipeline among those listed or a Meshroom file containing a custom pipeline '
          'to run on input images:\n' +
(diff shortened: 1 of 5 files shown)


Alxiice (approved):
I'm okay with the PR I'm just wondering if that would be better to keep the publish node with a depreciation warning for now but that's probably not really necessary

fabiencastan (changes requested):
I would prefer to name it "CopyFiles".
We already have many Export nodes to export in a different file formats.

fabiencastan (approved):

PR #2874, opened 2025-09-03
[submitter] Add tractor API

Alxiice:
# Description

> [!NOTE]
> The code is ready for review but I'm still adding test and documentation

- Rename the `submit` on nodes whose name was not super clear
- Tweak the plugins registration to make sure we can register submitters from external plugins. (There was an issue previously because the package name was set to "submitters" and therefore it was creating conflicts with the submitters in `meshroom/submitters`)
- Add `tractorSubmitter.py` with the tractor submitter
- UI changes : Add possibility to select the submitter
image

# Information for review

- I stripped a bit some tasks :
  - usually we create a "node" task, and tasks for every chunks that have the "node" task as parent
  - this task is useless when we have only one chunk so I stripped them in this case because I thought it helps create leaner graphs.
    image
  - However I can understand why we would want them.
  - Let me know if it was better before

> [!IMPORTANT]
> Another important modification was done on the task ordering, as you can see with the simpleFarm currently when we select multiple tasks and submit, tasks are duplicated
> Peek 2025-09-05 10-48
> _With the current submitter, CameraInit is submitted on its own AND a second time as a child of SleepNode_
> _Using the new API we make sure to only add tasks once to the job_
> 
> **However** it's clear that it's an error that happens somewhere else on the code because we should have a clean graph that is submitted in the first place, and **I will create another PR to resolve this issue**

# How to test

Launch a job with SimpleFarm submitter, then with Tractor submitter, compare both jobs created on tractor


natowi:
@Alxiice would you be so kind to document how to setup Meshroom with tractor? https://github.com/alicevision/Meshroom/wiki/Renderfarm-submitters Thank you!

Alxiice:
> @Alxiice would you be so kind to document how to setup Meshroom with tractor? https://github.com/alicevision/Meshroom/wiki/Renderfarm-submitters Thank you!

hello @natowi it's a bit difficult to explain exactly how to test tractor because it relies on code from Pixar that I cannot really share here. However if that helps I added some documentation on the wiki and explained a bit more how the submitter works and how to create one

meshroom/ui/qml/Application.qml
```diff
@@ -928,6 +928,22 @@ Page {
                     onTriggered: _reconstruction.stopExecution()
                     enabled: _reconstruction ? _reconstruction.computingLocally : false
                 }
+                MenuSeparator {}
+                Menu {
+                    id: submitterSelectionMenu
+                    title: "Submitter selection"
```
cbentejac:
```suggestion
                    title: "Submitter Selection"
```

meshroom/ui/qml/Application.qml
```diff
@@ -928,6 +928,22 @@ Page {
...
+                    enabled: submitterItems.model !== undefined && submitterItems.model.length > 0
+                    Repeater {
+                        id: submitterItems
+                        model: MeshroomApp.submittersListModel
+                        RadioButton {
+                            text: modelData["name"]
+                            checked: modelData["isDefault"]
+                            // checked: MeshroomApp.submittersListModel[0]["name"] == modelData["name"]
```
cbentejac:
```suggestion
```

meshroom/ui/app.py
```diff
@@ -698,7 +700,28 @@ def _getEnvironmentVariableValue(self, key: str, defaultValue: bool) -> bool:
...
+        Model provides :
+            name : the name of the submitter
+            isDefault : True if this is the current submitter
+        """
+        submittersList = []
+        for s in meshroom.core.submitters:
+            submitterName = s.name if isinstance(s, BaseSubmitter) else s
+            submittersList.append({
+                "name": submitterName,
+                "isDefault": True if submitterName == self._defaultSubmitterName else False
```
cbentejac: We need to add a protection in case there is no default submitter, otherwise we can end up with no submitter selected, prompting (as it should) the following error:
```
RuntimeError: [SUBMITTING] Unknown Submitter:
Unknown Submitter called ''. Available submitters are: 'dict_keys(['SimpleFarm', 'Tractor'])'.
```

meshroom/ui/app.py
```diff
@@ -265,6 +266,7 @@ def __init__(self, inputArgs):
 
         # instantiate Reconstruction object
         self._undoStack = commands.UndoStack(self)
+        self._defaultSubmitterName = os.environ.get('MESHROOM_DEFAULT_SUBMITTER', '')
```
cbentejac, line 270: Something needs to be done in case we fall back on the default (empty) value if some submitters are available, like taking the first one in the list of loaded submitters.

cbentejac (changes requested):
The `tractorConfig.json` is missing and needs to be added, otherwise the submitter for Tractor cannot be loaded.

Submissions work fine, but we need to handle the case where there's no default submitter through the `MESHROOM_DEFAULT_SUBMITTER` environment variable if there are submitters that have been successfully loaded.

Alxiice:
> Submissions work fine, but we need to handle the case where there's no default submitter through the `MESHROOM_DEFAULT_SUBMITTER` environment variable if there are submitters that have been successfully loaded.

I replaced `==1` by `>=1` here 
https://github.com/alicevision/Meshroom/blob/dev/tractor_api/meshroom/core/taskManager.py#L391
This should work

meshroom/ui/app.py
Alxiice: Okay this should be okay

cbentejac (approved):

PR #2881, opened 2025-09-09
Perform path resolution function in Meshroom core

demoulinv:
## Description
This PR implements the path resolution function within meshroom core in the new file fileUtils.py. That resolution was formerly implemented in meshroom ui components. The original method still exists but has been cleaned up and calls the new ones.

The path resolution function replace all the fields <FIELD> in a template name by their corresponding value extracted whether from an input filepath or a viewpoint.

A dictionary is built from the input filepath or the viewpoint. The fields to be replaced in the template name are supposed to be found amongst the keys of that dictionary. In case a key does not exist the corresponding input field is discarded to build the resolved path. Such a case can occur for example if the field <FRAMEID> is used in the template whereas the input filepath does not contain any value corresponding to a frame number.

The way the replacement is done has also been optimized using regex sub function.

### DIFF
--- a/meshroom/core/fileUtils.py
+++ b/meshroom/core/fileUtils.py
@@ -1,25 +1,30 @@
 import os
 import re
 
+pattern = r"(?P<FILESTEM_PREFIX>.*?)(?P<FRAMEID_STR>[-._]\d+)?(?P<EXTENSION>\.\w{3,4})"
+compiled_pattern = re.compile(pattern)
+compiled_frameId = re.compile(r"(\D+)?(?P<FRAMEID>\d+$)")
+
 def getFileElements(inputFilePath: str):
 
     filename = os.path.basename(inputFilePath)
-    pattern = r"(?P<FILESTEM>.*?)(?P<FRAME_ID>[-._]\d+)?(?P<EXTENSION>\.\w{3,4})"
-    match = re.search(pattern, filename)
-    frameId = match.group("FRAME_ID")
+    match = compiled_pattern.search(filename)
+    frameId_str = match.group("FRAMEID_STR")
     
     fileElements = {}
     if match:
         fileElements = {
             "<PATH>": inputFilePath,
             "<FILENAME>": filename,
-            "<FILESTEM>": match.group("FILESTEM"),
-            "<FILESTEM_PREFIX>": match.group("FILESTEM"),
+            "<FILESTEM>": match.group("FILESTEM_PREFIX"),
+            "<FILESTEM_PREFIX>": match.group("FILESTEM_PREFIX"),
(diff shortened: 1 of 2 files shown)


meshroom/core/fileUtils.py
```diff
@@ -0,0 +1,59 @@
+import os
+import re
+
+def getFileElements(inputFilePath: str):
+
+    filename = os.path.basename(inputFilePath)
+    pattern = r"(?P<FILESTEM>.*?)(?P<FRAME_ID>[-._]\d+)?(?P<EXTENSION>\.\w{3,4})"
```
fabiencastan:
```suggestion
    pattern = r"(?P<FILESTEM_PREFIX>.*?)(?P<FRAMEID_STR>[-._]\d+)?(?P<EXTENSION>\.\w{3,4})"
```

fabiencastan: re.compile() this static expression out of the function to do it only once

meshroom/core/fileUtils.py
```diff
@@ -0,0 +1,59 @@
...
+
+
+def resolvePath(input, outputTemplate: str) -> str:
+
+    if isinstance(input, str):
+        replacements = getFileElements(input)
+    else:
+        replacements = getViewElements(input)
+
+    resolved = replacePatterns(outputTemplate, r"<\w*>", replacements)
```
fabiencastan: re.compile

meshroom/core/fileUtils.py
```diff
@@ -0,0 +1,59 @@
...
+        fileElements = {
+            "<PATH>": inputFilePath,
+            "<FILENAME>": filename,
+            "<FILESTEM>": match.group("FILESTEM"),
+            "<FILESTEM_PREFIX>": match.group("FILESTEM"),
+            "<EXTENSION>": match.group("EXTENSION"),
+        }
+    if frameId is not None:
+        fileElements["<FRAMEID>"] = frameId
+        fileElements["<FILESTEM>"] += frameId
```
fabiencastan:
```suggestion
        fileElements = {
            "<PATH>": inputFilePath,
            "<FILENAME>": filename,
            "<FILESTEM>": match.group("FILESTEM_PREFIX"),
            "<FILESTEM_PREFIX>": match.group("FILESTEM_PREFIX"),
            "<EXTENSION>": match.group("EXTENSION"),
        }
    if frameId is not None:
        fileElements["<FRAMEIDSTR>"] = frameId
        # TODO: Extract FRAMEID with only the number
        fileElements["<FILESTEM>"] += frameId
```

fabiencastan (approved):

cbentejac (approved):

PR #2884, opened 2025-09-11
[ui] Add NodeActions to launch computation

Alxiice:
# Description

Add `NodeActions`
- Creates a header on top of the nodes with buttons to launch / force relaunch / submit the node
- The NodeActions follows the node but keeps a uniform scale

_Demonstration on how the items process work_
Peek 2025-09-30 15-37

> [!NOTE]
> Small remark : `selectNodeByIndex` triggers `onSelectedNodeDelegateChanged` twice...

### DIFF
--- a/meshroom/ui/graph.py
+++ b/meshroom/ui/graph.py
@@ -1169,8 +1169,29 @@ class UIGraph(QObject):
             logging.warning("Content is not a valid graph data.")
             return []
         return result
-
-
+    
+    @Slot(Node, result=bool)
+    def canComputeNode(self, node: Node) -> bool:
+        """ Check if the node can be computed """
+        if node.isCompatibilityNode or not node.isComputableType:
+            return False
+        if node.isComputed:
+            return True
+        if self._graph.canComputeTopologically(node) and self._graph.canSubmitOrCompute(node) % 2 == 1:
+            return True
+        return False
+    
+    @Slot(Node, result=bool)
+    def canSubmitNode(self, node: Node) -> bool:
+        """ Check if the node can be submitted """
+        if node.isCompatibilityNode or not node.isComputableType:
+            return False
+        if node.isComputed:
+            return True
+        if self._graph.canComputeTopologically(node) and self._graph.canSubmitOrCompute(node)> 1:
+            return True
+        return False
+    
     undoStack = Property(QObject, lambda self: self._undoStack, constant=True)
     graphChanged = Signal()
     graph = Property(Graph, lambda self: self._graph, notify=graphChanged)
--- a/meshroom/ui/graph.py
+++ b/meshroom/ui/graph.py
@@ -1173,7 +1173,7 @@ class UIGraph(QObject):
     @Slot(Node, result=bool)
     def canComputeNode(self, node: Node) -> bool:
         """ Check if the node can be computed """
-        if node.isCompatibilityNode or not node.isComputableType:
+        if node.isCompatibilityNode or not node.isComputableType or node.getLocked():
             return False
         if node.isComputed:
             return True
@@ -1184,7 +1184,7 @@ class UIGraph(QObject):
     @Slot(Node, result=bool)
     def canSubmitNode(self, node: Node) -> bool:
         """ Check if the node can be submitted """
-        if node.isCompatibilityNode or not node.isComputableType:
+        if node.isCompatibilityNode or not node.isComputableType or node.getLocked():
             return False
         if node.isComputed:
             return True
--- a/meshroom/ui/qml/Controls/NodeActions.qml
+++ b/meshroom/ui/qml/Controls/NodeActions.qml
@@ -74,11 +74,9 @@ Item {
         // Update position
         function updatePosition() {
             if (!selectedNodeDelegate || !draggable) return
-
             // Calculate node position in screen coordinates
             const nodeScreenX = selectedNodeDelegate.x * draggable.scale + draggable.x
             const nodeScreenY = selectedNodeDelegate.y * draggable.scale + draggable.y
-
             // Position header above the node (fixed offset in screen pixels)
             x = nodeScreenX + (selectedNodeDelegate.width * draggable.scale - width) / 2
             y = nodeScreenY - height - headerOffset
@@ -105,7 +103,11 @@ Item {
         // 
 
         property bool nodeIsLocked: false
-        property bool submittedExternally: false
+        property bool canComputeNode: false
+        property bool canStopNode: false
+        property bool canSubmitNode: false
+        property bool nodeSubmitted: false
+
         property int computeButtonState: NodeActions.ButtonState.LAUNCHABLE
         property string computeButtonIcon: {
             switch (computeButtonState) {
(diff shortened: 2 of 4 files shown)


fabiencastan:
We should have either 'Compute' or 'Recompute', but not both at the same time.

meshroom/ui/qml/GraphEditor/GraphEditor.qml
```diff
@@ -1075,6 +1075,41 @@ Item {
...
+        onSubmitRequest: function(node) {
+            if (node.isComputed) uigraph.clearSelectedNodesData();
+            root.submitRequest([node])
+        }
+        
+        // TODO : If we want this, we should add the possibility to stop job on farm first
+        // onReSubmitRequest: function(node) {
+        //     uigraph.clearSelectedNodesData();
+        //     root.submitRequest([node])
+        // }
```
fabiencastan: It would be better to avoid including that in this PR. It should be done when we will extend the possible actions on the submitters.

meshroom/ui/qml/Controls/NodeActions.qml
```diff
@@ -0,0 +1,263 @@
...
+                    border.color: reSubmitButton.hovered ? activePalette.highlight : Qt.darker(activePalette.window, 1.3)
+                    border.width: 1
+                    radius: 3
+                }
+                onClicked: {
+                    if (actionHeader.selectedNode) {
+                        root.reSubmitRequest(actionHeader.selectedNode)
+                    }
+                }
+            }
```
fabiencastan: To remove for now.

meshroom/ui/qml/Controls/NodeActions.qml
```diff
@@ -0,0 +1,263 @@
...
+        }
+        
+        // Set initial position
+        onSelectedNodeDelegateChanged: updatePosition()
+
+        // Listen to updates to status
+        Connections {
+            target: actionHeader.selectedNode
+            function onGlobalStatusChanged()   { actionHeader.currentStatus = target.globalStatus }
+            function onGlobalExecModeChanged() { actionHeader.currentExecMode = target.globalExecMode }
```
fabiencastan: They are properties, so it should be updated automatically without this.

meshroom/ui/qml/Controls/NodeActions.qml
```diff
@@ -0,0 +1,263 @@
...
+            function onIsComputedChanged()     { actionHeader.nodeCanBeStopped = target.canBeStopped() }
```
fabiencastan: canBeStopped should become a property if we need it like this. Currently it's a slot which does not allow event propagation.

meshroom/ui/qml/Controls/NodeActions.qml
```diff
@@ -0,0 +1,263 @@
...
+        property string currentExecMode: selectedNode ? selectedNode.globalExecMode : "NONE"
+        property string currentStatus: selectedNode ? selectedNode.globalStatus : "NONE"
+        property bool nodeCanBeStopped: selectedNode ? selectedNode.canBeStopped() : false
+        property bool nodeIsExternal: selectedNode ? selectedNode.isExternal : false
+        property bool nodeLocked: selectedNode ? selectedNode.locked : false
+
+        // Derived values
+        readonly property bool runningExternally: currentExecMode === "EXTERN"
+                                                  && ["SUBMITTED", "RUNNING"].includes(currentStatus)
+        readonly property bool stoppable: selectedNode && currentStatus === "RUNNING" && nodeCanBeStopped
```
fabiencastan: nodeCanBeStopped already check that selectedNode is valid and status is running. So this property seems useless.

meshroom/ui/qml/Controls/NodeActions.qml
```diff
@@ -0,0 +1,263 @@
...
+            }
+            ignoreUnknownSignals: true
+        }
+
+        Row {
+            id: actionItemsRow
+            anchors.centerIn: parent
+            spacing: 2
+            
+            // Compute button
```
fabiencastan, line 202: I would prefer a single button for the current compute/recompute.
This button would have 3 steps:
- Compute (when not computed)
- Stop (when computing)
- Delete Data (when computed)

When computed, if you want to recompute, you just have to click twice (Delete & Compute).

I would use an orange background on hover for Stop and a red one for Delete.

Alxiice: I did the modification the only drawback is that we can't delete data when the node is not fully computed (because in this case the first nodeaction is just restarting to compute the missing chunks) so if we have multiple chunks we can't recompute from the first chunk without manually deleting the node data first

meshroom/ui/qml/Controls/NodeActions.qml
```diff
@@ -0,0 +1,297 @@
...
+                        case NodeActions.ButtonState.LAUNCHABLE: 
+                            root.computeRequest(actionHeader.selectedNode)
+                            break
+                    }
+                }
+            }
+
+            // Clear node
+            MaterialToolButton {
+                id: restartButton
```
fabiencastan: Could be renamed clearButton or deleteButton

meshroom/ui/graph.py
```diff
@@ -1169,8 +1169,29 @@ def pasteNodes(self, serializedData: str, position: Optional[QPoint]=None) -> li
...
-
-
+    
+    @Slot(Node, result=bool)
+    def canComputeNode(self, node: Node) -> bool:
+        """ Check if the node can be computed """
+        if node.isCompatibilityNode or not node.isComputableType:
+            return False
+        if node.isComputed:
+            return True
```
fabiencastan, line 1179: If a node that depends on it is currently being computed, it cannot be recomputed.

meshroom/ui/graph.py
```diff
@@ -1169,8 +1169,29 @@ def pasteNodes(self, serializedData: str, position: Optional[QPoint]=None) -> li
...
+        return False
+    
+    @Slot(Node, result=bool)
+    def canSubmitNode(self, node: Node) -> bool:
+        """ Check if the node can be submitted """
+        if node.isCompatibilityNode or not node.isComputableType:
+            return False
+        if node.isComputed:
+            return True
```
fabiencastan, line 1190: Same

meshroom/ui/qml/GraphEditor/GraphEditor.qml
```diff
@@ -1075,6 +1075,35 @@ Item {
...
+                uigraph.cancelNodeComputation(node)
+            }
+        }
+        
+        onDeleteDataRequest: function(node) {
+            uigraph.clearSelectedNodesData();
+        }
+        
+        onSubmitRequest: function(node) {
+            if (node.isComputed) uigraph.clearSelectedNodesData();
```
fabiencastan: Asymmetrical between compute and submit.
onComputeRequest does not do this check.

meshroom/ui/graph.py
Alxiice: I added a check on `getLocked` it should be ok

meshroom/ui/qml/GraphEditor/GraphEditor.qml
Alxiice: Ok + I disabled submit when we haven't cleared the node

fabiencastan (approved):

PR #2892, opened 2025-09-17
[nodes] ExpressionTextField improvements

Alxiice:
## Description

- Moved the evaluation to python
- Correct text field bindings so that the slider/reset etc works well


meshroom/ui/qml/GraphEditor/AttributeItemDelegate.qml
```diff
@@ -664,7 +670,7 @@ RowLayout {
                             if (!pressed) {
                                 _reconstruction.setAttribute(attribute, formattedValue)
                                 updateAttributeLabel()
-                            }
+                            }
```
cbentejac: Trailing whitespace
```suggestion
                            }
```

meshroom/ui/qml/Utils/ExpressionTextField.qml
```diff
@@ -30,26 +30,16 @@ TextField {
...
+                _res = Math.round(_res)
             else
-                result = result.toFixed(decimals)
-            return result
-        } catch (err) {
-            console.error("Error evaluating expression (", _text, "):", err)
+                _res = _res.toFixed(decimals)
+            return _res
+        } else {
+            console.error("Error : Expression", _text, "is invalid")
```
cbentejac:
```suggestion
            console.error("Error: Expression", _text, "is invalid")
```

cbentejac (changes requested):
Besides the clean-up comments, I have a few notes:
- `mathEvaluator.js` (in meshroom/ui/qml/Utils) can be removed from the project since it is (unless I missed something) now fully replaced, and references to it in the `qmldir` can also be removed. 
- `getEvalExpression` is called when the value is validated, when the focus on the text field is lost or when the expression has just been validated. For example, setting the value "`10*2`" then pressing "Enter" then moving to another attribute will trigger the call 3 times: once with "`10*2`", immediately after with "20" and then again with "20". The beginning and end call are necessary, but is it avoidable to retrigger the evaluation as soon as it has ended?

Otherwise, I do not have any functional notes, everything is clean, the code is actually simpler than it was in the initial version, so bar the tiny comments, it's ready to be merged.

Alxiice:
> Besides the clean-up comments, I have a few notes:
> 
>     * `mathEvaluator.js` (in meshroom/ui/qml/Utils) can be removed from the project since it is (unless I missed something) now fully replaced, and references to it in the `qmldir` can also be removed.
> 
>     * `getEvalExpression` is called when the value is validated, when the focus on the text field is lost or when the expression has just been validated. For example, setting the value "`10*2`" then pressing "Enter" then moving to another attribute will trigger the call 3 times: once with "`10*2`", immediately after with "20" and then again with "20". The beginning and end call are necessary, but is it avoidable to retrigger the evaluation as soon as it has ended?
> 
> 
> Otherwise, I do not have any functional notes, everything is clean, the code is actually simpler than it was in the initial version, so bar the tiny comments, it's ready to be merged.

I removed one occurence by checking if the text has changed since last evaluation, I'm not sure if it's better but let me know what you think!

meshroom/ui/qml/Utils/ExpressionTextField.qml
cbentejac, line 95: Is it necessary? We only go through this when selecting the node (i.e. when the `ExpressionTextFields` are displayed), but before reaching `Component.onCompleted`, we already go through `onTextChanged`, which triggers `refreshStatus` anyways.

So basically with this, when selecting a node in the Graph Editor, we evaluate all the expressions twice before doing anything.

meshroom/core/evaluation.py
```diff
@@ -0,0 +1,48 @@
...
+    }
+    # Allowed AST node types
+    allowed_nodes = (
+        ast.Expression, ast.BinOp, ast.UnaryOp, ast.Call, ast.Name, ast.Load,
+        ast.Add, ast.Sub, ast.Mult, ast.Div, ast.Pow, ast.Mod, ast.FloorDiv,
+        ast.USub, ast.UAdd, ast.BitXor, ast.BitOr, ast.BitAnd,
+        ast.LShift, ast.RShift, ast.Invert,
+        ast.Constant
+    )
+    def _validate_ast(self, node):
```
Alxiice, line 34:
```suggestion

    def _validate_ast(self, node):
```

meshroom/core/evaluation.py
```diff
@@ -0,0 +1,48 @@
...
+        for child in ast.walk(node):
+            if not isinstance(child, self.allowed_nodes):
+                raise ValueError(f"Bad expression: {ast.dump(child)}")
+            # Check that all variable/function names are whitelisted
+            if isinstance(child, ast.Name):
+                if child.id not in self.allowed_symbols:
+                    raise ValueError(f"Unknown symbol: {child.id}")
+    def evaluate(self, expr: str):
```
Alxiice, line 43:
```suggestion

    def evaluate(self, expr: str):
```

meshroom/core/evaluation.py
```diff
@@ -0,0 +1,48 @@
...
+    """
+    
+    # Allowed math symbols
+    allowed_symbols = {
+        "e": math.e, "pi": math.pi,
+        "cos": math.cos, "sin": math.sin, "tan": math.tan, "exp": math.exp,
+        "pow": pow, "round": round, "abs": abs, "min": min, "max": max,
+        "sqrt": math.sqrt, "log": math.log
+    }
+    # Allowed AST node types
```
Alxiice, line 25:
```suggestion

    # Allowed AST node types
```

cbentejac (changes requested):
There's a regression that has probably been introduced with #2836: when editing a text field, not pressing "Enter" and clicking somewhere else (thus losing focus), the value of the field is lost (the attribute is then set with value "0"). This wasn't the case in Meshroom 2025.1.0: when losing the focus, the value was still validated and applied to the attribute. 

The `Component.onCreated` slot should also be removed, as it performs an evaluation right after the one triggered by `onTextChanged` while there could not possibly have been a value change.

Alxiice:
> There's a regression that has probably been introduced with #2836: when editing a text field, not pressing "Enter" and clicking somewhere else (thus losing focus), the value of the field is lost (the attribute is then set with value "0"). This wasn't the case in Meshroom 2025.1.0: when losing the focus, the value was still validated and applied to the attribute.
> 
> The `Component.onCreated` slot should also be removed, as it performs an evaluation right after the one triggered by `onTextChanged` while there could not possibly have been a value change.

The best way I found for now was to call `accepted` on `Component.onDestroyed`, it seems to fix the problem

cbentejac (approved):

PR #2897, opened 2025-09-19
[bin] batch: better flexibility to set InitNodes from command line

fabiencastan:
Allow to set the input on all InitNodes.
Allow to set the input to specific init nodes.

Note: Use "=" and "," instead of ":" and ";".

### DIFF
--- a/bin/meshroom_batch
+++ b/bin/meshroom_batch
@@ -158,10 +158,6 @@ args = parser.parse_args()
 
 logging.getLogger().setLevel(logStringToPython[args.verbose])
 
-if not args.input and not args.inputRecursive:
-    print('Nothing to compute. You need to set --input or --inputRecursive.')
-    sys.exit(1)
-
 meshroom.core.initPlugins()
 meshroom.core.initNodes()
 
--- a/bin/meshroom_batch
+++ b/bin/meshroom_batch
@@ -10,6 +10,27 @@ import sys
 import meshroom.core.graph
 from meshroom import setupEnvironment, logStringToPython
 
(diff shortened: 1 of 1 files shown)


bin/meshroom_batch
```diff
@@ -156,60 +177,30 @@ with meshroom.core.graph.GraphModification(graph):
...
+            if nodeName and nodeName not in initNodesNames:
+                raise RuntimeError(f"Failed to find the Init Node '{nodeName}' in your pipeline.\nAvailable Init Nodes: {initNodesNames}")
+
+        # feed inputs (recursive and non-recursive paths) to corresponding init nodes
+        for initNode in initNodes:
+            nodeName = initNode.getName()
+            # Retrieve input per node and inputs for all init node types
+            input = mapInput.get(nodeName, []) + mapInput.get("", [])
+            # Retrieve recursive inputs
+            inputRec = mapInputRecursive.get(nodeName, []) + mapInputRecursive.get("", [])
```
cbentejac, line 207: There is kind of a corner case when some of the `InitNodes` are not set. By calling `initialize` with `[]` values, we may run into errors within the `initialize` function that are not explicit (for example, with CameraInit nodes, we end up with this error: `ValueError("No valid input file or no image in the provided directories")`.

I think it'd be best to either raise an exception before calling `initialize` or to set the node with the value used to set the previous one (similarly to what is done when no specific instance of an `InitNode` is specified in the command line) when we get the default `[]` value.

fabiencastan: Done

PR #2898, opened 2025-09-20
[core] Rez plugins: Ensure dependencies for `meshroom_compute` are present

cbentejac:
## Description

This PR introduces a mechanism to ensure that required dependencies for `meshroom_compute`, specifically `psutil`, are always included in the node execution environment for rez plugins. Dependencies for `meshroom_compute` are added to the environment without specifying any version or attempting to match Meshroom's, in order for it to be set directly during the environment resolution.

**Dependency management improvements:**

* Introduced a new constant `_MESHROOM_COMPUTE_DEPS` in `meshroom/core/desc/node.py` to define required dependencies for `meshroom_compute`.
* Updated `meshroom/core/plugins.py` to import `_MESHROOM_COMPUTE_DEPS` and ensure all dependencies listed in it (currently `psutil`) are always included in the resolved execution environment for plugins, unless already present [[1]](diffhunk://#diff-1d5f2a46e59213721e5624ac787e491e5cc2333ab8838b226c6558ceb0d87913L17-R17) [[2]](diffhunk://#diff-1d5f2a46e59213721e5624ac787e491e5cc2333ab8838b226c6558ceb0d87913R197-R205).

### DIFF
--- a/meshroom/core/plugins.py
+++ b/meshroom/core/plugins.py
@@ -14,7 +14,7 @@ from pathlib import Path
 
 from meshroom.common import BaseObject
 from meshroom.core import desc
-from meshroom.core.desc.node import _MESHROOM_ROOT
+from meshroom.core.desc.node import _MESHROOM_ROOT, _MESHROOM_COMPUTE_DEPS
 
 
 def validateNodeDesc(nodeDesc: desc.Node) -> list:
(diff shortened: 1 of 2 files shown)


meshroom/core/plugins.py
```diff
@@ -194,6 +194,15 @@ def resolveRezSubrequires(self) -> list[str]:
                 if not definedInParentEnv:
                     packages.append(package)
 
+        for package in _MESHROOM_COMPUTE_DEPS:
+            # For packages that are required by meshroom_compute, do not specify any version
+            # or align it with Meshroom's: the version will be found during the resolution of
+            # the environment based on the other packages.
+            # If any of these packages is already part of the environment a plugin's dependency,
+            # do not add it
+            if not package in " ".join(packages):
```
fabiencastan:
```suggestion
            if package not in " ".join(packages):
```

meshroom/core/plugins.py
```diff
@@ -194,6 +194,15 @@ def resolveRezSubrequires(self) -> list[str]:
...
+                packages.append(package)
```
fabiencastan, line 216:
```suggestion
        def extract_package_name(package_string):
            return re.split(r'[-=<>!~]+', package_string, 1)[0]
        package_names = [extract_package_name(pkg) for pkg in packages]

        for package in _MESHROOM_COMPUTE_DEPS:
            # For packages that are required by meshroom_compute, do not specify any version
            # or align it with Meshroom's: the version will be found during the resolution of
            # the environment based on the other packages.
            # If any of these packages is already part of the environment a plugin's dependency,
            # do not add it
            if package not in package_names :
                packages.append(package)
```

cbentejac: Done

fabiencastan (approved):

PR #2907, opened 2025-09-26
[ci] Linting: Raise "'module' imported but not used" to error-level

cbentejac:


cbentejac:
Some intermediate imports are needed but flagged as "unused" since they are not directly used in the file in which they are imported.

PR #2910, opened 2025-10-07
[core] Correctly set `_sourceCodeFolder` in `BaseNode`'s constructor

cbentejac:
## Description

This PR fixes a regression introduced by the refactoring of the command line variables in #2888, where the static `nodeSourceCodeFolder` is initialized but never set with the correct value once it becomes available. 

This is fixed by initializing `_sourceCodeFolder` directly in the `BaseNode`'s constructor if the description is available (if it isn't, the node will be in `Unknown Node` compatibility error anyway). This both simplifies what was done prior to #2888 (where `_sourceCodeFolder` was correctly set later than it should have been) and ensures that `_staticExpVars["nodeSourceCodeFolder"]` is initialized with the correct value as well.

Additionally, simple unit tests are added to ensure that static variables are always correctly set. A test `InputNode` is also added to one of the test plugins to cover all cases.

### DIFF
--- a/meshroom/core/node.py
+++ b/meshroom/core/node.py
@@ -359,7 +359,7 @@ class NodeChunk(BaseObject):
             return f"{self.node.name}({self.index})"
         else:
             return self.node.name
-    
+
     @property
     def logManager(self):
         if self._logManager is None:
@@ -520,7 +520,7 @@ class NodeChunk(BaseObject):
         executionStatus = None
         self.statThread = stats.StatisticsThread(self)
         self.statThread.start()
-        
+
         try:
             self.node.nodeDesc.processChunk(self)
             # NOTE: this assumes saving the output attributes for each chunk
@@ -672,7 +672,7 @@ class BaseNode(BaseObject):
         self.packageName: str = ""
         self.packageVersion: str = ""
         self._internalFolder: str = ""
-        self._sourceCodeFolder: str = ""
+        self._sourceCodeFolder: str = self.nodeDesc.sourceCodeFolder if self.nodeDesc else ""
         self._internalFolderExp = "{cache}/{nodeType}/{uid}"
 
         # temporary unique name for this node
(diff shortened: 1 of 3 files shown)


meshroom/core/node.py
fabiencastan: Should be initialized from nodeDesc

meshroom/core/node.py
```diff
@@ -1830,6 +1830,10 @@ def __init__(self, nodeType, position=None, parent=None, uid=None, **kwargs):
         self.packageVersion = self.nodeDesc.packageVersion
         self._sourceCodeFolder = self.nodeDesc.sourceCodeFolder
 
+        # _sourceCodeFolder is not available in the BaseNode constructor, so it must be
+        # set here with the value coming from the description
+        self._staticExpVars["nodeSourceCodeFolder"] = self._sourceCodeFolder
+
```
fabiencastan, line 1831: Should be in BaseNode (common with InputNodes).

cbentejac: `InputNodes` exist in desc/node.py, not in node.py. This here is common to `Nodes` and `InputNodes` (but I should add a test for it!).

meshroom/core/node.py
cbentejac: Yes, you are correct. This is now done (it simplifies the changes of this PR).

meshroom/core/node.py
```diff
@@ -672,7 +672,7 @@ def __init__(self, nodeType: str, position: Position = None, parent: BaseObject
         self.packageName: str = ""
         self.packageVersion: str = ""
         self._internalFolder: str = ""
-        self._sourceCodeFolder: str = ""
+        self._sourceCodeFolder: str = "" if not self.nodeDesc else self.nodeDesc.sourceCodeFolder
```
```suggestion
        self._sourceCodeFolder: str = self.nodeDesc.sourceCodeFolder if self.nodeDesc else ""
```

fabiencastan (approved):

meshroom/core/node.py

fabiencastan (approved):

PR #2918, opened 2025-10-13
[core] Add support for dynamic chunks

Alxiice:
## Description

This PR goal is to change chunk system to be able to work with graphs where nodes can have no specific size set.

## Changes

**Compute system**
- New levels have been added on the computation : 
  - `extreme` (that targets higher specs that what intensive does)
  - `script` which is a mode for single-process simple process. It targets specific machines on the farm that have the ability to run multiple jobs in parallel so for simple jobs it should allow faster processing
- Delayed chunks creation
  - `node`
    - A `_chunksCreated` param has been added to check if the chunks have been correctly initialized
    - Now by default we use `resetChunks` when loading meshroom, that will create a list with a single chunk, and we can create chunks at any moment with `_createChunks`.
    - A new `NodeStatusData` has been added that tracks a `nodeStatus` file which is similar to the chunk statuses files but is specific to the node. This node status is used to get the cached range parametrization of the node so that we can retrieve it when possible with `_createChunksFromCache`.
  - `taskManager`
    - The `TaskThread` and `TaskManager` have been modified to launch the chunk creation when the node compute starts

**Submitters**
- Submitters have been moved to another package (TBD where...)
- The `BaseSubmitter` API have been updated to allow more flexibility. A new `BaseSubmittedJob` exists and is used to track jobs that have been created. The goal is to use this as an interface to call actions on it (stop/pause/restart...)
- A new `bin/meshroom_createChunks` script have been created, and handles the chunk creation and additional chunk tasks spooling
  - Calls the chunk creation
  - Checks if we can spool additioanl tasks
  - If so, execute queue tasks that will compute the chunks
  - If not, execute the chunks serially on the current process

Changes to the tractor API have been implemented here https://github.com/meshroomHub/mrSubmitters/pull/1

## Examples

Peek 2025-10-09 19-05

ezgif-1a582893f45983


meshroom/core/node.py
```diff
@@ -164,57 +313,39 @@ def initLocalSubmit(self):
...
+        self.computeSessionUid = meshroom.core.sessionUid
         self.endDateTime = datetime.datetime.now().strftime(self.dateTimeFormatting)
         if self._startTime != None:
             self.elapsedTime = time.time() - self._startTime
 
     @property
     def elapsedTimeStr(self):
         return str(datetime.timedelta(seconds=self.elapsedTime))
-
+
```
cbentejac: Trailing whitespaces

bin/meshroom_createChunks
```diff
@@ -0,0 +1,145 @@
...
+            # We ensure that no other instance has started to compute, by checking that the sessicomputeSessionUidonUid is empty.
+            if chunk.node.getMrNodeType() == meshroom.core.MrNodeType.NODE and \
+                not chunk.status.computeSessionUid and node._nodeStatus.submitterSessionUid:
+                continue
+            logging.warning(
+                f"[MeshroomCreateChunks] Node is already submitted with status " \
+                f"\"{chunk.status.status.name}\". See file: \"{chunk.statusFile}\". " \
+                f"ExecMode: {chunk.status.execMode.name}, computeSessionUid: {chunk.status.computeSessionUid}, " \
+                f"submitterSessionUid: {node._nodeStatus.submitterSessionUid}")
+            # sys.exit(-1)
```
cbentejac: To remove if it's not going to be used

meshroom/core/node.py
```diff
@@ -2230,3 +2599,5 @@ def upgrade(self):
     canUpgrade = Property(bool, canUpgrade.fget, constant=True)
     issueDetails = Property(str, issueDetails.fget, constant=True)
 
+    issueDetails = Property(str, issueDetails.fget, constant=True)
```
cbentejac:
```suggestion
```

This seems to be a duplication from line 2600, it should be removed.

meshroom/ui/qml/GraphEditor/TaskManager.qml
```diff
@@ -51,222 +62,430 @@ Item {
...
-                Layout.preferredHeight: parent.height
-                horizontalAlignment: Label.AlignHCenter
-                verticalAlignment: Label.AlignVCenter
-                background: Rectangle {
-                    color: headBgColor
+            
+            // Job toolbar
+            Rectangle {
+                // Layout.preferredWidth: childrenRect.width
+                // Layout.preferredHeight: childrenRect.height
```
cbentejac:
```suggestion
```

meshroom/ui/qml/GraphEditor/TaskManager.qml
```diff
@@ -51,222 +62,430 @@ Item {
...
+                        model: object.chunks
+                        property var node: object
+
+                        spacing: 3
+                        
+                        delegate: Loader {
+                            id: chunkDelegate
+                            // width: ListView.view.model 
+                            //     ? ListView.view.width / ListView.view.model.count - chunkList.spacing
+                            //     : 0
```
cbentejac:
```suggestion
```

meshroom/ui/qml/GraphEditor/TaskManager.qml
```diff
@@ -51,222 +62,430 @@ Item {
...
+                        // Placeholder for uninitialized chunks
+                        Label {
+                            enabled: chunkList.model.count == 0
+                            visible: enabled
                             anchors.fill: parent
-                            onPressed: {
-                                selectNode(chunkList.node)
+                            background: Rectangle {
+                                color: Colors.darkpurple
+                                // color: Colors.getNodeColor(chunkList.node, {"NONE": Colors.darkpurple})
```
cbentejac:
```suggestion
```

meshroom/ui/graph.py
```diff
@@ -570,6 +591,217 @@ def cancelNodeComputation(self, node):
...
+                logging.warning(f"Error on stopTask :\n{e}")
+            else:
+                node.updateNodeStatusFromCache()
+                node.upgradeStatusTo(Status.STOPPED)
+                # TODO : Stop depending nodes ?
+                self.parent().showMessage(f"Stopped node {node.label}")
+        else:
+            self.cancelNodeComputation(node)
+            node.stopComputation()
+            # self._taskManager.removeNode(node, processList=True)
```
cbentejac:
```suggestion
```

meshroom/core/desc/computation.py
```diff
@@ -1,13 +1,15 @@
...
 
 from .attribute import ListAttribute, IntParam
 
 
-class Level(Enum):
+class Level(IntEnum):
     NONE = 0
     NORMAL = 1
     INTENSIVE = 2
+    EXTREME = 2
```
cbentejac: Shouldn't this be:
```suggestion
    EXTREME = 3
```
as used in https://github.com/meshroomHub/mrSubmitters/pull/1/files#diff-9dd253882930191e431fa00981aa86f2b39630037daed18a8fc848eaf3766823R77?

meshroom/ui/graph.py
```diff
@@ -570,6 +591,217 @@ def cancelNodeComputation(self, node):
...
+        if job:
+            try:
+                job.pauseJob()
+            except Exception as e:
+                logging.warning(f"Error on pauseJob :\n{e}")
+                self.parent().showMessage(f"Failed to pause the job for node {node}", "error")
+            else:
+                self.parent().showMessage(f"Paused node {node.label} on farm")
+        elif not node.isExtern():
+            self.parent().showMessage(f"PauseJob is only available in local computation mode !", "warning")
```
cbentejac:
```suggestion
            self.parent().showMessage(f"PauseJob is only available in external computation mode!", "warning")
```

meshroom/ui/graph.py
```diff
@@ -570,6 +591,217 @@ def cancelNodeComputation(self, node):
...
+    def resumeJob(self, node: Node):
+        """ Resume the paused job
+        """
+        job = jobManager.getNodeJob(node)
+        if job:
+            # Node is submitted to farm
+            try:
+                job.resumeJob()
+            except Exception as e:
+                self.parent().showMessage(f"Failed to rsume node {node.label} on farm")
```
cbentejac:
```suggestion
                self.parent().showMessage(f"Failed to resume node {node.label} on farm")
```

meshroom/ui/graph.py
```diff
@@ -570,6 +591,217 @@ def cancelNodeComputation(self, node):
...
+                logging.warning(f"Error on resumeJob :\n{e}")
```
cbentejac:
```suggestion
                logging.warning(f"Error on resumeJob:\n{e}")
```

meshroom/ui/graph.py
```diff
@@ -570,6 +591,217 @@ def cancelNodeComputation(self, node):
...
+    def interruptJob(self, node: Node):
+        """ Interrupt the job that processes the node
+        """
+        job = jobManager.getNodeJob(node)
+        if job:
+            try:
+                job.interruptJob()
+            except Exception as e:
+                self.parent().showMessage(f"Failed to interrupt node {node.label} on farm", "error")
+                logging.warning(f"Error on interruptJob :\n{e}")
```
cbentejac:
```suggestion
                logging.warning(f"Error on interruptJob:\n{e}")
```

meshroom/ui/graph.py
```diff
@@ -570,6 +591,217 @@ def cancelNodeComputation(self, node):
...
+                    if jobManager.getNodeJob(chunk.node) == job:
+                        chunk.upgradeStatusTo(Status.SUBMITTED)
+                for node in self._graph.dfsOnFinish(None)[0]:
+                    if not node._chunksCreated and node._nodeStatus.status in (Status.ERROR, Status.STOPPED, Status.KILLED):
+                        node.upgradeStatusTo(Status.SUBMITTED)
+                job.restartErrorTasks()
+                job.resumeJob()
+            except Exception as e:
+                self.parent().showMessage(f"Failed to restart error tasks for node {node.label} on farm", "error")
+                logging.warning(f"Error on restartJobErrorTasks :\n{e}")
```
cbentejac:
```suggestion
                logging.warning(f"Error on restartJobErrorTasks:\n{e}")
```

meshroom/ui/graph.py
```diff
@@ -570,6 +591,217 @@ def cancelNodeComputation(self, node):
...
+            else:
+                self.parent().showMessage(f"Restared error tasks for the node {node}")
```
cbentejac:
```suggestion
                self.parent().showMessage(f"Restarted error tasks for the node {node}")
```

meshroom/ui/graph.py
```diff
@@ -570,6 +591,217 @@ def cancelNodeComputation(self, node):
...
+                for node in self._graph.dfsOnFinish(None)[0]:
+                    if not node._chunksCreated and node._nodeStatus.status in (Status.SUBMITTED, Status.RUNNING):
+                        node.upgradeStatusTo(Status.STOPPED)
+                self.parent().showMessage(f"Interrupted the job for node {node}")
+        elif not node.isExtern():
+            self._taskManager.clear()
+            for chunk in node._chunks:
+                if chunk._status.status == Status.RUNNING and not chunk.isExtern():
+                    chunk.stopProcess()
+            self.parent().showMessage(f"Cleared the task manager")
```
cbentejac: When entering this case, the task manager is correctly cleared but in the Graph Editor, nodes that follow the one that was interrupted still remain locked and in "SUBMITTED" status, and appear as though they will be computed (with the start/stop/restart button in the "stop" position).

For them to return to a "normal" state, it is necessary to do a "stop" from the Graph Editor.

meshroom/ui/graph.py
```diff
@@ -570,6 +591,217 @@ def cancelNodeComputation(self, node):
...
+    def pauseJob(self, node: Node):
+        """ Pause the running job : cancel all scheduled tasks.
+        Current task don't stop but future tasks won't be launched
+        """
+        job = jobManager.getNodeJob(node)
+        if job:
+            try:
+                job.pauseJob()
+            except Exception as e:
+                logging.warning(f"Error on pauseJob :\n{e}")
```
cbentejac:
```suggestion
                logging.warning(f"Error on pauseJob:\n{e}")
```

meshroom/ui/graph.py
```diff
@@ -570,6 +591,217 @@ def cancelNodeComputation(self, node):
...
+        chunk.updateStatusFromCache()
+        node = chunk.node
+        chunkIteration = chunk.range.iteration
+        job = jobManager.getNodeJob(node)
+        if job:
+            try:
+                job.skipChunkTask(node, chunkIteration)
+            except Exception as e:
+                self.parent().showMessage(f"Failed to skip chunk {chunkIteration} of {node.label}", "error")
+                logging.warning(f"Error on skipTask :\n{e}")
```
cbentejac:
```suggestion
                logging.warning(f"Error on skipTask:\n{e}")
```

meshroom/ui/graph.py
```diff
@@ -570,6 +591,217 @@ def cancelNodeComputation(self, node):
...
+            chunkIteration = chunk.range.iteration
+            try:
+                chunk.updateStatusFromCache()
+                chunk.upgradeStatusTo(Status.SUBMITTED)
+                job.restartChunkTask(node, chunkIteration)
+            except Exception as e:
+                chunk.updateStatusFromCache()
+                chunk.upgradeStatusTo(Status.ERROR)
+                self.parent().showMessage(f"Failed to relaunch chunk {chunkIteration} of {node.label}", "error")
+                logging.warning(f"Error on restartTask :\n{e}")
```
cbentejac:
```suggestion
                logging.warning(f"Error on restartTask:\n{e}")
```

meshroom/ui/graph.py
```diff
@@ -570,6 +591,217 @@ def cancelNodeComputation(self, node):
...
+    @Slot(Node)
+    def stopNode(self, node: Node):
+        """ Stop the selected task """
+        job = jobManager.getNodeJob(node)
+        if job:
+            try:
+                job.stopChunkTask(node, -1)
+            except Exception as e:
+                self.parent().showMessage(f"Failed to stop node {node.label}", "error")
+                logging.warning(f"Error on stopTask :\n{e}")
```
cbentejac:
```suggestion
                logging.warning(f"Error on stopTask:\n{e}")
```

meshroom/ui/graph.py
```diff
@@ -570,6 +591,217 @@ def cancelNodeComputation(self, node):
...
+            return
+        node = chunk.node
+        job = jobManager.getNodeJob(node)
+        if job:
+            chunkIteration = chunk.range.iteration
+            try:
+                job.stopChunkTask(node, chunkIteration)
+            except Exception as e:
+                self.parent().showMessage(f"Failed to stop chunk {chunkIteration} of {node.label}", "error")
+                logging.warning(f"Error on stopTask :\n{e}")
```
cbentejac:
```suggestion
                logging.warning(f"Error on stopTask:\n{e}")
```

meshroom/core/submitter.py
```diff
@@ -1,23 +1,228 @@
...
+            raise NotImplementedError("'resumeJob' method must be implemented in subclasses")
+        else:
+            raise RuntimeError(f"Submitter {self.__class__.__name__} cannot interrupt the job")
+
+    def interruptJob(self):
+        """ This will interrupt the job (and kill running tasks) """
+        if self.submitterOptions.includes(SubmitterOptionsEnum.INTERRUPT_JOB):
+            raise NotImplementedError("'interruptJob' method must be implemented in subclasses")
+        else:
+            raise RuntimeError(f"Submitter {self.__class__.__name__} cannot interrupt the job")
```
cbentejac, line 128: This works well to interrupt the job (on the farm, everything is updated as expected and the running process is killed), but there seems to be missing updates on the UI-side:

Image

All the submitted tasks have been stopped, and the running one still appears as running (both in the task manager and in the Graph Editor) although it has been successfully stopped on the farm. Refreshing the nodes' status has no effect.

meshroom/ui/qml/GraphEditor/TaskManager.qml
```diff
@@ -51,222 +62,430 @@ Item {
...
+                border.width: 2
+                radius: 8
+
+                ColumnLayout {
+                    id: jobColumn
+                    anchors.centerIn: parent
+                    spacing: 2
+
+                    MaterialToolButton {
+                        ToolTip.text: "Pause job"
```
cbentejac:
```suggestion
                        ToolTip.text: "Pause Job"
```

meshroom/ui/qml/GraphEditor/TaskManager.qml
```diff
@@ -51,222 +62,430 @@ Item {
...
+                        enabled: selectedChunk !== null
+                        text: MaterialIcons.replay_circle_filled
+                        font.pointSize: 15
+                        onClicked: {
+                            uigraph.restartTask(selectedChunk)
+                        }
+                    }
+
+                    MaterialToolButton {
+                        ToolTip.text: "Skip task"
```
cbentejac:
```suggestion
                        ToolTip.text: "Skip Task"
```

meshroom/ui/qml/GraphEditor/TaskManager.qml
```diff
@@ -51,222 +62,430 @@ Item {
...
 
-            width: parent.width
+                ColumnLayout {
+                    id: taskColumn
+                    anchors.centerIn: parent
+                    spacing: 2
 
-            z: 2
+                    MaterialToolButton {
+                        ToolTip.text: "Stop task"
```
cbentejac:
```suggestion
                        ToolTip.text: "Stop Task"
```

meshroom/ui/qml/GraphEditor/TaskManager.qml
```diff
@@ -51,222 +62,430 @@ Item {
...
-            Label {
-                text: qsTr("Exec Mode")
-                Layout.preferredWidth: execMetrics.width + 20
-                Layout.preferredHeight: parent.height
-                horizontalAlignment: Label.AlignHCenter
-                verticalAlignment: Label.AlignVCenter
-                background: Rectangle {
-                    color: headBgColor
+                    MaterialToolButton {
+                        ToolTip.text: "Restart task"
```
cbentejac:
```suggestion
                        ToolTip.text: "Restart Task"
```

meshroom/ui/qml/GraphEditor/TaskManager.qml
```diff
@@ -51,222 +62,430 @@ Item {
...
+                        enabled: root.uigraph.selectedNode !== null
+                        text: MaterialIcons.pause_circle_filled
+                        font.pointSize: 15
+                        onClicked: {
+                            uigraph.pauseJob(uigraph.selectedNode)
+                        }
+                    }
+
+                    MaterialToolButton {
+                        ToolTip.text: "Resume job"
```
cbentejac:
```suggestion
                        ToolTip.text: "Resume Job"
```

meshroom/ui/qml/GraphEditor/TaskManager.qml
```diff
@@ -51,222 +62,430 @@ Item {
...
+                        enabled: root.uigraph.selectedNode !== null
+                        text: MaterialIcons.play_circle_filled
+                        font.pointSize: 15
+                        onClicked: {
+                            uigraph.resumeJob(uigraph.selectedNode)
+                        }
+                    }
+
+                    MaterialToolButton {
+                        ToolTip.text: "Interrupt job"
```
cbentejac:
```suggestion
                        ToolTip.text: "Interrupt Job"
```

meshroom/ui/qml/GraphEditor/TaskManager.qml
```diff
@@ -51,222 +62,430 @@ Item {
...
+                        enabled: root.uigraph.selectedNode !== null
+                        text: MaterialIcons.stop_circle
+                        font.pointSize: 15
+                        onClicked: {
+                            uigraph.interruptJob(uigraph.selectedNode)
+                        }
+                    }
+                    
+                    MaterialToolButton {
+                        ToolTip.text: "Restart all error tasks"
```
cbentejac:
```suggestion
                        ToolTip.text: "Restart All Error Tasks"
```

meshroom/ui/qml/GraphEditor/TaskManager.qml
```diff
@@ -51,222 +62,430 @@ Item {
...
+                        Layout.alignment: Qt.AlignHCenter
+                        enabled: root.uigraph.selectedNode !== null
+                        text: MaterialIcons.replay_circle_filled  // MaterialIcons.restart_alt
```
cbentejac:
```suggestion
                        text: MaterialIcons.replay_circle_filled
```

cbentejac (changes requested):
Additional notes:

- Let's say I compute a graph (for example, the photogrammetry template) in a given location. All my nodes have been successfully computed, the statuses are all set to "SUCCESS". Now let's say I do a "save as" on the graph I just computed and save it to a new location. The cache changes, so we expect all the nodes' status to be reset. For nodes that have dynamic chunks, the display of the status is not refreshed, and we end up with nodes that appear as computed, although they have lost all their status files.
Below is a screenshot of the photogrammetry graph that was computed and saved in a new location (the first branch was computed, the second was added from the template for the sake of comparison):
Image

- There are some unpredictable behaviours when performing actions that are allowed but probably shouldn't be. An example would be clicking on "stop task" for a task that is already done computing and is in "SUCCESS" state, while nodes are being computed later on in the graph. The UI allows to perform it, there is an info message saying that the task has been successfully stopped, and the node's status is updated to "STOPPED" in the task manager and Graph Editor; it does not enable the "resume job" button. This seems to cause the job on the farm to finish computing the current task and then pausing the rest of the job (to be verified).  

- I have noticed on several occasions that when clicking on buttons from the "JOB" tab, the info message that is sent contains the name of a node that is not part of the job at all (if all my submitted tasks are for NodeName_2, I may get a message about NodeName_1, which is not being computed at all).

cbentejac (changes requested):
Additional notes:

- Let's say I compute a graph (for example, the photogrammetry template) in a given location. All my nodes have been successfully computed, the statuses are all set to "SUCCESS". Now let's say I do a "save as" on the graph I just computed and save it to a new location. The cache changes, so we expect all the nodes' status to be reset. For nodes that have dynamic chunks, the display of the status is not refreshed, and we end up with nodes that appear as computed, although they have lost all their status files.
Below is a screenshot of the photogrammetry graph that was computed and saved in a new location (the first branch was computed, the second was added from the template for the sake of comparison):
Image

- There are some unpredictable behaviours when performing actions that are allowed but probably shouldn't be. An example would be clicking on "stop task" for a task that is already done computing and is in "SUCCESS" state, while nodes are being computed later on in the graph. The UI allows to perform it, there is an info message saying that the task has been successfully stopped, and the node's status is updated to "STOPPED" in the task manager and Graph Editor; it does not enable the "resume job" button. This seems to cause the job on the farm to finish computing the current task and then pausing the rest of the job (to be verified).  

- I have noticed on several occasions that when clicking on buttons from the "JOB" tab, the info message that is sent contains the name of a node that is not part of the job at all (if all my submitted tasks are for NodeName_2, I may get a message about NodeName_1, which is not being computed at all).

meshroom/core/desc/computation.py
Alxiice: yes

meshroom/ui/qml/GraphEditor/TaskManager.qml
Alxiice: Here the darkpurple was a debug tool. Instead I will use the commented line, so that for scheduled nodes without chunks we see the submission status

meshroom/core/submitter.py
Alxiice: It should be better now

meshroom/core/desc/computation.py
```diff
@@ -1,13 +1,15 @@
...
 from .attribute import ListAttribute, IntParam
 
 
-class Level(Enum):
+class Level(IntEnum):
     NONE = 0
     NORMAL = 1
     INTENSIVE = 2
+    EXTREME = 3
+    SCRIPT=-1
```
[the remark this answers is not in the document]
Alxiice: We can put it before NONE if it's better but the reason is that's levels are going from less system-inensive to more intensivity.

By the way jobs with script level are not starting now in our tractor instance we should see with Thibault why is that

meshroom/core/taskManager.py
```diff
@@ -426,17 +534,20 @@ def submit(self, graph, submitter=None, toNodes=None, submitLabel="{projectName}
         try:
             res = sub.submit(nodesToProcess, edgesToProcess, graph.filepath, submitLabel=submitLabel)
             if res:
+                if isinstance(res, BaseSubmittedJob):
+                    jobManager.addJob(res, nodesToProcess)
+            else:
                 for node in nodesToProcess:
-                    node.destroyed.connect(lambda obj=None, name=node.name: self.onNodeDestroyed(obj, name))
-                    node.initStatusOnSubmit()  # update node status
+                    # TODO : Notify the node that there was an issue on submit
```
Alxiice, line 541: to do

meshroom/core/graph.py
```diff
@@ -1730,11 +1755,19 @@ def submitGraph(graph, submitter, toNodes=None, submitLabel="{projectName}"):
...
+
     try:
         res = sub.submit(nodesToProcess, edgesToProcess, graph.filepath, submitLabel=submitLabel)
         if res:
+            if isinstance(res, BaseSubmittedJob):
+                jobManager.addJob(res, nodesToProcess)
+        else:
             for node in nodesToProcess:
-                node.initStatusOnSubmit()  # update node status
+                # TODO : Notify the node that there was an issue on submit
```
Alxiice, line 1773: to do

meshroom/core/node.py
```diff
@@ -1644,26 +1930,28 @@ def updateDuplicates(self, nodesPerUid):
...
     def submitterStatusInThisSession(self) -> bool:
-        if not self._chunks:
-            return False
-        for chunk in self._chunks:
-            if chunk.status.submitterSessionUid != meshroom.core.sessionUid:
-                return False
-        return True
+        """ Check if the node is submitted by the current session
+        TODO: Not used -> depreciate ?
+        """
```
Alxiice:
```suggestion
```

meshroom/core/node.py
```diff
@@ -1644,26 +1930,28 @@ def updateDuplicates(self, nodesPerUid):
...
+        """ Check if chunks of the node are being computed in the current session
+        TODO: Not used -> depreciate ?
+        """
         if not self._chunks:
             return False
         for chunk in self._chunks:
-            if chunk.status.sessionUid != meshroom.core.sessionUid:
+            if chunk._status.computeSessionUid != meshroom.core.sessionUid:
                 return False
         return True
```
Alxiice:
```suggestion

```

cbentejac (approved):

PR #2920, opened 2025-10-13
[core] node: Update status when reevaluating a node's locked status

cbentejac:
## Description

This PR fixes an issue that occurs for nodes with dynamic output attributes that are computed externally: once the node's status was updated to "SUCCESS", the values of the dynamic output attributes were not refreshed, and they remained as they were prior to the computation (although they had correctly been written in the node's `values` file). This never occurred for locally-computed nodes. To work around this issue, the user needed to use the "Refresh Nodes Status" option. 

With this PR, if a node has been successfully computed and its locked status is re-evaluated, then its status is updated from the cache, ensuring all the information are correctly reflected in the Graph Editor without needing an action from the user.

### DIFF
--- a/meshroom/core/node.py
+++ b/meshroom/core/node.py
@@ -1603,6 +1603,11 @@ class BaseNode(BaseObject):
             return
 
         if currentStatus == Status.SUCCESS:
+            # Update the status when the lock is reevaluated while the node is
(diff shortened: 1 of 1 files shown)


fabiencastan (changes requested):
Locked is a consequence of the status update, not the source.

PR #2930, opened 2025-10-29
[Shapes] `TextLayer`: Enable wrapping for large texts

cbentejac:
## Description

This PR updates the layout for `Text` shapes to improve the display of large texts (e.g. image captions) by enabling the text wrapping. Additionally, the text is not centered anymore on the obervation's `x`- and `y`-coordinates, which now represent its top-left corner.

## Features list

- [x] Changed the calculation of `x` and `y` so that text is now positioned directly at the observation's coordinates, rather than being centered using the implicit width and height.
- [x] Set the `width` and `height` of the `Text` element to the remaining space from the observation center to the edge of the container, allowing for better text wrapping and layout.
- [x] Updated `wrapMode` from `Text.NoWrap` to `Text.Wrap` to allow text to wrap within the specified width, improving readability.

### DIFF
--- a/meshroom/ui/qml/Shapes/Viewer/Layers/TextLayer.qml
+++ b/meshroom/ui/qml/Shapes/Viewer/Layers/TextLayer.qml
@@ -16,11 +16,14 @@ BaseLayer {
     id: textLayer
 
     Text {
-        x: textLayer.observation.center.x - implicitWidth * 0.5   // Center text horizontally
-        y: textLayer.observation.center.y - implicitHeight * 0.5  // Center text vertically
+        width: ShapeViewerHelper.containerWidth - textLayer.observation.center.x
+        height: ShapeViewerHelper.containerHeight - textLayer.observation.center.y
+
+        x: textLayer.observation.center.x
+        y: textLayer.observation.center.y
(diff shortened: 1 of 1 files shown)


gregoire-dl:
I have two minor comments, up for discussion:

- If I’m not mistaken, the shape `BaseLayer` completely fills the `Viewer2D` image container. So, you could use `parent.width` and `parent.height` instead of relying on `ShapeViewerHelper.containerWidth` and `ShapeViewerHelper.containerHeight`. 
That would help to keep `ShapeViewerHelper` scope focused solely on the automatic positioning of shapes.

- If the coordinates of a `Text` shape now represent the top-left corner, you might want to rename `textLayer.observation.center.x` and `textLayer.observation.center.y`, for exemple: 
    - `textLayer.observation.x`
    - `textLayer.observation.position.x`
    - `textLayer.observation.coordinates.x`
    - `textLayer.observation.topLeft.x`
    -  etc. 
    
    This change should only affect `TextLayer.qml`, since this shape doesn’t have any associated `ShapeAttribute`. 
    Might be best to do this now, before this shape gets widely used in JSON files.

PR #2933, opened 2025-10-31
[nodes] Add new `InputFile` input node

cbentejac:
## Description

This PR adds a new basic `InputNode` that receives a file or directory as the input, and that can be used as the entry point of a graph for `meshroom_batch`.
<div display="block" align="center">InputFile node</div>

### DIFF
--- a/meshroom/nodes/general/InputFile.py
+++ b/meshroom/nodes/general/InputFile.py
@@ -1,0 +1,48 @@
+__version__ = "1.0"
+
+import logging
+import os
+
+from meshroom.core import desc
+
+class InputFile(desc.InputNode, desc.InitNode):
+    """
+This node is an input node that receives a File.
+"""
+    category = "Other"
+
+    inputs = [
+        desc.File(
+            name="inputFile",
+            label="Input File",
+            description="A file or folder to use as the input.",
+            value="",
+        )
+    ]
+
+    def initialize(self, node, inputs, recursiveInputs):
+        self.resetAttributes(node, ["inputFile"])
+
+        if len(inputs) >= 1:
+            if os.path.isfile(inputs[0]) or os.path.isdir(inputs[0]):
+                self.setAttributes(node, {"inputFile": inputs[0]})
+
+                if len(inputs) > 1:
+                    logging.warning(f"Several inputs were provided ({inputs}).")
+                    logging.warning(f"Only the first one ({inputs[0]}) will be used.")
+            else:
+                raise RuntimeError(f"{inputs[0]} is not a valid file or directory.")
+
+        elif len(recursiveInputs) >= 1:
+            if os.path.isfile(recursiveInputs[0]) or os.path.isdir(recursiveInputs[0]):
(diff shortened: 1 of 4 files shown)


meshroom/nodes/general/InputFile.py
```diff
@@ -0,0 +1,55 @@
+__version__ = "1.0"
+
+import logging
+import os
+
+from meshroom.core import desc
+
+logger = logging.getLogger(__name__)
```
fabiencastan: not needed anymore.
directly use logging.info(xx)

meshroom/nodes/general/InputFile.py
```diff
@@ -0,0 +1,55 @@
...
+    ]
+
+    def initialize(self, node, inputs, recursiveInputs):
+        logger.setLevel(logging.INFO)
+        self.resetAttributes(node, ["inputFile"])
+
+        attributeSet = False
+        if len(inputs) >= 1:
+            for i in inputs:
+                if os.path.isfile(i) or os.path.isdir(i):
```
fabiencastan: Need a warning if it's not a file or dir.

meshroom/nodes/general/InputFile.py
```diff
@@ -0,0 +1,55 @@
...
+                    self.setAttributes(node, {"inputFile": i})
+                    attributeSet = True
+
+                    if len(inputs) > 1:
+                        logger.info(f"Several inputs were provided ({inputs}).")
```
fabiencastan: warning level

meshroom/nodes/general/InputFile.py
```diff
@@ -0,0 +1,55 @@
...
+                        logger.info(f"Only the first one ({i}) will be used.")
+
+                    break
+
+        if not attributeSet and len(recursiveInputs) >= 1:
```
fabiencastan: no need for "attributeSet", only need an if/elif/else.

meshroom/nodes/general/InputFile.py
cbentejac: Done

meshroom/nodes/general/InputFile.py
cbentejac: Done

meshroom/nodes/general/InputFile.py
cbentejac: Done

meshroom/nodes/general/InputFile.py
cbentejac: We can't rely on the `if/elif/else` only, because we might have cases where inputs are provided, but are all incorrect, which we won't know until we've entered the `if`. In those cases, the `elif/else` cases would be ignored, when we'd rather try our luck with the recursive inputs (if there are any).

fabiencastan: I would prefer to keep it simple/strict and have explicit errors.

cbentejac: Done

fabiencastan (approved):

PR #2946, opened 2025-12-04
[core] Logging: Fix several issues on the log files

### LINKED ISSUE #2807: [bug] I
**Describe the bug**
Squares in the Log view of the ImageDetectionPrompt node:

Image

**To Reproduce**
Run the node

**Expected behavior**
Probably a progress bar?

**Desktop (please complete the following and other pertinent information):**
 - OS: Win 10
 - Meshroom version: 2025.1 Beta

Alxiice:
## Description of the problem

We had several issues on the logs :
1. first when we had a `raise ValueError(...)` for example in the `processChunk` then on the log we only had the error and not the full log
2. also there was sometimes logs being overwritten and thus missing in the log file
3. there was additional `log` files (in addition to `0.log`) in parallelized nodes with 1 chunk
4. finally sometimes we had this kind of thing appearing in the log : 
image

## Why we had the issues and how this PR addresses them
- for (1) the reason is that the log file was opened in write mode and therefore changing it to append on `executeChunkCommandLine` fixes it
- for (2) the main reason is that we were systematically clearing the file when we call `chunk.logManager.start` 
- for (3) I made sure we never use `log`. Even on cases where we have a single chunk I think it's simply better to use `0.log` that way we have a better naming consistency anyways.
- for (4) I don't have a defenitive explanation but the logs were written in binary mode and for some reason the QML part was adding some unknown characters while writing thee log on the log tab. Changing the file opening mode to append is what fixed the issue

### DIFF
--- a/meshroom/core/node.py
+++ b/meshroom/core/node.py
@@ -516,8 +516,6 @@ class NodeChunk(BaseObject):
             self.node.saveOutputAttr()
             executionStatus = Status.SUCCESS
         except Exception:
-            # with open(self.logFile, "r") as fo:
-            #     print(f"logfile:<<<{fo.read()}>>>")
             self.updateStatusFromCache()  # check if the status has been updated by another process
             if self._status.status != Status.STOPPED:
                 executionStatus = Status.ERROR
--- a/meshroom/core/node.py
+++ b/meshroom/core/node.py
@@ -243,9 +243,13 @@ class LogManager:
             self.logger.addHandler(h)
         self.logger.setLevel(self._previousLevel)
 
-    def start(self, level):
-        # Clear log file
+    def clearLogFile(self):
         open(self.logFile, 'w').close()
+
+    def start(self, level):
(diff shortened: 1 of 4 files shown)


meshroom/core/node.py
```diff
@@ -514,11 +510,14 @@ def process(self, forceCompute=False, inCurrentEnv=False):
...
 
         try:
+            logging.info(f"[Process chunk] Start processing...")
             self.node.nodeDesc.processChunk(self)
             # NOTE: this assumes saving the output attributes for each chunk
             self.node.saveOutputAttr()
             executionStatus = Status.SUCCESS
         except Exception:
+            # with open(self.logFile, "r") as fo:
+            #     print(f"logfile:<<<{fo.read()}>>>")
```
cbentejac: To remove

cbentejac (changes requested):

cbentejac (approved):

PR #2947, opened 2025-12-04
Correctly catch syntax errors in node descriptions when reloading plugins

cbentejac:
## Description

This PR fixes an issue that occurred during the reload of nodes that had syntax errors in their description. This caused an uncaught `SyntaxError` exception that was raised during the automatic parsing of the file in `importlib.reload`, before reaching the validation step for the description (which ensures all the attributes have been correctly written, with the expected types and so on).

On the UI side, this meant that the "Reload All Plugins" feature would stop without any notice as soon as a syntax error was encountered. This translated by the "Reloading plugins..." message being sent but never followed by another message.

With the changes in the PR, the `SyntaxError` exception is correctly caught, and the node status set accordingly. This does not interrupt the reloading process for all the nodes anymore, and if an error is encountered, an error message is sent with the list of all the nodes that were not successfully reloaded.

Nodes that are not reloaded because there has been no change in their description since the last time they were (re)loaded are not included in the error list.

A corresponding unit test is also added.

### DIFF
--- a/meshroom/core/plugins.py
+++ b/meshroom/core/plugins.py
@@ -476,7 +476,12 @@ class NodePlugin(BaseObject):
                          f"at {self.path} has not been modified since the last load.")
             return False
 
-        updated = importlib.reload(sys.modules.get(self.nodeDescriptor.__module__))
+        try:
+            updated = importlib.reload(sys.modules.get(self.nodeDescriptor.__module__))
+        except Exception as exc:
+            logging.error(f"[Reload] {self.nodeDescriptor.__name__}: {exc} ({type(exc).__name__})")
+            self.status = NodePluginStatus.DESC_ERROR
+            return False
         descriptor = getattr(updated, self.nodeDescriptor.__name__)
 
         if not descriptor:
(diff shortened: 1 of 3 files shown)


meshroom/core/plugins.py
```diff
@@ -476,7 +476,12 @@ def reload(self) -> bool:
                          f"at {self.path} has not been modified since the last load.")
             return False
 
-        updated = importlib.reload(sys.modules.get(self.nodeDescriptor.__module__))
+        try:
+            updated = importlib.reload(sys.modules.get(self.nodeDescriptor.__module__))
+        except SyntaxError as exc:
```
Alxiice: It might not be enough to catch `SyntaxError`.
I tested with 
```py
class TestNode(desc.CommandLineNode):
    category = 'TestPlugin'
    commandLine = 'echo test'
    documentation = "a"+1
```
(in this case a `TypeError`)

Maybe the example is not the best but we should make sure to handle other exceptions too ?

cbentejac: Good catch! I had not thought of other exceptions. I extended the `except` clause to `Exception` and added the type of exception in the log.

Alxiice (approved):
looks good!

PR #2948, opened 2025-12-04
[ui] Use dedicated color for "fatal" and "critical" logs

cbentejac:
## Description

This PR introduces a specific color for logs that are of the "fatal" or "critical" level. Prior to this, only logs up to the "error" level were colored. Past it, the used color was the same as the "info" (and below) level, making it difficult to identify critical logs.

### DIFF
--- a/meshroom/ui/qml/Controls/TextFileViewer.qml
+++ b/meshroom/ui/qml/Controls/TextFileViewer.qml
@@ -285,8 +285,10 @@ Item {
                                     // Color line according to log level
                                     if (text.indexOf("[warning]") >= 0)
                                         return Colors.orange
-                                    else if(text.indexOf("[error]") >= 0)
+                                    else if (text.indexOf("[error]") >= 0)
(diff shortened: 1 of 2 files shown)


meshroom/ui/qml/Controls/TextFileViewer.qml
```diff
@@ -285,8 +285,10 @@ Item {
                                     // Color line according to log level
```
Alxiice, line 285:
```suggestion
                                    // Color line according to log level
                                    if (text.indexOf("[trace]") >= 0)
                                        return Qt.darker(palette.text, 2)
                                    if (text.indexOf("[debug]") >= 0)
                                        return Qt.darker(palette.text, 1.5)
```
I took the liberty to add this suggestion
Image

Alxiice (approved):
Small non blocking suggestion but I'm okay with the PR

cbentejac:
The suggestion has been applied, thank you!

PR #2968, opened 2025-12-26
Change Python version requirement to 3.9

JustACluelessKidAtSchool:
Specify exact Python version required for Windows and Linux. Python 3.9 is the only version compatible with all packages in requirements.txt and dev_requirements.txt.

### DIFF
--- a/INSTALL.md
+++ b/INSTALL.md
@@ -48,8 +48,8 @@ To use Meshroom nodal system without any visualization option, you can rely on a
 
 #### Python environment
 
-* Windows: Python 3 (>=3.9)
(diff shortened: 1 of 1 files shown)


JustACluelessKidAtSchool:
Just want to make sure I'm not the only one who needs specifically Python 3.9 to install the dependencies. Please respond as a sort-of sanity check before I set as ready for review.

JustACluelessKidAtSchool:
Just realized that it works with other Python versions if you aren't using dev_requirements.txt. Also the syntax requires Python>=3.10.

PR #2981, opened 2026-01-20
[ui] NodeActions: fix position on swiching status

Alxiice:
# Description of the PR

Fix a little but that occurred when we switched the status
fail
_failing node action position_
kindofworking
_-> fixed_

# Description of the bug

The event when the height is changed was not catched so what happened was :
- We launch the computation : upstream nodes are locked
- We select the node that is done, so all the buttons in the NodeActions are hidden because it is locked -> the height change to 0
- We select the computing node, buttons show up
  - The `updatePosition` is called on the delegate change and it is set before we update the height
  - The height is recomputed (to 33)
  - However we didn't have a `onHeightChanged` function therefore the NodeAction position is off

# Special point of interest during the review

I added the `onHeightChanged` which fixes the bug. However one little thing seems off to me :
- the `updatePosition` is called twice : one with `height=0` and a second time with `height=33`. We can see a small glitch because of these 2 updates. I tried to avoid that by using `Qt.callLater` but this didn't work. If you have an idea on something we could do, it would be great !

### DIFF
--- a/meshroom/ui/qml/Controls/NodeActions.qml
+++ b/meshroom/ui/qml/Controls/NodeActions.qml
@@ -106,6 +106,14 @@ Item {
             function onScaleChanged() { Qt.callLater(actionHeader.updatePosition) }
         }
 
+        // Update position when nodes are moved
+        Connections {
+            target: actionHeader.selectedNodeDelegate
+            function onXChanged() { actionHeader.updatePosition() }
+            function onYChanged() { actionHeader.updatePosition() }
+            ignoreUnknownSignals: true
+        }
+
         // 
         // ===== Manage buttons =====
         // 
--- a/meshroom/ui/qml/Controls/NodeActions.qml
+++ b/meshroom/ui/qml/Controls/NodeActions.qml
@@ -90,24 +90,20 @@ Item {
             }
         }
 
+        onHeightChanged: {
+            Qt.callLater(actionHeader.updatePosition)
+        }
+
         onWidthChanged: {
-            updatePosition()
+            Qt.callLater(actionHeader.updatePosition)
         }
 
         // Update position when the user moves on the graph
         Connections {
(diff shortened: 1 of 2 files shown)


meshroom/ui/qml/Controls/NodeActions.qml
```diff
@@ -90,16 +90,20 @@ Item {
             }
         }
 
+        onHeightChanged: {
+            Qt.callLater(actionHeader.updatePosition)
```
cbentejac:
```suggestion
            actionHeader.updatePosition()
```

meshroom/ui/qml/Controls/NodeActions.qml
```diff
@@ -90,16 +90,20 @@ Item {
...
+        }
+
         onWidthChanged: {
-            updatePosition()
+            Qt.callLater(actionHeader.updatePosition)
```
cbentejac:
```suggestion
            actionHeader.updatePosition()
```

meshroom/ui/qml/Controls/NodeActions.qml
```diff
@@ -90,16 +90,20 @@ Item {
             }
         }
```
cbentejac, line 104: Before the `if (!selectedNodeDelegate || !draggable) return`, I would add:
```

if (width == 0 && height == 0) {
    actionItemsRow.visible = true
    return
} else if (width == 0 || height == 0) {
    actionItemsRow.visible = false
    return
}
actionItemsRow.visible = true

if (!selectedNodeDelegate || !draggable) return
...
```

cbentejac (changes requested):
This needs to be tested but I feel like the suggested changes are solving the graphical update issues, and with them we see only one update instead of two when selecting another node.

cbentejac (approved):

PR #2987, opened 2026-01-27
[ui] Add copy path feature on the homepage

Alxiice:
# Description
Add a "Copy path" menu button on the homepage.
Useful to retrieve scene paths
copy_path

### DIFF
--- a/meshroom/ui/qml/Homepage.qml
+++ b/meshroom/ui/qml/Homepage.qml
@@ -387,6 +387,14 @@ Page {
                                 }
 
                                 MenuItem {
+                                    text: "Copy path"
+                                    onTriggered: {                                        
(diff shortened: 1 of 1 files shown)


meshroom/ui/qml/Homepage.qml
```diff
@@ -386,6 +386,14 @@ Page {
                                     }
                                 }
 
+                                MenuItem {
+                                    text: "Copy path"
```
cbentejac:
```suggestion
                                    text: "Copy Path"
```

cbentejac (changes requested):
Words should always be capitalized in such menus.

cbentejac (approved):

PR #2990, opened 2026-02-06
[core] Make `BackdropNodes` effectively uncomputable

cbentejac:
## Description

This PR fixes a bug introduced by 378b91323db9045f5622cf8e3caee8651a5daccc, which excludes non-computable nodes from the list of nodes to execute, but fails to check before-hand that the provided list of nodes is not  `None`, thus causing issues when launching the computation of the whole graph through Meshroom header's "Compute" button.

The goal of 378b91323db9045f5622cf8e3caee8651a5daccc was to prevent attempting to compute `BackdropNodes` by removing them from the list of nodes to compute before actually starting the execution process. This PR addresses this by making the `BackdropNode` core object uncomputable: chunks will never be created for it, and unlike `InputNodes` which are not computable but still need to be part of the evaluated graph dependencies, the `BackdropNodes` are removed altogether from the graph evaluation in the context of its execution (they will thus not appear in the Task Manager, for example).

Additionally, a commit is pushed to prevent attempting to compute `InputNodes` in `meshroom_compute`: such nodes were correctly detected, but instead of stopping there, the process was still going through. This would for example cause errors on `InputNodes` sent to the farm.

### DIFF
--- a/meshroom/core/taskManager.py
+++ b/meshroom/core/taskManager.py
@@ -435,7 +435,7 @@ class TaskManager(BaseObject):
         computed = []
         inputNodes = []
         for node in toNodes:
-            if not node.isComputableType:
+            if node.isInputNode:
                 inputNodes.append(node)
             elif context == "COMPUTATION":
                 if graph.canComputeTopologically(node) and graph.canSubmitOrCompute(node) % 2 == 1:
--- a/meshroom/core/taskManager.py
+++ b/meshroom/core/taskManager.py
@@ -283,6 +283,7 @@ class TaskManager(BaseObject):
             if not toNodes:
                 toNodes = graph.getLeafNodes(dependenciesOnly=True)
             toNodes = list(toNodes)
+            toNodes = [node for node in toNodes if not node.isBackdropNode]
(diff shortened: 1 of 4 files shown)


meshroom/core/taskManager.py
```diff
@@ -283,6 +283,7 @@ def compute(self, graph: Graph = None, toNodes: list[Node] = None, forceCompute:
             if not toNodes:
                 toNodes = graph.getLeafNodes(dependenciesOnly=True)
             toNodes = list(toNodes)
+            toNodes = [node for node in toNodes if not node.isBackdropNode]
```
Alxiice, line 286: Maybe this ?
```suggestion
            toNodes = [node for node in toNodes if not node.isComputableType]
```

Alxiice (approved):

meshroom/core/taskManager.py
cbentejac: `isComputableType` would exclude `InputNodes`, but we want to keep them in the computed graph. When submitting the graph, it makes no sense to have the `BackdropNodes` being executed, but it does to have the `InputNodes`, although they are not computable.

PR #3000, opened 2026-02-13
Enable connecting multiple list Attributes

servantftransperfect:
Enable the connection of multiple listAttributes to the same destination listAttribute

### DIFF
--- a/meshroom/core/graph.py
+++ b/meshroom/core/graph.py
@@ -954,6 +954,73 @@ class Graph(BaseObject):
         return [edge.src, edge.dst], deletedEdge
 
     @changeTopology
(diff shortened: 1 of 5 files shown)


cbentejac:
This seems to work correctly, including when performing undos and redos, but some unit tests should definitely be added.

fabiencastan (changes requested):
In my opinion, it is not a good idea to create connections between lists and, in some cases, between elements within the lists, without providing the end user with any information. This leads to errors and confusion.

fabiencastan:
In the short term, I would recommend to create a node with a list<list<file>> in input that creates a flatten list<files> in output.

fabiencastan:
https://github.com/alicevision/Meshroom/pull/3061

PR #3017, opened 2026-02-25
[bug] Dynamic node size is not correctly transmitted

Alxiice:
## Description

Dynamic node size is not correctly transmitted.

## What was the issue 

Example with 3 nodes (`A - B - C`):
- A : dynamic size that will be set to 2
- B : no parallelization (so 1 chunk) but inherit previous node size
- C : parallelized, inherit previous node size

When we launch the process :
1. A size is 0 by default
2. B size is set to A size so 0
3. C size is set to B size so 0
4. We launch the process
5. A expands and 2 chunks are created
6. B processing starts, still 0 chunk
7. C processing starts, still 0 chunks

## Changes

Explicitely call a function that will set the correct size before processing the node.

Now here would be the modified steps if we reuse the same example :
6. B processing starts, we set the size again so B size becomes 2
7. C processing starts, we set the size again so C size becomes 2

### DIFF
--- a/meshroom/core/node.py
+++ b/meshroom/core/node.py
@@ -2259,9 +2259,6 @@ class Node(BaseNode):
             'internalInputs': {k: v for k, v in internalInputs.items() if v is not None},
             'outputs': outputs,
         }
-        
-    def updateNodeSize(self):
-        self.setSize(self.nodeDesc.size.computeSize(self))
 
     def _resetChunks(self):
         """ Set chunks on the node.
@@ -2282,12 +2279,12 @@ class Node(BaseNode):
         self._nodeStatus.status = Status.NONE
         # Recreate list with reset values (1 chunk or the static size)
         if not self.isParallelized:
-            self.setSize(self.nodeDesc.size.computeSize(self))
+            self._updateNodeSize()
             self._chunks.setObjectList([NodeChunk(self, desc.Range())])
             self._chunks[0].statusChanged.connect(self.globalStatusChanged)
             self._chunksCreated = True
         elif isinstance(self.nodeDesc.size, desc.computation.StaticNodeSize):
-            self.setSize(self.nodeDesc.size.computeSize(self))
+            self._updateNodeSize()
             self._chunks.setObjectList([NodeChunk(self, desc.Range())])
             self._chunks[0].statusChanged.connect(self.globalStatusChanged)
             self._chunksCreated = True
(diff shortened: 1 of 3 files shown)


meshroom/core/node.py
```diff
@@ -2259,6 +2259,9 @@ def toDict(self):
             'internalInputs': {k: v for k, v in internalInputs.items() if v is not None},
             'outputs': outputs,
         }
+        
+    def updateNodeSize(self):
+        self.setSize(self.nodeDesc.size.computeSize(self))
```
fabiencastan: Replace the other places where we use `self.setSize(self.nodeDesc.size.computeSize(self))` to use updateNodeSize().

meshroom/core/node.py
```suggestion
            self.updateNodeSize()
```

```suggestion
            self.updateNodeSize()
```

meshroom/core/taskManager.py
fabiencastan:
```suggestion
            if node._chunksCreated:
```
Does not make sense to check attributes. It's initialized in the constructor.
Same in multiple places.

tests/test_compute.py
```diff
@@ -413,3 +413,92 @@ def test_noUpstreamLockAfterParameterChange(self, graphSavedOnDisk):
...
+            description='Output',
+            value=os.path.join("{nodeCacheFolder}"),
+            group='',
+        ),
+    ]
+    def processChunk(self, chunk):
+        pass
+
+class TestNode_SizeB(TestNode_SizeA):
+    """ Inherit the linked node size but not parallelized """
```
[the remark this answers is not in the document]
Alxiice: it inherits testnode_sizeA

tests/test_compute.py
```diff
@@ -413,3 +413,92 @@ def test_noUpstreamLockAfterParameterChange(self, graphSavedOnDisk):
...
+        assert node.globalStatus == status.name
+
+    def test_correctSizeUpdate(self, graphSavedOnDisk):
+        graph: Graph = graphSavedOnDisk
+        nodeA = graph.addNewNode("TestNode_SizeA")
+        nodeB = graph.addNewNode("TestNode_SizeB")
+        nodeA.output.connectTo(nodeB.nodeInput)
+        nodeC = graph.addNewNode("TestNode_SizeC")
+        nodeB.output.connectTo(nodeC.nodeInput)
+        graph.save()
```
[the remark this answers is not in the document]
Alxiice: yeah but we don't have tests for the taskmanager yet as it's a UI process

fabiencastan (approved):

PR #3026, opened 2026-03-05
Enable lambda with node parameter in attributes 

servantftransperfect:
This pull request introduces a new helper method to improve how callables are executed for attribute values in `meshroom/core/attribute.py`. The main goal is to make callable evaluation more flexible by automatically determining whether to pass the attribute or its node as an argument, which helps with legacy code compatibility.

Before :

`lambda attr : attr.node.xxx`

After

`lambda node : node.xxx`

**Callable evaluation improvements:**

* Added the `executeValue` method to the attribute class, which inspects the callable's signature and decides whether to pass the attribute instance or its node when executing. This makes the code more robust and future-proof for legacy callable handling.
* Refactored `_setValue` and `getDefaultValue` methods to use the new `executeValue` helper, ensuring consistent callable execution throughout the class. [[1]](diffhunk://#diff-346bf2284498d029adfca15a85216268454d6143e70541993cd79f5ea9c8076eL187-R208) [[2]](diffhunk://#diff-346bf2284498d029adfca15a85216268454d6143e70541993cd79f5ea9c8076eL295-R316)

**Imports update:**

* Imported `inspect` to support signature analysis in the new `executeValue` method.

### DIFF
--- a/meshroom/core/attribute.py
+++ b/meshroom/core/attribute.py
@@ -6,6 +6,7 @@ import os
 import re
 import weakref
 import logging
+import inspect
 
 from collections.abc import Iterable, Sequence
 from string import Template
@@ -123,6 +124,25 @@ class Attribute(BaseObject):
         if self.node:
             self.node.updateInternalAttributes()
 
+    def executeValue(self, value):
+        """
+        Assume value is a callable
+        Analyze value signature to detect if we want to use node or attr as parameter.
+        This method may be removed when all the legacy code is transformed.
+
+        Args:
+            value (Callable): the callable to execute
+
+        Return the result value of the callable
+        """
+        # The new behavior is to provide the node to the callable.
+        # For compatibility with the old behavior providing the attribute, we check if the attribute is named "attr" and provide the attribute.
+        params = inspect.signature(value).parameters
+        if len(params) == 1 and list(params)[0] == "attr":
+            return value(self)
+        
+        return value(self.node)
+
     def _initValue(self):
         """
         Initialize the attribute value.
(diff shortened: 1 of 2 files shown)


Alxiice:
Maybe it would be better to put the tests in `test_attributes.py` ?

meshroom/core/attribute.py
```diff
@@ -123,6 +124,26 @@ def requestNodeUpdate(self):
...
+            value (Callable): the callable to execute
+
+        Return the result value of the callable
+        """
+        try:
+            params = inspect.signature(value).parameters
+            if len(params) == 1 and list(params)[0] == "node":
+                return value(self.node)
+            else:
+                return value(self)
```
fabiencastan:
```suggestion
            # The new behavior is to provide the node to the callable.
            # For compatibility with the old behavior providing the attribute, we check if the attribute is named "attr" and provide the attribute.
            params = inspect.signature(value).parameters
            if len(params) == 1 and list(params)[0] == "attr":
                return value(self)
            return value(self.node)
```

If someone is creating a `lambda n: n.foo.value` or any name (other than attr), we expect Meshroom to use the new behavior.

tests/test_attributeLambda.py
```diff
@@ -0,0 +1,134 @@
...
+            name="callableNodeInput",
+            label="Callable Node Input",
+            description="Input with a callable default that receives the node.",
+            value=lambda node: node.fixedInput.value * 2,
+            range=(0, 200, 1),
+        ),
+        desc.StringParam(
+            name="callableAttrInput",
+            label="Callable Attr Input",
+            description="Input with a callable default that receives the attribute.",
```
fabiencastan:
```suggestion
            label="Callable Attr Input (Compatibility)",
            description="Input with a callable default that receives the attribute for compatibility with the old behavior.",
```

meshroom/core/attribute.py
```diff
@@ -123,6 +124,26 @@ def requestNodeUpdate(self):
...
+        except Exception:
+            return value(self)
```
fabiencastan: No need for try/except IMO. It will generate duplicate calls in case of a real issue within the lambda and create more confusion than help.

fabiencastan (approved):

PR #3027, opened 2026-03-05
Replacing group by commandLineGroup

servantftransperfect:
Rename attribute parameter "group" to "commandLineGroup". 

Keep temporary compatibility with old naming

Before :
```
desc.ChoiceParam(
            name="describerPreset",
            label="Describer Density",
            description="Control the ImageDescriber density (low, medium, normal, high, ultra).\n"
                        "Warning: Use ULTRA only on small datasets.",
            value="normal",
            values=["low", "medium", "normal", "high", "ultra", "custom"],
            group=lambda node: 'allParams' if node.describerPreset.value != 'custom' else None,
        ),
```

After:
```
desc.ChoiceParam(
            name="describerPreset",
            label="Describer Density",
            description="Control the ImageDescriber density (low, medium, normal, high, ultra).\n"
                        "Warning: Use ULTRA only on small datasets.",
            value="normal",
            values=["low", "medium", "normal", "high", "ultra", "custom"],
            commandLineGroup=lambda node: 'allParams' if node.describerPreset.value != 'custom' else None,
        ),
```

### DIFF
--- a/meshroom/core/desc/attribute.py
+++ b/meshroom/core/desc/attribute.py
@@ -3,7 +3,7 @@ import os
 from collections.abc import Iterable
 from enum import auto, Enum
 
-from meshroom.common import BaseObject, JSValue, Property, Variant, VariantList, strtobool
+from meshroom.common import BaseObject, JSValue, Property, Variant, VariantList, strtobool, deprecated
 
 class ValueTypeErrors(Enum):
     NONE = auto()  # No error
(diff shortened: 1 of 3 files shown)


meshroom/core/desc/attribute.py
```diff
@@ -134,15 +140,17 @@ def matchDescription(self, value, strict=True):
 
 class ListAttribute(Attribute):
     """ A list of Attributes """
-    def __init__(self, elementDesc, name, label, description, group="allParams", advanced=False, semantic="",
-                 enabled=True, joinChar=" ", visible=True, exposed=False):
+    def __init__(self, elementDesc, name, label, description, group="allParams", commandLineGroup=_legacySentinel,
```
fabiencastan: I would expect:
`group=_legacySentinel, commandLineGroup="allParams",`

Alxiice:
Could you add depreciation warnings when we use `group` instead of `commandLineGroup` ? Thanks !

meshroom/core/desc/attribute.py
servantftransperfect: I renamed _legacySentinel to _setParamSentinel

Alxiice (approved):

PR #3065, opened 2026-04-01
[core] Force setgid permissions on folders

Alxiice:
# Description

When created the folders set the setgid permission bit


Alxiice:
I'm closing this PR. We will let the OS set default permissions for the folders

PR #3070, opened 2026-04-03
[meshroom_batch] Invalidation string on scene creation

Alxiice:
# Description

- New parameter on `meshroom_batch` : we can now set an invalidation string on all nodes
- Mini fix on a log

### DIFF
--- a/bin/meshroom_batch
+++ b/bin/meshroom_batch
@@ -126,6 +126,10 @@ advanced_group.add_argument(
     help='Override specific parameters directly from the command line (by node type or by node names).')
 
 advanced_group.add_argument(
+    '--setInvalidationString', metavar='STRING', type=str, 
+    help='Set an invalidation string on ALL nodes.')
+
+advanced_group.add_argument(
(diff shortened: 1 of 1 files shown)


bin/meshroom_batch
```diff
@@ -172,6 +176,13 @@ with meshroom.core.graph.GraphModification(graph):
         # custom pipeline
         graph.initFromTemplate(args.pipeline, copyOutputs=True if args.output else False)
 
+    # Set invalidation string
+    if invalidationString:=args.setInvalidationString:
+        nodes, _ = graph.dfsOnFinish()
+        for node in nodes:
+            if node.isComputableType:
+                node.setInternalAttributeValues({"invalidation": invalidationString})
```
fabiencastan:
```suggestion
    if args.setInvalidationString:
        nodes, _ = graph.dfsOnFinish()
        for node in nodes:
            if node.isComputableType:
                node.setInternalAttributeValues({"invalidation": args.setInvalidationString})
```

fabiencastan (approved):

PR #3072, opened 2026-04-07
Add utilities nodes in Meshroom

Alxiice:
# Description

Add new general nodes :
- `InputString`
- `InputInt`
- `GetParentFolder` : get the parent folder of a path
- `GenerateMeshroomScene` : generate a meshroom scene based on a template and node overrides
- `MeshroomSceneParameter` : utility tool to create overrides strings used on the `GenerateMeshroomScene` node
- `ComputeMeshroomScene` : launch the compute/submit on a meshroom scene

# How to use these nodes

I'll skip the InputString, InputInt & GetParentFolder that are pretty straightforward.

For the other nodes here is a pipeline example : 

image

Here is the node breakdown :

- **MeshroomSceneParameter :** <br/>image
  - `Node` is the node instance name (or node type) to override
  - `Parameter` is the parameter name. For input overrides you must leave this empty and for parameter overrides you must fill it.
  - `Value` is the value it will be set to. If empty the output will be set to an empty string to avoid errors.
  - `Mode` is either `node_instance` (override only one node instance) or `node_type` (override all nodes from a specific type).

- **GenerateMeshroomScene :** <br/>image
  - `Template` is the template scene
  - `Scene Path` is the scene destination. If empty it will be a `scene.mg` file in the node internal folder
  - `Input Overrides` are the overrides for input nodes (CameraInit)
  - `Parameters Overrides` are the overrides for other nodes
  - `Invalidation String` can be used to set an invalidation string on all nodes (to make sure the graph is unique)
  - `Meshroom Scene` will be set to the created scene file

- **ComputeMeshroomScene :** <br/>image
  - `Scene` is the path to the meshroom scene file 
  - `Force Compute` can be set to True to compute nodes even if their status is success.
  - `Submit` can be used to submit on farm (in opposition to compute locally)
  - `Submitter` can be used to select the submitter
  - `Submit Label` can be used to set the submitted job name


meshroom/nodes/general/ComputeMeshroomScene.py
```diff
@@ -0,0 +1,93 @@
...
+PYTHON_EXE = "python"
+
+
+class ComputeMeshroomScene(desc.CommandLineNode):
+    """
+    Compute or Submits a meshroom scene on the farm.
+    """
+
+    category = "Utils"
+    commandLine = f"{PYTHON_EXE} {str(_MESHROOM_BATCH)}" + " -p {node.scene.value} --save {node.scene.value}"
```
cbentejac: Why not directly 
```suggestion
    commandLine = f"{node.nodeDesc.pythonExecutable} {str(_MESHROOM_BATCH)}" + " -p {node.scene.value} --save {node.scene.value}"
```
since `pythonExecutable` is set to `python` by default for all `BaseNodes`?

(this comment applies to all the other nodes of this PR as well)

meshroom/nodes/general/GenerateMgScene.py
```diff
@@ -0,0 +1,224 @@
+# -*- coding: utf-8 -*-
+
+__version__ = "1.1"
```
cbentejac: Since the node is added in this PR, shouldn't it be
```suggestion
__version__ = "1.0"
```
?

meshroom/nodes/general/ComputeMeshroomScene.py
```diff
@@ -0,0 +1,93 @@
...
+                cmd += f" --submitLabel \"{node.submitLabel.value}\""
+        else:
+            cmd += " --compute yes"
+        return cmd
+
+    inputs = [
+        desc.File(
+            name="scene",
+            label="Scene",
+            description="Meshroom scene",
```
cbentejac:
```suggestion
            description="Meshroom scene.",
```

meshroom/nodes/general/ComputeMeshroomScene.py
```diff
@@ -0,0 +1,93 @@
...
+            enabled=lambda node: node.submit.value is True
+        ),
+        desc.StringParam(
+            name="submitLabel",
+            label="Submit Label",
+            description=(
+                "The label that will be set at the submitted job name.\n"
+                "An empty string will set a default string : '[Meshroom] {projectName}'.\n"
+                "You can put the following strings under brackets and they will be replaced :\n"
+                "- projectName: the name of the scene file"
```
cbentejac:
```suggestion
                "The label that will be set for the submitted job name.\n"
                "An empty string will set a default string: '[Meshroom] {projectName}'.\n"
                "The following strings between brackets can be used as they will be automatically replaced:\n"
                "- projectName: the name of the scene file"
```
It's generally better to use indirect phrasing for the attribute descriptions.

meshroom/nodes/general/GenerateMgScene.py
```diff
@@ -0,0 +1,224 @@
...
+
+
+
+class MeshroomSceneParameter(desc.Node):
+    """ Build a parameter/input override.
+
+There are 2 modes of overrides :
+- **node_instance** mode (`NODEINSTANCE.param=value`) : only one node instance is overriden
+- **node_type** mode (`NODETYPE:param=value`) : all nodes of the type are overrided
+    """
```
cbentejac:
```suggestion
    """ Build a parameter/input override.

There are 2 modes of overrides:
- **node_instance** mode (`NODEINSTANCE.param=value`): only one node instance (identified by its name) is overridden
- **node_type** mode (`NODETYPE:param=value`): all nodes of the given type are overridden
    """
```

meshroom/nodes/general/ComputeMeshroomScene.py
```diff
@@ -0,0 +1,93 @@
...
+            description=(
+                "Set True to force compute. If nodes are already computed, the status will"
+                "be reset to None and the cache will be deleted."
+            ),
+            value=True,
+        ),
+        desc.BoolParam(
+            name="submit",
+            label="Submit",
+            description="Set True to submit, False to compute locally",
```
cbentejac:
```suggestion
            description="Set True to submit, False to compute locally.",
```

meshroom/nodes/general/GenerateMgScene.py
```diff
@@ -0,0 +1,224 @@
...
+            name="nodeName",
+            label="Node",
+            description="Node instance name or node type.",
+            value="",
+            exposed=True,
+        ),
+        desc.StringParam(
+            name="paramName",
+            label="Parameter",
+            description="Parameter name",
```
cbentejac:
```suggestion
            description="Parameter name.",
```

Also there's an ambiguity between the naming of "parameters" and "attributes". Since we tend to refer to attributes across the code, I'd be partial to referring to "attribute" (`attrName`, `attrValue`) only, but it's up for discussion (especially considering that in `meshroom_batch`, the option refers to `param` rather than `attribute`).

meshroom/nodes/general/GenerateMgScene.py
```diff
@@ -0,0 +1,224 @@
...
+            value="",
+            exposed=True,
+        ),
+        desc.StringParam(
+            name="paramValue",
+            label="Value",
+            description="",
```
cbentejac: Even if the name of the attribute is straightforward, we should not leave `description` fields empty.

meshroom/nodes/general/GenerateMgScene.py
```diff
@@ -0,0 +1,224 @@
...
+            value="",
+            exposed=True,
+        ),
+        desc.ChoiceParam(
+            name="mode",
+            label="Mode",
+            description=(
+                "Override modes :\n"
+                "- node_instance: Override the node instance\n"
+                "- node_type: Override all nodes having this type"
```
cbentejac:
```suggestion
                "Override modes:\n"
                "- node_instance: Override the node instance\n"
                "- node_type: Override all nodes having this type"
```

meshroom/nodes/general/GenerateMgScene.py
```diff
@@ -0,0 +1,224 @@
...
+        inputOverrides = self.get_overrides(node.inputOverrides)
+        paramOverrides = self.get_overrides(node.paramOverrides)
+        sceneDestination = str(node.sceneDestination.getValueStr(withQuotes=False))
+        if sceneDestination:
+            sceneDestination = Path(sceneDestination)
+        else:
+            sceneDestination = Path(node.internalFolder) / "scene.mg"
+
+        logging.info(f"- Using template scene : {templateScene}")
+        logging.info(f"- Scene destination : {sceneDestination}")
```
cbentejac:
```suggestion
        logging.info(f"- Using template scene: {templateScene}")
        logging.info(f"- Scene destination: {sceneDestination}")
```

meshroom/nodes/general/GenerateMgScene.py
```diff
@@ -0,0 +1,224 @@
...
+
+        if inputOverrides or paramOverrides:
+            logging.info(f"{'='*10} Scene overrides {'='*10}")
+            for item in inputOverrides:
+                logging.info(f"- Override input : {item}")
+            for item in paramOverrides:
+                logging.info(f"- Override parameter : {item}")
```
cbentejac:
```suggestion
                logging.info(f"- Override input: {item}")
            for item in paramOverrides:
                logging.info(f"- Override parameter: {item}")
```

meshroom/nodes/general/GenerateMgScene.py
```diff
@@ -0,0 +1,224 @@
...
+        for override in overridesList:
+            overrideValue = override.value
+            if overrideValue:
+                overrides.append(overrideValue)
+        return overrides
+
+    def process(self, node):
+        templateScene = node.templatePath.getValueStr(withQuotes=False)
+        if not templateScene or not Path(templateScene).exists():
+            raise ValueError(f"{node} Invalid template scene : {templateScene}")
```
cbentejac:
```suggestion
            raise ValueError(f"{node} Invalid template scene: {templateScene}")
```

meshroom/nodes/general/GenerateMgScene.py
```diff
@@ -0,0 +1,224 @@
...
+        if inputOverrides or paramOverrides:
+            logging.info(f"{'='*10} Scene overrides {'='*10}")
+            for item in inputOverrides:
+                logging.info(f"- Override input : {item}")
+            for item in paramOverrides:
+                logging.info(f"- Override parameter : {item}")
+
+        sceneRoot = sceneDestination.parent
+        if not sceneRoot.exists():
+            logging.info(f"Creating parent folder : {sceneRoot}")
```
cbentejac:
```suggestion
            logging.info(f"Creating parent folder: {sceneRoot}")
```

meshroom/nodes/general/GenerateMeshroomScene.py
cbentejac, line 1: Also up to discussion, but I think it's preferable to have each node in its own file, for two main reasons:
- it allows a better granularity when it comes to node versions: updating one node in the file doesn't cause the version update of all the others.
- from a development standpoint, I feel like it is far easier if we expect to find a .py file with the name of the node in the `nodes/general` rather than having to search for it in another file.
Overall it also makes it easier to track a node's evolution over time if it is self-contained in its own file.

cbentejac (changes requested):

meshroom/nodes/general/GenerateMgScene.py
Alxiice: Yeah because I moved it from another package where the version was 1.0

cbentejac (approved):

PR #3075, opened 2026-04-09
Fix issue where computation starts before the chunks are created

cbentejac:
## Description

This PR improves the handling and synchronization of chunk setup information in the node status logic to prevent stale or invalid chunk data from being written or reloaded. The changes ensure that chunk-related fields are only present in the node status when valid, and that stale chunk information is cleared in key workflow steps to avoid errors or inconsistencies during reloads and computation.

This essentially fixes an issue that might occur on farm where tasks are created before the chunks have effectively been created, leading to the following error: 
> Error: Computing chunk x of node Y before chunks have been created.

This issue is fixed by ensuring that chunk information in the status file is always synchronized with the actual chunks before writing to disk; this prevents mismatches after cache reloads or chunk resets, and having a `nodeStatus` file that does not contain any chunk-related keys while the `.status` files for the chunks have been written on disk becomes impossible.

## Implementation remarks

In addition to the fix itself, this PR includes:
- a fallback in `meshroom_compute` that aims at pushing against filesystems propagation delays: on a render farm, there may be cases where the `nodeStatus` file has been properly updated but its changes are not immediately visible. If the `nodeStatus` file does not contain any chunk information whereas we know that chunks should have been created, we attempt to reload it every 2s (up to 10s). If the chunk information are found between the timeout, the process goes on without a hitch; otherwise, we exit with an error.
- Chunk information are only serialized if valid chunks exist, and only deserialized if they do not create an invalid chunk setup.
- Stale chunk information are cleared from the `nodeStatus` file before submitting or computing, ensuring up-to-date chunk parameters are used and preventing the re-creation of chunks from outdated cache data.
-  `updateStatusFromCache` is improved to recover from missed chunk creation by checking both for chunk setup changes and for cases where chunks haven't been created yet, ensuring chunks are recreated from cache when necessary.

### DIFF
--- a/bin/meshroom_compute
+++ b/bin/meshroom_compute
@@ -103,9 +103,29 @@ if args.node:
         submittedStatuses.append(Status.SUBMITTED)
 
     if args.iteration >= 0 and not node._chunksCreated:
-        print(f"Error: Computing chunk {args.iteration} of node {node} before chunks have been created. " \
-              f"See file: \"{node.nodeStatusFile}\".")
-        sys.exit(-1)
+        # On a render farm, the nodeStatus file written by meshroom_createChunks may not be
+        # immediately visible on this blade due to NFS propagation delays.
+        # Wait for the file to appear and retry loading before giving up.
+        import time
+        retryInterval = 2    # seconds between retries
+        maxRetryTime = 10    # total seconds before giving up
+        elapsed = 0
+        while not node._chunksCreated and elapsed < maxRetryTime:
+            logging.warning(
+                f"Chunks not created from cache for node {node.name} "
+                f"(nodeStatusFile: \"{node.nodeStatusFile}\"). "
+                f"Retrying in {retryInterval}s ({elapsed}/{maxRetryTime}s elapsed)..."
+            )
+            time.sleep(retryInterval)
+            elapsed += retryInterval
+            node.updateStatusFromCache()
+
+        if not node._chunksCreated:
+            logging.error(
+                f"Computing chunk {args.iteration} of node {node} but chunks have not been "
+                f"created after waiting {maxRetryTime}s. See file: \"{node.nodeStatusFile}\"."
+            )
+            sys.exit(-1)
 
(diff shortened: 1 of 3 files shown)


meshroom/core/node.py
```diff
@@ -161,7 +167,8 @@ def fromDict(self, d):
             blockSize = int(d.pop("chunksBlockSize") or 0)
             fullSize = int(d.pop("chunksFullSize") or 0)
             nbBlocks = int(d.pop("chunksNbBlocks") or 0)
-            self.chunks = NodeChunkSetup(blockSize, fullSize, nbBlocks)
+            if nbBlocks > 0:
+                self.chunks = NodeChunkSetup(blockSize, fullSize, nbBlocks)
```
fabiencastan, line 171: We should set self.chunks to something else its value depends on a random precedent state.

servantftransperfect (approved):
Behavior is ok

Alxiice (approved):

meshroom/core/node.py
fabiencastan: Done by the reset() function.

bin/meshroom_compute
```diff
@@ -103,9 +103,26 @@ if args.node:
...
+            try:
+                node.createChunks()
+            except Exception as e:
+                logging.error(f"Recovery failed — could not create chunks: {e}")
+            if not node._chunksCreated:
+                print(
+                    f"Error: Computing chunk {args.iteration} of node {node} before "
+                    f"chunks have been created. See file: \"{node.nodeStatusFile}\"."
+                )
+                sys.exit(-1)
```
fabiencastan: Will just hide the problem instead of fixing it.

meshroom/core/node.py
cbentejac: Done

bin/meshroom_compute
cbentejac: Removed

PR #3077, opened 2026-04-14
[graph] Add possibility to set a specific cacheDir in the mg scene

Alxiice:
# Description

Add possibility to set a specific cache folder in the mg scene
- In `meshroom_batch` a new option enables saving the scene with a specific cache folder.
- a `_explicitCacheDir` parameter in the `Graph` tracks the explicit path we set (can be absolute or relative)
- functions were added to set the default cache path or a custom one

About the Meshroom file :
- `graphIO` : a `CacheDir` key was added to the serialized format. Serialization is designed to save the cacheDir inside the `.mg` file
- `graph` : the deserialization is designed to load the explicit cache folder if we get one, and if we don't to select the default path (`./MeshroomCache`)
- the serialized cacheDir can be a list, and in this case when deserialized we will iterate over elements and take the first one whose folder exists.

We can see the custom cache directory on the status bar :
image

# Tests

Here is a command that can be used to set a new cache directory in a meshroom file
```sh
ROOT="..."
MG_FILE_IN="$ROOT/scene.mg"
MG_FILE_OUT="$MG_FILE_IN"
CACHE_DIR="$ROOT/MeshroomCache-v001"
meshroom_batch -p $MG_FILE_IN -s $MG_FILE_OUT --compute no --overrideCacheDir $CACHE_DIR
```

The following steps can test this feature :
1. Create a new meshroom scene
2. Compute the graph
3. Close it
4. Follow above script to set a new custom cache
5. Open the scene
6. You should see that the nodes are considered not computed (because we changed the cache to an empty folder). The new cache folder is set and can be seen on the status bar (at the bottom)
7. Empty the previous cache folder (`MeshroomCache`)
8. Launch the process on the current (new) scene
9. It doesn't affect the original cache folder but the new cache is filled with the current scene nodes computation data

### DIFF
--- a/bin/meshroom_batch
+++ b/bin/meshroom_batch
@@ -4,7 +4,6 @@ import json
 import logging
 import os
 import re
-import sys
 
 import meshroom.core.graph
 from meshroom.common import strtobool
@@ -122,6 +121,15 @@ advanced_group.add_argument(
     help='A JSON file containing the graph parameters override.')
 
 advanced_group.add_argument(
+    '--overrideCacheDir', metavar='STRING', type=str, default=None,
+    help=(
+        'Override the cache location for the scene. If the scene is saved, the cache will be '
+        'set as the selected location. By default the cache location will be set as a folder '
+        'named MeshroomCavhe at the project file root folder.'
+    )
+)
+
+advanced_group.add_argument(
     '--paramOverrides', metavar='NODETYPE:param=value NODEINSTANCE.param=value', type=str, default=None, nargs='*',
     help='Override specific parameters directly from the command line (by node type or by node names).')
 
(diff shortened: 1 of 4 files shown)


ghost (approved):
.

cbentejac (dismissed):
It'd be nice to add some tests on the serialization/deserialization of graphs with a specific cache directory to ensure we're covering all cases. 

The test that's currently part of the PR only checks that the `setExplicitCacheDir` works as expected. 

Otherwise, it all looks good to me!

Alxiice:
I added another system following @fcastan recommendations :
We can have a list as the explicit cacheDir
image
We browse the list fist-to-last and we check for existence :
image
If none of the elements in the list is found then it fallsback to the default location (`./MeshroomCache`)

fabiencastan (approved):

cbentejac (approved):

Issue #3016, opened 2026-02-25
[bug] If images or project are saved in a folder that contains spaces, the processing will eventually fail.


tamius-han:
**Describe the bug**
If images or project are saved in a folder that contains spaces, the processing will eventually fail.

**To Reproduce**
1. Save images and/or project inside a folder with spaces in the file name
2. Open meshroom, create new 'Photogrammetry object' pipeline
3. Load images into the project
4. Start processing
5. Pipeline will eventually fail

**Expected behavior**
The pipeline finishes, eventually.

**Screenshots**
Here's the pipeline. Ignore the blue parts — that's me trying to re-run the pipeline after renaming the folder. (It didn't work because turns out you can't do that, as the project still tries to read images from the old path)

<img width="2744" height="716" alt="Image" src="https://github.com/user-attachments/assets/f33ee161-1c44-4114-8359-537c197527c9" />

**Log**

This appears in the console output a lot when trying to (re-)run the pipeline:

```
[...]

PR #3087, opened 2026-04-28
[tests] Add regression test for graph paths with spaces

brosequist:
## Summary
- add a regression test for isolated node compute command construction when the graph filepath contains spaces
- verify the generated command keeps the .mg path as one argument after shell-style parsing

## Context
The runtime fix for graph paths with spaces is already present on develop in commit da05aae0 and included in v2025.1.1+. This test protects against regressing the failure reported in #3016, where meshroom_compute received the project path split into extra arguments.

## Tests
- pytest tests/test_compute.py -k processChunkInEnvironment_quotesGraphFilepathWithSpaces
- pytest tests/test_compute.py

### DIFF
--- a/tests/test_compute.py
+++ b/tests/test_compute.py
@@ -8,6 +8,8 @@ TODO : We could directly test by launching the executable (`desc.node._MESHROOM_
 
 import os
 import re
+import shlex
+from types import SimpleNamespace
 from pathlib import Path
 import logging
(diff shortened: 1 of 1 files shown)


fabiencastan (approved):

cbentejac (approved):
Thanks for this contribution!

PR #3088, opened 2026-04-29
[ui/core] Add Validators paradigm

nicolas-lambert-tc:
## Description

[Alicevision related PR](https://github.com/alicevision/AliceVision/pull/1875)

Add a standard way to validate the attribute value.
It add a way to apply validators to an attribute description and to represent in the UI when these attributes are not valid.

Screenshot_2025-05-23_17-35-17

image

#### Validators
```python
class AttributeValidator(object):

    def __call__(self, node: "Node", attribute: "Attribute") -> tuple[bool, list[str]]: raise NotImplementedError()


class NotEmptyValidator(AttributeValidator):

    def __call__(self, node: "Node", attribute: "Attribute") -> tuple[bool, list[str]]:

        if attribute.value is None or attribute.value == "":
            return error("Empty value are not allowed")
        
        return success()
```

#### Validators usage (can be a lambda)

```python
desc.StringParam(
            name="input",
            label="Test input",
            description="",
            value="",
            validators=[
                NotEmptyValidator(),
                lambda node, attribute: success() if attribute == "test" else error("Attribute value should be 'test'")
                ]
        ),
```

## Features list
- [X] Add the notion of reusable validators
- [X] Validator embed the error messages
- [X] Validator can be a lambda
- [X] An icon with error message appear when an attribute value is invalid (at least one validator return error messages)
- [X] A node containing an invalid attribute is tagged with an orange icon and the selection rectangle is orange
- [X] Attribute value is checked every time the attribute value is set (in the Command) 
- [X] A Warning Dialog is displayed before trying to submit a graph where a node is in warning state

### DIFF
--- a/meshroom/core/attribute.py
+++ b/meshroom/core/attribute.py
@@ -12,6 +12,7 @@ import inspect
 from collections.abc import Iterable, Sequence
 from string import Template
 from meshroom.common import BaseObject, Property, Variant, Signal, ListModel, DictModel, Slot
+from meshroom.core.desc.validators import NotEmptyValidator
 from meshroom.core import desc, hashValue
 from meshroom.core.keyValues import KeyValues
 from meshroom.core.exception import InvalidEdgeError
@@ -84,6 +85,7 @@ class Attribute(BaseObject):
         self._enabled: bool = True
         self._depth: int = root.depth + 1 if root is not None else 0
         self._exposed: bool = root.exposed if root is not None else attributeDesc.exposed
+        self._description: str = attributeDesc.description
         self._invalidate = False if self._isOutput else attributeDesc.invalidate
         self._invalidationValue = ""  # invalidation value for output attributes
         self._value = None
@@ -234,6 +236,15 @@ class Attribute(BaseObject):
             # Internal attributes are set as inputs
             self.requestNodeUpdate()
         self.valueChanged.emit()
+
+    def _get_description(self):
+        return self._description
+
+    def _set_description(self, desc):
+        if self._description == desc:
+            return
+        self._description = desc
+        self.descriptionChanged.emit()
 
     def _getKeyValues(self):
         """
@@ -401,21 +412,6 @@ class Attribute(BaseObject):
         else:
             return self._getValue() == self.getDefaultValue()
 
-    def _isValid(self):
-        """
-        Check attribute description validValue:
-            - If it is a function, execute it and return the result
-            - Otherwise, simply return true
-        """
-        if callable(self._desc.validValue):
-            try:
-                return self._desc.validValue(self.node)
-            except Exception as exc:
-                if not self.node.isCompatibilityNode:
-                    logging.warning(f"Failed to evaluate 'isValid' (node lambda) for attribute '{self.fullName}': {exc}")
-                return True
-        return True
-
     def _is2dDisplayable(self) -> bool:
         """
         Return True if the current attribute is considered as a displayable 2D file.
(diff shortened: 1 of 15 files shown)


meshroom/core/attribute.py
```diff
@@ -235,6 +237,15 @@ def _setValue(self, value):
...
 
+    def _get_description(self):
+        return self._description
+
+    def _set_description(self, desc):
+        if self._description == desc:
+            return
+        self._description = desc
+        self.descriptionChanged.emit()
+
```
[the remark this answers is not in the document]
nicolas-lambert-tc: We don't want a useless attribute description property

meshroom/core/attribute.py
```diff
@@ -401,21 +412,6 @@ def _isDefault(self):
...
-        """
-        Check attribute description validValue:
-            - If it is a function, execute it and return the result
-            - Otherwise, simply return true
-        """
-        if callable(self._desc.validValue):
-            try:
-                return self._desc.validValue(self.node)
-            except Exception as exc:
-                if not self.node.isCompatibilityNode:
```
servantftransperfect, line 414: I can't see a corresponding check on the new code ?

meshroom/core/attribute.py
```diff
@@ -742,6 +775,9 @@ def validateIncomingConnection(self, connectingAttribute: Attribute) -> bool:
 
     expressionApplied = Signal()
 
+    errorMessageChanged = Signal()
+    errorMessages = Property(Variant, lambda self: self.getErrorMessages(), notify=errorMessageChanged)
```
servantftransperfect, line 765: I can't see that signal emitted ?

servantftransperfect, line 765: Shouldn't we have some caching here ? I guess getErrorMessages() is called many times, even if the attribute value did not change.

Alxiice (commented):
If I read well `validValue` is not used anymore so is it possible to add a `@deprecated.depreciateParam` on it (like for the "group" attribute) and maybe check where it is used ?

meshroom/core/attribute.py
nicolas-lambert-tc: It's needed because it's used internally by Qml. If not set, an error occure:
`QQmlExpression: Expression file:///s/apps/users/lambertni/packages/meshroom/lambertni/repo/meshroom/ui/qml/GraphEditor/AttributeItemDelegate.qml:28:5 depends on non-NOTIFYable properties:`

nicolas-lambert-tc: I tried to use caching, but I failed trying to keep the same robustness. Problems with connect/disconnect, undo/redo
It seems that signal have to be cared manually in a lot of places to be reactive on any user actions.
For now it's the more robust version i manage to have in a reasonable dev time.

meshroom/core/desc/validators.py
```diff
@@ -0,0 +1,67 @@
...
+
+            :returns: The validtion response: (True, []) if it's valid, (False, [errorMessage1, errorMessage2, ...]) if error exists
+
+        """
+        raise NotImplementedError()
+
+
+class NotEmptyValidator(AttributeValidator):
+    """ The attribute value should not be empty
+        This class is used to determine if an attribute label should be considered as mandatory/required
```
cbentejac:
```suggestion
        This class is used to determine if an attribute value should be considered as mandatory/required
```

meshroom/core/desc/validators.py
```diff
@@ -0,0 +1,67 @@
...
+
+class RangeValidator(AttributeValidator):
+    """ Check if the attribute value is in the given range
+    """
+
+    def __init__(self, min, max):
+        self._min = min
+        self._max = max
+
+    def __call__(self, node:"Node", attribute: "Attribute") -> tuple[bool, list[str]]:
```
cbentejac:
```suggestion
    def __call__(self, node: "Node", attribute: "Attribute") -> tuple[bool, list[str]]:
```

cbentejac:
```suggestion
    def __call__(self, node: "Node", attribute: "Attribute") -> tuple[bool, list[str]]:
```

meshroom/core/graph.py
```diff
@@ -1710,6 +1710,7 @@ def setVerbose(self, v):
     statusUpdated = Signal()
     canComputeLeavesChanged = Signal()
     canComputeLeaves = Property(bool, lambda self: self._canComputeLeaves, notify=canComputeLeavesChanged)
+    attributeValueChanged = Signal(Attribute)
```
cbentejac: Not used anywhere.

meshroom/core/attribute.py
```diff
@@ -84,6 +85,7 @@ def __init__(self, node, attributeDesc: desc.Attribute, isOutput: bool, root=Non
         self._enabled: bool = True
         self._depth: int = root.depth + 1 if root is not None else 0
         self._exposed: bool = root.exposed if root is not None else attributeDesc.exposed
+        self._description: str = attributeDesc.description
```
cbentejac: Unused

meshroom/core/attribute.py
cbentejac: Neither the methods nor the signal seem to be used

cbentejac (changes requested):
Functionally ready to be merged, but some docstrings need to be cleaned-up and some unused code removed.

cbentejac (approved):

PR #3093, opened 2026-05-07
[GraphEditor] Dynamically adjust `Backdrop`'s z-level based on its size

cbentejac:
## Description

When backdrops are overlapping, their superposition depends on which was created first. However, we would rather have the larger backdrop only below the smaller one, and have the z-levels updated dynamically when backdrops are resized.

This PR adjusts the z-level of the backdrop based on its size, which allows to dynamically re-order the backdrops. Nodes always remain on top.
backdrops

### DIFF
--- a/meshroom/ui/qml/GraphEditor/Backdrop.qml
+++ b/meshroom/ui/qml/GraphEditor/Backdrop.qml
@@ -77,8 +77,9 @@ Item {
     x: root.node ? root.node.x : undefined
     y: root.node ? root.node.y : undefined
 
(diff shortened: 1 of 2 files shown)


nicolas-lambert-tc (approved):
Clever fix, that should work on most cases, and for fixing actual bug.
For future improvement, it could be nice to handle the node/backdrop so the selected is always on top.
I also think that the backdrop children Z (node and backdrop) should be relative to the backdrop parent, and not to their surface area.

PR #2984, opened 2026-01-21
[desc] Execute pre & post process in the farm

Alxiice:
# PR description

We introduce a new feature that enables preprocessing and postprocessing tasks.
In this new version, these tasks are considered as new chunks and can be implemented on custom nodes.
On the scheduling/submitter part, preprocessing are executed before all chunks, and postprocessing are executed after them.
This way these can be used to prepare data for processing in chunks (preprocessing), and then concatenate data from all chunks (postprocessing).

### Adding preprocessing/postprocessing tasks
It's very simple and follows what we already use with `process`/`processChunk`
```py
class PluginSubmitterDPrePost(desc.Node):
    category = 'MyPlugins'
    inputs = INPUTS
    outputs = OUTPUTS

    def preprocess(self, node):
        LOGGER.info(f"> [{node.name}](preprocess) Start")
        # TODO
        LOGGER.info(f"> [{node.name}](preprocess) Done")

    def postprocess(self, node):
        LOGGER.info(f"> [{node.name}](postprocess) Start")
        # TODO
        LOGGER.info(f"> [{node.name}](postprocess) Done")
```

### Impact on the UI
image

If there no preprocess/postprocess the chunks will not be displayed on the UI.

# Implementation notes

- Create a `add_method_flag` decorator to be able to check whether a method is overloaded or not. `preprocess` and `postprocess` methods are flagged with it, and we know if we have a pre/postprocess using this system
- On the submitter part a new task graph system is introduced to rebuild the list of tasks. This way when we get to the submitter it's extremely simple to send tasks to the submitter. 

# Note to the reviewers

- Check the name of classes (Ordered...) on the submitter

If you use rez you can add this to the package commands :
```py
def commands():
    import os
    ...
    enabled_options = ('y', 'yes', 't', 'true', 'on', '1')
    if os.environ.get("ENABLE_LOCAL_SUBMITTERS", "0") in enabled_options:
        env.MESHROOM_SUBMITTERS_PATH.append('{base}/repo/meshroom/submitters')
    ...
```

# Merging
- First a PR to make sure nothing is changes on the submitter side : [mrSubmitters/4](https://github.com/meshroomHub/mrSubmitters/pull/4)
- Then this PR
- Finally a PR on the submitter side to use the new API : TODO

PR #3103, opened 2026-05-19
[submitter] Fix task chain errors

Alxiice:
## Description

Try to fix task chain errors on the submitter.
Some issues come from the previous PR : #2984
The tasks were incorrectly chained one after another. 

## Proposed fix

Rewriting the `__createNodeTasks` to handle better task chaining :
- [x] Core issue : bad chaining was caused by the `firstTask` and `lastTask` being the same placeholder, not 2 different ones.
- [x] Remove unnecessary placeholders for single chunks : using `firstTask`/`lastTask` to track first and last tasks on a node and avoid always creating placeholders (we only need them if we have more than 1 chunk)
- [x] Make sure to use preprocess/psotprocess instead of placeholder when we have them

Additionally :
  - [x] Adding an explicit logging handler
  - [x] OrderedTask : add shortName / uid to get a display that is more clear for future debugging
  - [x] taskManager : prune duplicate nodeToProcess + sort by depth
  - [x] dfs : important modification to avoid duplicates on several cases

### DIFF
--- a/meshroom/__init__.py
+++ b/meshroom/__init__.py
@@ -77,6 +77,13 @@ logStringToPython = {
     'trace': logging.TRACE,
 }
 logging.getLogger().setLevel(logStringToPython[os.environ.get('MESHROOM_VERBOSE', 'warning')])
+# Add default handler
+handler = logging.StreamHandler()
+handler.setFormatter(logging.Formatter(
+    '[%(asctime)s.%(msecs)03d][%(name)s][%(levelname)s] %(message)s', 
+    '%Y-%m-%d %H:%M:%S'
+))
+logging.getLogger().addHandler(handler)
 
 
 class MeshroomExitStatus(IntEnum):
--- a/meshroom/core/submitter.py
+++ b/meshroom/core/submitter.py
@@ -80,25 +80,48 @@ class OrderedTaskType(IntFlag):
 
 
 class OrderedTask:
+    _usedUids = set()
+
     def __init__(self, taskType, node = None, iteration : int = -1):
         self.taskType: OrderedTaskType = taskType
         self.node = node  # BaseNode
         self.iteration = iteration
         self.dependencies = []
-    
+        self.uid = self._generateUid()
+
+    @property
+    def nodeName(self):
+        return self.node.name if self.node else "NONE"
+
+    @classmethod
+    def _generateUid(cls) -> int:
+        nextUid = max(cls._usedUids) + 1 if len(cls._usedUids) > 0 else 0
+        cls._usedUids.add(nextUid)
+        return nextUid
+
     def addDependency(self, otherTask: OrderedTask):
         self.dependencies.append(otherTask)
-    
+
+    @property
+    def shortName(self):
+        sn = self.nodeName if self.node else "NONE"
+        if self.taskType == OrderedTaskType.CHUNK:
+            sn += f"_{self.iteration if self.iteration >=0 else 'allchunks'}"
+        else:
+            sn += f"_{self.taskType.name}"
+        return f"{self.uid:03d} {sn}"
+
     def __repr__(self):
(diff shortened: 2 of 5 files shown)


meshroom/__init__.py
```diff
@@ -77,6 +77,13 @@ def logToRoot(message, *args, **kwargs):
     'trace': logging.TRACE,
 }
 logging.getLogger().setLevel(logStringToPython[os.environ.get('MESHROOM_VERBOSE', 'warning')])
+# Add default handler
+handler = logging.StreamHandler()
+handler.setFormatter(logging.Formatter(
+    '[%(asctime)s.%(msecs)03d][%(name)s][%(levelname)s] %(message)s', 
+    '%Y-%m-%d %H:%M:%S'
+))
+logging.getLogger().addHandler(handler)
```
fabiencastan, line 86: logging changes should be in a separate commit

meshroom/core/submitter.py
```diff
@@ -15,7 +15,6 @@
 
 
 logger = logging.getLogger("Submitter")
-logger.setLevel(logging.INFO)
```
fabiencastan, line 18: separate commit

meshroom/core/submitter.py
```diff
@@ -259,54 +321,113 @@ def __createNodeTasks(self, orderedNode: OrderedNode, parentTask: OrderedTask):
...
+                            lastTask = postprocessTask
+                        elif preprocessTask:
+                            firstTask = preprocessTask
+                            lastTask = chunkTask
+                        elif postprocessTask:
+                            firstTask = chunkTask
+                            lastTask = postprocessTask
+                        else:
+                            firstTask = chunkTask
+                            lastTask = chunkTask
```
fabiencastan:
```suggestion
                        firstTask = preprocessTask or chunkTask
                        lastTask = postprocessTask or chunkTask
```

meshroom/core/submitter.py
```diff
@@ -81,22 +80,47 @@ class OrderedTaskType(IntFlag):
...
+        return self.node.name if self.node else "NONE"
+
+    @classmethod
+    def _generateUid(cls) -> int:
+        nextUid = list(cls._usedUids)[-1] if len(cls._usedUids) > 0 else 0
+        while nextUid in cls._usedUids:
+            nextUid += 1
+        uid = nextUid
+        cls._usedUids.add(uid)
+        return uid
```
fabiencastan: why not using the real UID of the node?

meshroom/core/submitter.py
```diff
@@ -211,27 +250,50 @@ def __orderNodes(self, nodes, edges):
...
+            childNode = nodeToOrderedNode[v._uid]
+            parentNode.dependencies.append(childNode)
 
         # Create a task 
         rootNode = OrderedNode(None, dependencies=nodeToOrderedNode.values())
         # Find each node depth (= what level the node is)
         depthByNode = {}
-        self.__updateDepth([rootNode], depthByNode, currentDepth=-1)
+        def __updateDepth(nodes: List[OrderedNode], currentDepth=0):
+            """ Compute the depth for each """
```
cbentejac, line 260:
```suggestion
            """ Recursively compute the depth for each node. """
```

meshroom/core/submitter.py
Alxiice: Because the goal of this uid is to easily debug the task graph that is sent to the submitter.
If we use the node uid first this identifies the node but not the task itself, but also this means we get a long string that is difficult to recognize.
If I use an icremental identifier, I will be able to easily identify tasks and their dependencies :
```
[001] : node {nodeName} ({nodeUid}) : {type} ({iteration if chunk task})
[002] : ...

Task graph :

  [001] - [002]

```
This is what is used below in the display function on the OrderedTasks class

cbentejac (approved):

PR #3107, opened 2026-05-21
Add support for nested `ListAttributes` and improve unit test coverage

cbentejac:
## Description

This PR adds the support for nested `ListAttributes`, while ensuring that all the accessing and setting features `GroupAttributes` remain functional. Up until now, only a single level of `ListAttribute` was correctly handled by Meshroom's core. With these changes, attributes with nested lists, which are written as `listAttr[x][y][z]`, are recognized with the dedicated regex, and are not confused with `GroupAttributes` anymore. 

The test coverage for `ListAttributes` is greatly improved, as well as the one for `GroupAttributes` that may contain lists.

Finally, a minor change is done to the `Node.qml` file to remove the direct injection of parameters into the signal handlers, which is deprecated and should be replaced with the use of JavaScript functions.

## Features list

- [x] Handle nested `ListAttributes` in nodes;
- [x] Add test coverage for nested `ListAttributes`, which now includes:
  - Nodes with nested list attributes.
  - Nodes with list-of-groups attributes.
  - Nodes with nested list-of-groups attributes, supporting multi-level indexed access.
- [x] Improve test coverage for `GroupAttributes` containing lists, which now includes:
  - Access and mutation of fields in lists of groups inside group attributes using path notation.
  - Access and mutation of simple lists inside group attributes.
  - Removal and shifting behavior in nested lists.
  - Deep path access for nested group attributes and lists of groups inside nested groups.
- [x] QML: Remove remaining parameter injections into signal handlers;
- [x] Apply linting on `test_groupAttributes.py` ("too many blank lines");
- [x] Rename `test_listAttribute.py` to `listAttributes.py`.

## Implementation remarks

* Updated the `attributeRE` regular expression in `BaseNode` to handle empty names, enabling support for advanced notations like `a[0][1]` (list of lists), and improved the attribute traversal logic to correctly handle empty name/index steps when accessing nested list and group attributes. [[1]](diffhunk://#diff-a9e65d402c922897d6a13963b45db2b469b20423b6f5f382c12c9d2b25fbe669L838-R839) [[2]](diffhunk://#diff-a9e65d402c922897d6a13963b45db2b469b20423b6f5f382c12c9d2b25fbe669L1061-R1076)


meshroom/core/node.py
```diff
@@ -1058,18 +1058,22 @@ def getNodeInfo(self):
...
-                # first step: get root attribute
+                if n == '' and idx == '':
+                    continue
+                # First step: get root attribute
                 if att is None:
                     att = self._attributes.get(n)
-                else:
+                elif n != '':
                     # get child Attribute in Group
                     assert isinstance(att, GroupAttribute)
```
fabiencastan: should be an if/throw instead of assert

meshroom/core/node.py
```diff
@@ -1058,18 +1058,22 @@ def getNodeInfo(self):
...
                     att = att.value.get(n)
+                # Else: n is empty, meaning this is an [index] step into a nested ListAttribute;
+                # fall through to the idx handling below
                 if idx != '':
                     # get child Attribute in List
                     assert isinstance(att, ListAttribute)
```
fabiencastan: if/throw

meshroom/core/node.py
```diff
@@ -1058,18 +1058,22 @@ def getNodeInfo(self):
...
             p = self.attributeRE.findall(name)
 
             for n, idx in p:
-                # first step: get root attribute
+                if n == '' and idx == '':
+                    continue
+                # First step: get root attribute
                 if att is None:
                     att = self._attributes.get(n)
-                else:
```
fabiencastan: It would be a bit simpler and more explicit like this:
```
if not n and not idx:
    raise ValueError(f"Empty access to attribute {attr.fullName} of type {type(att) if att else None}.")
elif n:
    if att is None:
        # get root attribute
        att = self._attributes.get(n)
    else:
        # It should be a group to access by name
        if not isinstance(att, GroupAttribute):
            raise ValueError(f"Cannot access '{n}': {attr.fullName} is not a GroupAttribute.")
        # get child Attribute in Group
        att = att.value.get(n)
elif idx:
        # It should be a list to access by index
        if not isinstance(att, ListAttribute):
            raise ValueError(f"Cannot access index '{idx}': {attr.fullName} is not a ListAttribute.")
        # get child Attribute in List
        att = att.value.at(int(idx))
```

cbentejac: Done, with a slight change:
```python
if not n and not idx:
    continue
```
Empty matches are expected considering the regex we're using and how `findall` works, so we should not throw. I added comments to explain it in the code.

meshroom/core/node.py

meshroom/core/node.py

fabiencastan (approved):

PR #3115, opened 2026-05-28
Add support for `onAttributeNameChanged` callbacks for nested `Attributes`

cbentejac:
## Description

This PR adds the support of the `on{AttributeName}Changed` callbacks for attributes that are nested in `GroupAttributes`. Prior to it, callbacks could not be used for nested attributes; either the callback was defined for the highest parent `GroupAttribute` or `ListAttribute` (and was thus triggered for every single change occurring in that `Group` or `List`), or no change could be tracked for that attribute.

Elements within `ListAttributes` still cannot use these callbacks; the callback can be setup for the list itself, meaning it will be triggered for any insertion/deletion in that list, but not for element value changes.

To determine whether an attribute with a parent can be bound to a callback, we introduce a new `isInsideList` property, that is evaluated once upon the attribute's creation, and determines whether there is a `ListAttribute` in the hierarchy of the current attribute. 

Tests are added for both the binding of callbacks and the `isInsideList` property.

## Features list

- [x] Catch `ValueError` exception in `_getRootName` when it is called for a list element while the element is being added to the list.
- [x] Add a new `isInsideList` property to determine whether an attribute is, directly or not, part of a `ListAttribute`.
- [x] Add a unit test to ensure `isInsideList` always evaluates correctly.
- [x] Support the binding of `on{AttributeName}Changed` callbacks for attributes that are nested within `GroupAttributes`.
- [x] Add unit tests to ensure the `on{AttributeName}Changed` callbacks are bound correctly depending on their attribute's status.

### DIFF
--- a/meshroom/core/node.py
+++ b/meshroom/core/node.py
@@ -1555,11 +1555,13 @@ class BaseNode(BaseObject):
     def _getAttributeChangedCallback(self, attr: Attribute) -> Optional[Callable]:
         """ Get the node descriptor-defined value changed callback associated to `attr` if any. """
 
-        # Callbacks cannot be defined on nested attributes.
-        if attr.root is not None:
+        # Callbacks cannot be defined on ListAttributes, but may be defined on nested attributes in GroupAttributes
+        if attr.root is not None and attr.isInsideList:
             return None
 
-        attrCapitalizedName = attr.name[:1].upper() + attr.name[1:]
+        attrCapitalizedName = ""
+        for sub in attr.rootName.split("."):
+            attrCapitalizedName += sub[:1].upper() + sub[1:]
         callbackName = f"on{attrCapitalizedName}Changed"
 
         callback = getattr(self.nodeDesc, callbackName, None)
--- a/meshroom/core/attribute.py
+++ b/meshroom/core/attribute.py
@@ -92,6 +92,14 @@ class Attribute(BaseObject):
         self._linkExpression: Optional[str] = None
         self._initValue()
 
+        self._isInsideList: bool = False
+        current = self._root() if self._root else None
+        while current is not None:
+            if isinstance(current, ListAttribute):
+                self._isInsideList = True
+                break
+            current = current._root() if current._root else None
+
     def _getFullName(self) -> str:
         """
         Get the attribute name following the path from the node to the attribute.
@@ -112,6 +120,10 @@ class Attribute(BaseObject):
         elif isinstance(self.root, GroupAttribute):
             return f'{self.root.rootName}.{self._desc.name}'
         return self._desc.name
+
+    @property
+    def isInsideList(self) -> bool:
+        return self._isInsideList
 
     def asLinkExpr(self) -> str:
         """
--- a/meshroom/core/attribute.py
+++ b/meshroom/core/attribute.py
@@ -105,7 +105,10 @@ class Attribute(BaseObject):
         Return: groupName.subGroupName.name
         """
         if isinstance(self.root, ListAttribute):
-            return f'{self.root.rootName}[{self.root.index(self)}]'
+            try:
+                return f'{self.root.rootName}[{self.root.index(self)}]'
+            except ValueError:
+                return f'{self.root.rootName}'
         elif isinstance(self.root, GroupAttribute):
             return f'{self.root.rootName}.{self._desc.name}'
         return self._desc.name
--- a/tests/test_attributes.py
+++ b/tests/test_attributes.py
@@ -1,10 +1,11 @@
+import logging
+import pytest
+
+from meshroom.core.attribute import Attribute
 from meshroom.core.graph import Graph
 from tests.utils import registerNodeDesc
 from tests.nodes.test.nodeValidators import NodeWithValidators
 
-import pytest
-
-import logging
 logger = logging.getLogger('test')
 
 valid3DExtensionFiles = [(f'test.{ext}', True) for ext in ('obj', 'stl', 'fbx', 'gltf', 'abc', 'ply', 'usda', 'usdc')]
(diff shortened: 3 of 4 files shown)


Alxiice:
So I have a picky remark concerning performances.
I created a simple test node :
```py
import inspect

class TestNodeA(desc.Node):
    category = 'TestPlugin'
    
    inputs = [
        desc.GroupAttribute(
            name="parentGrp",
            exposed=True,
            items=[
                desc.GroupAttribute(
                    name="nestedGrp",
                    exposed=True,
                    items=[
                        desc.IntParam(
                            name="x",
                            value=0,
                            exposed=True
                        ),
                        desc.IntParam(
                            name="y",
                            value=0,
                            exposed=True
                        ),
                    ],
                ),
            ],
        ),
    ]

    def display_func_infos(self, fname, node):
        print(f"[Func {fname}] (node {node.name})")
    
    def onParentGrpChanged(self, node):
        self.display_func_infos(inspect.stack()[0][3], node)
    
    def onParentGrpNestedGrpChanged(self, node):
        self.display_func_infos(inspect.stack()[0][3], node)
    
    def onParentGrpNestedGrpXChanged(self, node):
        self.display_func_infos(inspect.stack()[0][3], node)
    
    def onParentGrpNestedGrpYChanged(self, node):
        self.display_func_infos(inspect.stack()[0][3], node)
    
    def process(self, node):
        self.display_func_infos(inspect.stack()[0][3], node)
```
Here is the update execution order
image
Before (on develop) :
```bash
# TestA_1.parentGrp.nestedGrp.x -> cannot call callback because of nested attr
# TestB_1.parentGrp.nestedGrp.x -> cannot call callback because of nested attr
# TestB_1.parentGrp.nestedGrp   -> cannot call callback because of nested attr
[Func onParentGrpChanged] (node TestB_1)
# TestA_1.parentGrp.nestedGrp   -> cannot call callback because of nested attr
# TestB_1.parentGrp.nestedGrp   -> cannot call callback because of nested attr
[Func onParentGrpChanged] (node TestB_1)
[Func onParentGrpChanged] (node TestA_1)
[Func onParentGrpChanged] (node TestB_1)
```
After (on this branch) :
```bash
[Func onParentGrpNestedGrpXChanged] (node TestA_1)
[Func onParentGrpNestedGrpXChanged] (node TestB_1)
[Func onParentGrpNestedGrpChanged] (node TestB_1)
[Func onParentGrpChanged] (node TestB_1)
[Func onParentGrpNestedGrpChanged] (node TestA_1)
[Func onParentGrpNestedGrpChanged] (node TestB_1)
[Func onParentGrpChanged] (node TestB_1)
[Func onParentGrpChanged] (node TestA_1)
[Func onParentGrpChanged] (node TestB_1
```

It bugs me a little that we don't have a way to batch the updates on the parent node and then execute the updates on the children.

I extracted a stack trace to explain how the signals are called:
```
Change TestA.parentGrp.nestedGrp.x
    SetAttributeCommand.redoImpl
        Attribute._setValue
            Attribute._onValueChanged
                BaseNode._onAttributeChanged
                |   get the callback (BaseNode._getAttributeChangedCallback)
                |   callback is onParentGrpNestedGrpXChanged
                |   call callback
                |   for each edge on the graph:
                |       emit valueChanged
                |           [TestB.parentGrp.nestedGrp.x]
                |           |   valueChanged signal received on attribute TestB.parentGrp.nestedGrp.x
                |           |   Attribute._onValueChanged
                |           |       BaseNode._onAttributeChanged
                |           |           get the callback (BaseNode._getAttributeChangedCallback)
                |           |           callback is onParentGrpNestedGrpXChanged
                |           |           call callback
                |           |   Then because the TestB.parentGrp.nestedGrp _value is driven by the
                |           |   content inside the GroupAttribute, the signal valueChanged is called
                |           |   on the attribute TestB.parentGrp.nestedGrp
                |           |   Execute BaseNode._onAttributeChanged TestB.parentGrp.nestedGrp
                |           `-  Then execute BaseNode._onAttributeChanged TestB.parentGrp
                The attribute TestA.parentGrp.nestedGrp detects its content have changed 
                so it will trigger an update
                [On TestA.parentGrp.nestedGrp]
                |   BaseNode._onAttributeChanged
                |       get and call the callback -> onParentGrpNestedGrpChanged
                |       emit valueChanged on linked attributes
                |       -> onValueChanged on TestB.parentGrp.nestedGrp
                |          Then : onValuChanged on TestB.parentGrp
                [On TestA.parentGrp]
                |   BaseNode._onAttributeChanged
                |       get and call the callback -> onParentGrpChanged
                |       emit valueChanged on linked attributes
                `-       -> onValueChanged on TestB.parentGrp
```

Technically it's a "new" issue because we didn't see it before, but I would uderstand if you want to tackle this in another PR because this require probably a huge update on the signals... So maybe at least we could open an issue for this ?

meshroom/core/node.py
Alxiice, line 1607: It's just an idea but in this implementation if an element of the list change we don't even call the list update callback.
It's a draft but here's a snippet that make it _partially_ work :
```py
    def _onAttributeChanged(self, attr: Attribute):
        """
        When an attribute value has changed, a specific function can be defined in the descriptor
        and be called.

        Args:
            attr: The Attribute that has changed.
        """

        if self.isCompatibilityNode:
            # Compatibility nodes are not meant to be updated.
            return

        if attr.isOutput and not self.isInputNode:
            # Ignore changes on output attributes for non-input nodes
            # as they are updated during the node's computation.
            # And we do not want notifications during the graph processing.
            return

        if not attr.keyable and attr.value is None:
            # Discard dynamic values depending on the graph processing.
            return

        if self.graph and self.graph.isLoading:
            # Do not trigger attribute callbacks during the graph loading.
            return

        callback = self._getAttributeChangedCallback(attr)
        
        if not callback and attr.root is not None and attr.isInsideList:
            def getLastCallable(attr):
                if not attr.root or not attr.isInsideList:
                    return attr
                return getLastCallable(attr._root())
            attr = getLastCallable(attr)  # Important to replace attr for the call on edges
            callback = self._getAttributeChangedCallback(attr)

        if callback:
            callback(self)

        self.hasInvalidAttributeChanged.emit()

        if self.graph:
            # If we are in a graph, propagate the notification to the connected output attributes
            for edge in self.graph.outEdges(attr):
                edge.dst.valueChanged.emit()
```

Once again my test :
```py
class TestNodeB(desc.Node):
    category = 'TestPlugin'
    
    inputs = [
        desc.GroupAttribute(
            name="parentGrp",
            exposed=True,
            items=[
                desc.ListAttribute(
                    name="listAttr",
                    exposed=True,
                    elementDesc=desc.IntParam(
                        name="listItem",
                        value=0,
                    )
                )
            ]
        )
    ]

    def display_func_infos(self, fname, node):
        print(f"[Func {fname}] (node {node.name})")
    
    def onParentGrpChanged(self, node):
        self.display_func_infos(inspect.stack()[0][3], node)

    def onParentGrpListAttrChanged(self, node):
        self.display_func_infos(inspect.stack()[0][3], node)
    
    def onParentGrpListAttrListItemChanged(self, node):
        self.display_func_infos(inspect.stack()[0][3], node)
    
    def process(self, node):
        self.display_func_inf
```

Image

And the output :
```bash
[Func onParentGrpListAttrChanged] (node TestC_1)  # Update the closest attr that we can update
[Func onParentGrpListAttrChanged] (node TestD_1)  # Call update on child param
[Func onParentGrpChanged] (node TestD_1)          # Update parent on child param
# MISSING : Call onParentGrpChanged on node TestC_1
```

PR #3123, opened 2026-06-05
Add QML Actions callback system

Alxiice:
## Description

Propose a new system to add callbacks on specific actions.
For now we would add an optional callback to propose to save as template or save as project before saving.
This way this would avoid people saving templates as projects and therefore avoid compatibility node issues.

### DIFF
--- a/meshroom/core/callbacks.py
+++ b/meshroom/core/callbacks.py
@@ -1,0 +1,125 @@
+"""
+Callback registry system for Meshroom.
+
+This module provides a mechanism to register, query, and trigger callbacks
+that can execute before certain actions (e.g., save, saveAs).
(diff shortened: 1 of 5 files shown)


Alxiice:
Seems to work pretty well already.
I would like to clean up the code.
Maybe it's too over-engineered for what we actually want.

> [!WARNING]
> If we don't need it for other things than the save, then maybe do something more specific ?
> For example I started a while ago a [new custom file browser](https://github.com/alicevision/Meshroom/tree/dev/new_file_explorer) we could use this instead ? 

https://github.com/user-attachments/assets/51ef4971-443e-4277-9dfc-2227b5c70a68

PR #3126, opened 2026-06-08
[core] node: Prevent using default output value if it is already defined

cbentejac:
## Description

This pull request enhances the handling and testing of output attribute value templates in node descriptions. The main focus is on ensuring that when a node's output value template changes, existing graphs preserve the saved values, while new nodes adopt the updated template. The changes include improvements to attribute value formatting logic and the addition of comprehensive tests to verify compatibility behavior.

**Testing and compatibility improvements:**

* Added `OutputTemplateNodeV1` and `OutputTemplateNodeV2` test node classes in `tests/test_compatibility.py` to simulate a node whose output value template changes from `file.abc` to `file.cba`.
* Introduced the `test_outputValueAfterDefaultOutputChange` test, which verifies that saved graphs retain old output values after a template change, while new nodes use the updated template. This ensures backward compatibility and correct application of new defaults.

**Core logic enhancement:**

* Updated `_buildAttributeExpVars` in `meshroom/core/node.py` to improve output attribute value formatting. The logic now always recomputes the invalidation value from the template, ensuring that the attribute's state reflects the current UID and the latest template, regardless of whether the value was previously set.

### DIFF
--- a/meshroom/core/node.py
+++ b/meshroom/core/node.py
@@ -1272,8 +1272,16 @@ class BaseNode(BaseObject):
                         logging.warning(f'Invalid lambda evaluation for "{self.name}.{attr.name}"')
                     if defaultValue is not None:
                         try:
-                            attr.value = defaultValue.format(**self._expVars)
+                            if attr.value and (attr.value != defaultValue and
+                                               attr.value != defaultValue.format(**self._expVars)):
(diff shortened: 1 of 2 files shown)


cbentejac:
Replaced by https://github.com/alicevision/Meshroom/pull/3132.

PR #3128, opened 2026-06-09
Add API module to provide a simple interface to the Meshroom scene

Alxiice:
## Description

Add a new `api` module to offer an easier interface to scene and nodes.

## Features

- core:
  - set the log level
  - initialize (load plugins, nodes, etc)
  - list registered plugins
  - register a plugin
  - unregister a plugin
  - list registered NodePlugins
  - register a NodePlugin
  - unregister a NodePlugin
- scene:
  - load a graph
  - get nodes inside a graph, possibility to filter on node type
  - get backdrop nodes
  - get node by instance name
  - list nodes inside a backdrop node
  - get nodes attributes

### DIFF
--- a/meshroom/api/__init__.py
+++ b/meshroom/api/__init__.py
@@ -1,0 +1,22 @@
+# -*- coding: utf-8 -*-
+
+
+from meshroom.api.core import (
+    setLoglevel,
+    initialize,
+    listPlugins,
+    unregisterPlugin,
+    registerPlugin,
+    listNodes,
+    unregisterNode,
+    registerNode,
+)
+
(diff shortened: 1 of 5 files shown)


fabiencastan (changes requested):
Meshroom currently has no notion of public versus private APIs. Adding such a concept would require a larger architectural change and project-wide agreement, which is beyond the scope of this PR.

In its current state, the proposed changes introduce additional complexity and ambiguity without establishing the underlying framework needed to support them. Consequently, this PR will not be merged.

The new features to manipulate scenes should be moved in an existing module.

PR #3137, opened 2026-06-15
Add user-defined Menus

Alxiice:
## Description

New feature to add user-defined menus.
image

### Features

- Create Menu
- Add different items inside :
  - A sub-menu
  - A button
  - A checkbox
  - A radio button (choose option in a list of checkbox)
- Attach callbacks on clickable items
- Callbacks are either callable object or a function with specific params 
- Callbacks have access to the `MeshroomApp` instance so that we can access the node graph
- Callbacks have access to the menu object : for radio buttons, it enables having access to menu items, ...
- We also pass kwargs which change depending on the context (checked status for checkbox, selected item for radio button, etc)
- Each menu entry can have these properties :
  - `name` : name
  - `label` : display label (`name` by default)
  - `icon` : for menu entry and buttons
  - `tooltip` : display detailed message about the entry
  - `shortcut` : shortcut to call the entry
  - `index` : insert index (by default last one)
- Meshroom looks for a `menu.py` script in the `meshroom` folder of plugins to load menus.

### How the API looks

```bash
import logging
from meshroom.api import register_menu, Menu, MenuItem, MenuCallback

# === Callbacks ===

def say_hello(menu, app, **kwargs):
    logging.info("[MenuTest] Hello from the 'Say Hello' button!")
    app.showMessage("Hello from the Menu Test plugin!", duration=3000)


class ToggleVerboseLogging(MenuCallback):
    def __call__(self, menu, app, **kwargs):
        enabled = kwargs.get("enabled", False)
        level = "debug" if enabled else "info"
        logging.getLogger().setLevel(level.upper())
        logging.info(f"[MenuTest] Verbose logging {'enabled' if enabled else 'disabled'}.")


# === Create the Menu ===

user_menu = Menu("Settings")

user_menu.addButton(
    "sayHello",
    label="Say Hello",
    callback=say_hello,
    tooltip="Show a status bar message",
    icon="testMenu/meshroom/agitant-la-main.png", 
    shortcut="Ctrl+Alt+H",
)

user_menu.addSeparator()

user_menu.addCheckbox(
    "verboseLogging",
    label="Verbose Logging",
    callback=ToggleVerboseLogging(),
    tooltip="Toggle debug-level logging",
    checked=False,
)


# === Register the Menu ===

register_menu(user_menu)
```

Another example with a radio button :
```python
from meshroom.api import register_menu, Menu, MenuItem


def radioButtonCallback(menu, app, **kwargs):
    selected: MenuItem = kwargs["selected"]
    print(f"[MenuTest] Radiobutton selected: {selected.label}")
    app.showMessage(f"Radiobutton selected: {selected.label}", duration=2000)


menu_test = Menu("Test")

menu_test.addRadioButton(
    "testRadioButton",
    items=[
        MenuItem("a", label="A"),
        MenuItem("b", label="B"),
        MenuItem("c", label="C"),
    ],
    callback=radioButtonCallback,
)

register_menu(menu_test)
```

### Details

Here's how the code is organized. I ordered every file by logical order so you can review from the first to last file here :
- `core/`
  - `core/menu.py` : Define the Menu, MenuItem, MeshroomMenuManager, etc. All the objects used in the rest of the code, specific to a menu. Also build the list model for the QML part.
  - `core.__init__.py` : Add a `loadPluginMenu` function to find menus on plugins when launching Meshroom.
- `api/`
  - `api/menu.py` gives a top-level access to a register function. This is what should be used in plugins
  - `api/__init__.py` creates redirections to every useful class/function to create menus in plugins. That way every import should be :
  ```python
  from meshroom.api import Menu, register_menu, ...
  ```
- `ui/`
  - `ui/app.py` : set the menu manager as a context property to access it in QML
  - `ui/qml/Controls/userMenu.qml` : define the menu and components for every instanciable widget
    - `ui/qml/Application.qml` : Add Instantiator using the menu model.

### DIFF
--- a/meshroom/common/__init__.py
+++ b/meshroom/common/__init__.py
@@ -13,6 +13,7 @@ class Backend(Enum):
     PYSIDE = 2
 
 
+CurrentBackend = None
 DictModel = None
 ListModel = None
 Slot = None
@@ -25,7 +26,8 @@ JSValue = None
 
 
 def init(backend):
-    global DictModel, ListModel, Slot, Signal, Property, BaseObject, Variant, VariantList, JSValue
+    global CurrentBackend, DictModel, ListModel, Slot, Signal, Property, BaseObject, Variant, VariantList, JSValue
+    CurrentBackend = backend
     if backend == Backend.PYSIDE:
         # PySide types
         from .qt import DictModel, ListModel, Slot, Signal, Property, BaseObject, Variant, VariantList, JSValue
--- a/meshroom/core/__init__.py
+++ b/meshroom/core/__init__.py
@@ -343,6 +343,32 @@ def nodeVersion(nodeDesc: desc.Node, default=None):
     return moduleVersion(nodeDesc.__module__, default)
 
 
+def loadPluginMenu(pluginMenuFile):
+    """
+    Registers a menu from a python script
+
+    Args:
+        folder: the folder where we search for the menu.
+    """
+    # Generate a custom module name to avoid collisions
+    moduleName = f"mrMenus.{uuid.uuid4().hex}"
+    try:
+        spec = importlib.util.spec_from_file_location(moduleName, pluginMenuFile)
+        module = importlib.util.module_from_spec(spec)
+        sys.modules[moduleName] = module
+        spec.loader.exec_module(module)
+        logging.debug(f"Loaded menu registration from '{pluginMenuFile}'.")
+    except Exception as exc:
+        tb = traceback.extract_tb(exc.__traceback__)
+        last_call = tb[-1]
+        logging.warning(
+            f"  * Failed to load menu file '{pluginMenuFile}' ({type(exc).__name__}): {str(exc)}\n"
+            f"{last_call.filename}:{last_call.lineno} {last_call.name}\n"
+            f"{last_call.line}"
+            f"\n{traceback.format_exc()}\n\n"
+        )
+
+
 def loadNodes(folder, packageName, pluginUid) -> list[NodePlugin]:
     if not os.path.isdir(folder):
         logging.error(f"Node folder '{folder}' does not exist.")
@@ -376,6 +402,11 @@ def loadPluginFolder(folder, userPlugin: bool = False) -> list[Plugin]:
     if not mrFolder.exists():
         logging.info(f"Plugin folder '{folder}' does not contain a 'meshroom' folder.")
         return []
+
+    # Register any menu(s) declared by this plugin
+    pluginMenuFile = mrFolder / "menu.py"
+    if pluginMenuFile.exists():
+        loadPluginMenu(pluginMenuFile)
 
     plugins = loadAllNodes(folder=mrFolder)
     if plugins:
--- a/meshroom/core/__init__.py
+++ b/meshroom/core/__init__.py
@@ -1,6 +1,7 @@
 from contextlib import contextmanager
 import hashlib
 import importlib
+import importlib.util
 import inspect
 import logging
 import os
@@ -363,8 +364,11 @@ def loadPluginMenu(pluginMenuFile):
         last_call = tb[-1]
         logging.warning(
             f"  * Failed to load menu file '{pluginMenuFile}' ({type(exc).__name__}): {str(exc)}\n"
+            # filename:lineNumber functionName
             f"{last_call.filename}:{last_call.lineno} {last_call.name}\n"
+            # line of code with the error
             f"{last_call.line}"
+            # Full traceback
             f"\n{traceback.format_exc()}\n\n"
         )
 
--- a/meshroom/core/__init__.py
+++ b/meshroom/core/__init__.py
@@ -19,6 +19,7 @@ try:
 except Exception:
     pass
 
+from meshroom.common import CurrentBackend, Backend
 from meshroom.core.plugins import NodePlugin, NodePluginManager, Plugin, processEnvFactory, formatNodeDescriptionErrorMessage
 from meshroom.core.submitter import BaseSubmitter
 from meshroom.env import EnvVar, meshroomFolder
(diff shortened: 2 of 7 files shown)


meshroom/api/qml.py
```diff
@@ -0,0 +1,7 @@
+#!/usr/bin/env python
+
+from meshroom.ui.extensions import QmlExtensions
+
+
+def registerQmlSource(folder: str, name: str, major: int = 1, minor: int = 0):
+    QmlExtensions.registerQmlModule(folder, name, major, minor)
```
nicolas-lambert-tc: I guess is to load custom qml file, but it doesn't seem to be used anywhere?
If it's usable an example would be helpful

meshroom/core/menu.py
```diff
@@ -0,0 +1,302 @@
...
+    def __call__(self, menu, app, **kwargs):
+        pass
+
+
+class MenuObjectType(Enum):
+    MENU = "menu"
+    SEPARATOR = "separator"
+    BUTTON = "button"
+    CHECKBOX = "checkbox"
+    RADIOBUTTON = "radioButton"
```
nicolas-lambert-tc: Why not creating classes instead, that inherits from MenuIttem?
It could help to add special behaviors (example for the radioButton to give the actual selected, this kind of things)
It could also be responsible of it's own display

fabiencastan (changes requested):
Do not create an API module.

raphaelKoskas (changes requested):
First of all, I strongly suggest you keep the initial comment in your PR up-to-date to facilitate the diving into it :
    - meshroom.api does not exist anymore, consequently to Fabien's remark,
    - register_menu was renamed,
    - sub-menus and radio buttons do not support shortcuts, your comment is misleading,
    - sub-menus and radio buttons do not support icons, (but do not raise errors when having an icon assigned, it is just not displayed).
I would also suggest adding safeguards to prevent users from going wild when defining their menus. For instance, I would advise to not support having items with the same name or label inside those user-defined menus.

I also noticed a few issues from a functionality point of view :
    - One cannot define multiple menus in one menu.py. It may be a desirable feature, especially users would like to use this user-defined feature to define meshroom shortcuts for themselves and want them neatly organised. In the current state of the code, they would need to declare one plugin per family of shortcuts.
    - Currently, sub-menus are placed one after another with the same y coordinate, which implies in the case of large nested sub-menus structure, submenus will completely overlap. I would advise to add an offset on the y coordinate to prevent this effect and to facilitate browsing within said large nested sub-menus structure.
    - Image
    - I got myself a weird recursing component creation error  I don't understand with a display error `QQmlComponent: Component creation is recursing - aborting
QQmlComponent: Component creation is recursing - aborting
QQmlComponent: Component creation is recursing - aborting`
    Image
. Here is the code snippet that triggered it :
`
    mySubMenua = mySubMenu9.addSubmenu("Adding here a sub-menu",
                                    icon = "/usr/share/icons/Adwaita/32x32/legacy/accessories-dictionary.png",
                                    tooltip = "Just some test",
                                    index = -1)
    mySubMenub = mySubMenua.addSubmenu("Adding here a sub-menu",
                                    icon = "/usr/share/icons/Adwaita/32x32/legacy/accessories-dictionary.png",
                                    tooltip = "Just some test",
                                    index = 0)
    mySubMenuc = mySubMenua.addSubmenu("Adding here a sub-menu",
                                    icon = "/usr/share/icons/Adwaita/32x32/legacy/accessories-dictionary.png",
                                    tooltip = "Just some test",
                                    index = 1)
    mySubMenud = mySubMenuc.addSubmenu("Adding here a sub-menu",
                                    icon = "/usr/share/icons/Adwaita/32x32/legacy/accessories-dictionary.png",
                                    tooltip = "Just some test",
                                    index = -1)
    mySubMenue = mySubMenua.addSubmenu("Adding here a sub-menu",
                                    icon = "/usr/share/icons/Adwaita/32x32/legacy/accessories-dictionary.png",
                                    tooltip = "Just some test",
                                    index = -1)
`
    - You should add safeguards to prevent items in a radio button to have the same names or labels. Right now it is supported, and the behaviour seems to be that on startup, the last occurrence of the first object is selected, when selecting a duplicated object, the display selects one at random, the callback seems to always call the first duplicated object.
    - Image
    - I also got a weird behaviour when defining a button : 
         - the button got executed on meshroom startup,
         - when opening the scene, I got a recursing component error `QQmlComponent: Component creation is recursing - aborting
QQmlComponent: Component creation is recursing - aborting
QQmlComponent: Component creation is recursing - aborting`,
         - and when pressing the button once the scene had opened i got an error : 
         `Traceback (most recent call last):
  File "/s/apps/users/koskasdediegor/packages/mikrosVfx/multiview/meshroom/raph_develop/repo/meshroom/ui/menu.py", line 358, in triggerButton
    self.trigger(uid)
  File "/s/apps/users/koskasdediegor/packages/mikrosVfx/multiview/meshroom/raph_develop/repo/meshroom/ui/menu.py", line 354, in trigger
    return obj.trigger(self.app, **kwargs)
           ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "/s/apps/users/koskasdediegor/packages/mikrosVfx/multiview/meshroom/raph_develop/repo/meshroom/ui/menu.py", line 81, in trigger
    return self.callback(self.parentMenu, app, **kwargs)
           ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
TypeError: 'CompletedProcess' object is not callable
file:///s/apps/users/koskasdediegor/packages/mikrosVfx/multiview/meshroom/raph_develop/repo/meshroom/ui/qml/Controls/UserMenu.qml:41: TypeError: 'CompletedProcess' object is not callable`
    - Some characters are not supported (at least full-width characters and emojis).

PR #3139, opened 2026-06-18
Plugins: Use unique names

Alxiice:
## Description

This feature addresses a current limitation about node plugins.
Before this, if we had 2 plugins :
```
/pluginA
    meshroom/
        nodes/
            __init__.py
            pluginANodes.py

/pluginB
    meshroom/
        nodes/
            __init__.py
            pluginBNodes.py
```
Then as the module name is the same for both plugins ("nodes"), then we would have one of the plugins that would not load.

## Details

The way I addressed this issue is :
- declare a unique ID on the plugins. When we create an instance of a plugin, it generates a new UID.
- use the unique name in the plugin manager dictionary
- updating the `add_to_path` function so that it will modify the loaded modules to use the plugin UID. Modules will be named `pluginUID_moduleName` instead of `moduleName`. This way another plugin will always have a different prefix.

## Features

- we can also have in 2 plugins a node script with the same name as in another plugin and it will still work
- nothing breaks : moving a node from a plugin to another is still automatically resolved, reloading plugins still work etc

## Limitations / drawbacks

- In the UI documentation tab we can see the UID prefix in the module name, it can be a bit confusing (but at the same time it enables identifying modules in a unique way)
- it only works for node plugins, not for submitter plugins, because they don't really declare a plugin. So if we have 2 submitters with the same module name it will have the same behavior as before. This is not a huge issue in this case because on [mrSubmitters](https://github.com/meshroomHub/mrSubmitters/tree/master/meshroom) for example we would use the submitter name as the module folder

### DIFF
--- a/meshroom/core/plugins.py
+++ b/meshroom/core/plugins.py
@@ -293,10 +293,14 @@ class Plugin(BaseObject):
         configFullEnv: the static merge of os.environ and configEnv, with os.environ taking precedence
         processEnv: the environment required for the nodes' processes to be correctly executed
     """
+    
+    _pluginLastUid = 0
 
     def __init__(self, name: str, path: str):
         super().__init__()
 
+        Plugin._pluginLastUid += 1
+        self._uid: str = f"{Plugin._pluginLastUid:04d}"
         self._name: str = name
         self._path: str = path
 
@@ -309,10 +313,18 @@ class Plugin(BaseObject):
         self.loadTemplates()
         self.loadConfig()
 
+    def __repr__(self):
+        return f"<Plugin {self._name} (uid={self._uid})>"
+
     @property
     def name(self):
         """ Return the name of the plugin. """
         return self._name
+    
+    @property
+    def uname(self):
+        """ Return the unique name of the plugin. """
+        return f"{self._uid}_{self._name}"
 
     @property
     def path(self):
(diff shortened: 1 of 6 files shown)


meshroom/core/plugins.py
```diff
@@ -634,18 +646,28 @@ def getPlugins(self) -> dict[str: Plugin]:
...
-            return self._plugins[name]
+        if uname:
+            # Find plugin with unique name
+            if name in self._plugins:
+                return self._plugins[name]
+        else:
+            for plugin in self._plugins.values():
+                if plugin.name == name:
+                    return plugin
         return None
```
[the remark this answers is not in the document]
Alxiice: this is true, however :
- the dict key is now the uname so this is logical to use the uname as the default behaviour for this method
- using `uname=False` can lead to collisions that were ignored previously because we could not register 2 plugins with the same module name.

tests/test_compute.py
```diff
@@ -245,7 +245,7 @@ def setup_class(cls):
         folder = os.path.join(os.path.dirname(__file__), "plugins", "meshroom")
         package = "pluginA"
         cls.plugin = Plugin(package, folder)
-        nodes = loadClassesNodes(folder, package)
+        nodes = loadClassesNodes(folder, package, pluginUid=cls.plugin._uid)
         for node in nodes:
```
nicolas-lambert-tc: It would be preferable to expose an uid getter property to the Plugin interface.
It could also let the opportunity to override the uid generation behavior for future specific Plugins

meshroom/core/plugins.py
```diff
@@ -293,10 +293,14 @@ class Plugin(BaseObject):
         configFullEnv: the static merge of os.environ and configEnv, with os.environ taking precedence
         processEnv: the environment required for the nodes' processes to be correctly executed
     """
+    
+    _pluginLastUid = 0
```
nicolas-lambert-tc: I would change `_pluginLastUid -> _instancesCount` because it's used to compute the uid, but it's not an uid.

nicolas-lambert-tc (commented):
Nice to have

nicolas-lambert-tc (approved):

tests/test_plugins.py
```diff
@@ -28,14 +28,24 @@ def teardown_class(cls):
             pluginManager.unregisterNode(node)
         pluginManager.removePlugin(cls.plugin)
         cls.plugin = None
-
-    def test_loadedPlugin(self):
+
```
cbentejac: Trailing whitespaces

```suggestion

```

tests/test_compute.py
cbentejac: +1

Alxiice: Ok for the getter.

About the uid generation, I'm not sure on what occasion we would want that. Also it would be tricky as the uid is used to identify the python module so it would need to be set when meshroom initialize the plugins and cannot be changed after.

If there really is a reason why we would set a specific uid for a plugin, then it would be better to use a set for the `_instancesCount` to make sure we keep track of the module names we cannot use again

cbentejac (approved):

PR #3141, opened 2026-06-18
Advanced submitter settings

Alxiice:
## Description

Add a system to setup advanced settings on submitter

Before this PR there was 2 issues :
- cpu/ram/gpu are the same for preprocess/process/postprocess
- it's quite difficult to add new parameters in a proper way

Also for the [mrSubmitters](https://github.com/meshroomHub/mrSubmitters) submitter it was really challenging to setup a proper job for Windows.

## Details

In this PR we are adding 2 key elements :
- `SubmitionSettings` & `StageSettings` classes to describe settings for a specific processing stage (`preprocess`, `process`/`processChunk` and `postprocess`)
- A new method on `desc.BaseNode` : `getSubmitionSettings`. By default it is built using the nodeDesc : `cpu`/`ram`/`gpu` are settings used for the `process` stage only, `pre/post process` are using default settings.

## Features
- by default use the same behaviour as currently, except for preprocess/postprocess stages that use default parameters which is more logical.
- by stage we can set :
  - cpu, ram, gpu
  - a setup/teardown commands that the submitter can use
  - a command wrapper
- the `StageSettings` is built using `SimpleNamespace` so we can technically set anything on a stage settings and use it later on the submitter.

## Example

```python
class TestWindowsNode(desc.Node):
    def getSubmitSettings(self, node) -> desc.SubmitionSettings:
        nodeSubmitSettings = desc.SubmitionSettings(node)
        nodeSubmitSettings.process.env_key = {
            "PROD_ROOT": self.get_win_path("P:\\my_prod"),
            "PROD_TMP": self.get_win_path("P:\\my_prod\\tmp"),
            "PYTHONPATH": convert_to_windows(os.getenv("PYTHONPATH")),
            "USER": "me",
        }
        nodeSubmitSettings.process.target_os = "windows"
        nodeSubmitSettings.process.wrapCommand = windowsCommandWrapper
        nodeSubmitSettings.process.setup_command = get_mount_volumes_command()
        return nodeSubmitSettings

        ...
```

### DIFF
--- a/meshroom/core/desc/node.py
+++ b/meshroom/core/desc/node.py
@@ -1,5 +1,3 @@
-# desc/node.py
-
 import enum
 from inspect import getfile, getattr_static
 from pathlib import Path
@@ -63,19 +61,12 @@ class MrNodeType(enum.Enum):
     BACKDROP = enum.auto()
 
 
-class MeshroomCommandWrapper:
-    def __call__(self, node, command: str) -> str:
-        return command
-
-
 class StageSettings(SimpleNamespace):
     def __init__(self, cpu=Level.NORMAL, ram=Level.NORMAL, gpu=Level.NONE, **kwargs):
         super().__init__(**kwargs)
         self.cpu = cpu.value if hasattr(cpu, "value") else cpu
         self.ram = ram.value if hasattr(cpu, "value") else cpu
         self.gpu = gpu.value if hasattr(cpu, "value") else cpu
-        # Wraps the stage command
-        self.wrapCommand = MeshroomCommandWrapper()
 
     def __contains__(self, item):
         return item in self.__dict__
--- a/meshroom/core/desc/node.py
+++ b/meshroom/core/desc/node.py
@@ -10,6 +10,7 @@ import sys
 import signal
 import subprocess
 from collections import OrderedDict
+from types import SimpleNamespace
 import psutil
 
 from meshroom import _MESHROOM_ROOT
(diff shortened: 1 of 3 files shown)


meshroom/core/desc/node.py
```diff
@@ -60,6 +62,60 @@ class MrNodeType(enum.Enum):
...
+
+
+class StageSettings(SimpleNamespace):
+    def __init__(self, cpu=Level.NORMAL, ram=Level.NORMAL, gpu=Level.NONE, **kwargs):
+        super().__init__(**kwargs)
+        self.cpu = cpu.value
+        self.ram = ram.value
+        self.gpu = gpu.value
+        # command executed right before the stage
+        if kwargs.get("setup_command"):
```
servantftransperfect: Those four lines are no-op, right ?

meshroom/core/submitter.py
```diff
@@ -82,31 +85,24 @@ class OrderedTaskType(IntFlag):
...
         self.taskType: OrderedTaskType = taskType
-        self.node = node  # BaseNode
-        self.iteration = iteration
-        self.dependencies = []
-        self.uid = self._generateUid()
+        self.node: BaseNode = node
+        self.iteration: int = iteration
+        self.dependencies : list["OrderedTask"] = []
+        # Generate UID
+        self.uid: int = max(self._usedUids) + 1 if len(self._usedUids) > 0 else 0
```
servantftransperfect: Can you consider using cls ?

PR #3142, opened 2026-06-19
Feature/table view

raphaelKoskas:
## Description

This PR adds a new QML to visualise ListAttributes of GroupAttributes as a table.

It contains the following:
- a collapse button
- an element counter
- add/remove buttons
- a windowed button
- a resizable table view

## Features list

- [ ] Added the tableViewAttributeComponent
- [ ] Added the listAttributeComponent/tableViewAttributeComponent switch, in the case of listAttribtes of groupsAttributes
- [ ] Added a collapse button for each table views
- [ ] Added an elements counter to display the number of groups inside the list
- [ ] Added add/remove buttons to add/remove groups from the list
- [ ] Mouse wheel is supported (mouse wheel : vertical scroll, ctrl + mouse wheel : horizontal scroll)
- [ ] Editing values is supported
- [ ] Windowed visualisation of the tableview
- [ ]  choiceComponent supported
- [ ] choiceMultiComponent supported
- [ ] sliderComponent supported
- [ ] checkboxComponent supported
- [ ] textFieldComponent supported
- [ ] textAreaComponent supported
- [ ] colorComponent supported
- [ ] pushButtonComponent supported
- [ ] colorHueComponent supported
- [ ] Always visible headers
- [ ] Filling height/width windowed visualisation

## Implementation remarks

- We use repeaters instead of a real tableview as tablemodel does not seem to support dynamic layout generation.
- GroupAttributes and ListAttributes are not supported to prevent complex nested structures


meshroom/ui/qml/GraphEditor/AttributeItemDelegate.qml
```diff
@@ -297,7 +297,10 @@ RowLayout {
                 case "BoolParam":
                     return checkboxComponent
                 case "ListAttribute":
-                    return listAttributeComponent
+                    if (attribute.baseType == "GroupAttribute")
+                        return tableViewAttributeComponent
```
Alxiice, line 301:
```suggestion
                    if (attribute.baseType == "GroupAttribute" and attribute.desc.semantic.includes('table-view'))
                        return tableViewAttributeComponent
```
I would suggest to let the possibility to have a list of group instead of a table view. And if you want to use this semantic by default, add it in the ListAttribute desc

meshroom/ui/qml/GraphEditor/AttributeItemDelegate.qml
```diff
@@ -830,6 +833,696 @@ RowLayout {
...
+                    palette.buttonText:      tableLayout.appPalette.buttonText
+                    palette.highlight:       tableLayout.appPalette.highlight
+                    palette.highlightedText: tableLayout.appPalette.highlightedText
+                    palette.mid:             tableLayout.appPalette.mid
+                    palette.dark:            tableLayout.appPalette.dark
+                    palette.light:           tableLayout.appPalette.light
+                    palette.midlight:        tableLayout.appPalette.midlight
+                    palette.shadow:          tableLayout.appPalette.shadow
+                    palette.toolTipBase:     tableLayout.appPalette.toolTipBase
+                    palette.toolTipText:     tableLayout.appPalette.toolTipText
```
Alxiice:
```suggestion
                    palette: tableLayout.appPalette
```

Alxiice (changes requested):
So I have a few general changes that I would prefer to see before diving again in this PR:
- First I don't think this is a good design to have a 700-lines component inserted here, I think you could use the `AttributeControls` folder to put your delegates which would help cleaning this PR a lot. I see at least a `TableViewAttribute.qml`, `TableViewColumnDelegate.qml` and `TableViewCellDelegate.qml` components that you can create.
- I think this would make it much easier [if you use this](https://doc.qt.io/qt-6/qml-qtquick-tableview.html) which is a native Table View QML component.
- Please avoid patterns like `var h = rowHeights` to make "smaller" variables, this makes the code less readable.
- Finally, I have several UX remarks:
  - The table view is not responsibe (no layout fillWidth)
  - I think you should add the "add row" action to the table view Window too.

raphaelKoskas:
And I would have preferred you read the implementation remarks I left before diving into your comment ;-) The native Table View component was rejected due to, to my knowledge, lack of dynamic layout support.
Keeping a 700-lines component inserted was a concerted decision with @cbentejac, as it would be no trivial change, and it is supposed to be a temporary measure.
Also, I fail to see why we would like to keep the possibility for a list of groups, as the choice to migrate it was made out of ergonomy concerns. Therefore, your suggestions seems a bit overkill to me.
I agree on all your other points and they'll be adressed.

meshroom/ui/qml/GraphEditor/AttributeControls/TableViewRowDelegate.qml
```diff
@@ -0,0 +1,52 @@
...
+    MouseArea {
+        id: colResizeHandle
+        width: parent.width
+        height: 6
+        anchors.bottom:  parent.bottom
+        cursorShape: Qt.SizeVerCursor
+        preventStealing: true
+        property real lastY: 0
+        onPressed:  function(mouse) {
+            colResizeHandle.grabMouse();
```
cbentejac:
```suggestion
            colResizeHandle.grabMouse()
```

Both are correct but uniformity is key and we tend, in Meshroom, to not use the semi-colon in these statements.

meshroom/ui/qml/GraphEditor/AttributeControls/TableView.qml
```diff
@@ -0,0 +1,420 @@
...
+import MaterialIcons 2.2
+import Utils 1.0 
+
+ColumnLayout {
+    id: root
+    spacing: 0
+    property bool editable: true
+    required property var attribute
+    property var columnNames: {
+        if (!attribute || !attribute.value || attribute.value.count === 0) return []
```
cbentejac:
```suggestion
        if (!attribute || !attribute.value || attribute.value.count === 0)
            return []
```


For readability purposes, we prefer having these if/for as two lines (braces are not necessary) rather than a single one.

meshroom/ui/qml/GraphEditor/AttributeControls/TableView.qml
```diff
@@ -0,0 +1,420 @@
...
+    }
+    property real totalTableHeight: {
+        if (!rowHeights || rowHeights.length === 0) return 0
+        var t = 0
+        for (var i = 0; i < rowHeights.length; i++) t += rowHeights[i]
+        t += Math.max(0, rowHeights.length - 1)
+        return t
+    }
+    property var  scaledColumnWidths: []
+    property real scaledTableWidth:   0
```
cbentejac:
```suggestion
    property real scaledTableWidth: 0
```
Generally speaking, we do not align variables; a single space is needed here.

meshroom/ui/qml/GraphEditor/AttributeControls/TableView.qml
```diff
@@ -0,0 +1,420 @@
...
+    property var  scaledRowHeights:  []
+    property real scaledTableHeight: 0
+    property real availableW: outerFrame.width > 0 ? outerFrame.width  - fixedStrip.width  - vBar.width : 600
```
cbentejac:
```suggestion
    property real availableW: outerFrame.width > 0 ? outerFrame.width - fixedStrip.width - vBar.width : 600
```

meshroom/ui/qml/GraphEditor/AttributeControls/TableView.qml
```diff
@@ -0,0 +1,420 @@
...
+                outerFrame.width   = Qt.binding(function() { return fullscreenWindow.width  })
+                outerFrame.height  = Qt.binding(function() { return fullscreenWindow.height })
+                outerFrame.isFullscreen = true
+                fullscreenWindow.show()
+            }
+        }
+    }
+    Window {
+        id: fullscreenWindow
+        color: "#2d2d2d"
```
cbentejac: Better to use the palette here rather than arbitrary colors. If you need variations, you can play with "Qt.lighter" or "Qt.darker". Otherwise the colors won't adapt depending on whether the dark or light palette is used.

meshroom/ui/qml/GraphEditor/AttributeControls/TableViewCellDelegate.qml
```diff
@@ -0,0 +1,287 @@
...
+        return item.activeFocus ||
+               (item.children && item.children.length > 0 &&
+                item.children[0] && item.children[0].activeFocus)
+    }
+    Rectangle {
+        anchors.centerIn: parent
+        width:   cellLoader.width  + 8
+        height:  cellLoader.height + 4
+        radius:  3
+        color:   "#444444"
```
cbentejac: Use the palette here as well.

meshroom/ui/qml/GraphEditor/AttributeControls/TableViewCellDelegate.qml
cbentejac, line 1: I had not realized that the components from `AttributeItemDelegate` were almost all duplicated here. It will be a bit of a hassle but in the end I think it'd be best to externalize them into their own components, and instantiate them both in `AttributeItemDelegate` and here. 

I might even go as far as saying that removing the components' definition from `AttributeItemDelegate.qml` could be a stand-alone PR on which this one would rely. To be discussed.

meshroom/ui/qml/GraphEditor/AttributeControls/TableView.qml
```diff
@@ -0,0 +1,420 @@
...
+                    : 0
+            })
+            outerFrame.visible = Qt.binding(function() {
+                return root.expanded
+            })
+        }
+    }
+    FontMetrics {
+        id: fontMetrics
+        font.bold: false
```
cbentejac: From an aesthetic standpoint, I actually think it'd be good to use bold font for the header.  I guess the definition of said header isn't even in this file, but since I'm thinking of it just now... :)

meshroom/ui/qml/GraphEditor/AttributeControls/TableView.qml
```diff
@@ -0,0 +1,420 @@
...
+                outerFrame.isFullscreen = true
+                fullscreenWindow.show()
+            }
+        }
+    }
+    Window {
+        id: fullscreenWindow
+        color: "#2d2d2d"
+        width:  root.totalTableWidth  + 30 + 16 + 20
+        height: root.totalTableHeight + 10 + 16 + 20
```
cbentejac: Lots of magic numbers here, it'd be good to either add a comment to explain them and/or why we are adding them to get the final width/height, or declare them as constants if they are used regularly across the component.

Either way, a comment could go a long way :)
(this is applicable to a lot of magic numbers across this PR)

N.B: I think I have seen some of these that are actually useful for padding purposes, but are not used _in_ a padding property. If they are indeed used for padding, they should be set with the padding property rather than the width/height.

cbentejac (changes requested):
Architecturally speaking, the split between the 3 QML files is relevant and works well! It is functional, and I already see where the component could be re-used (outside of the Attribute Editor and of node-custom interfaces).

If I were to sum up some of my in-file comments, I'd say:
- Avoid using absolute colors; use the palette instead, and variations using `Qt.lighter` and `Qt.darker` to ensure we stick to the current theme (dark/light palettes).
-  Syntax-wise, do not align variables with spaces when assigning them, do not use one-liners for `if`/`for` statements. Beware of magic numbers in width/height computations.
- I think we do need to take the definition of the components out of `AttributeItemDelegate.qml` (and out of `TableViewCellDelegate.qml`), but most likely in another PR (that would require to be merged prior to this one).

Now from a purely functional point of view, I noticed a few issues (that I haven't connected to the code, since I have not looked at it in-depth):
- The initial height display for lists with a single element is incorrect, especially when compared to lists with several elements:
Image.
It is also the case when opening it in full screen:
Image. 

- When extending a "full screen" window to the actual full screen, the display is weird if there are few elements:
Image
We would prefer having fixed-size rows, and if the window is higher than the number of elements there are to display, then it should be empty when there's no element to fill it. This for example creates issue when clicking out of an attribute, as there is absolutely nowhere to "empty" to click. This may lead to involuntary changes of attribute values that may not even be noticed until it is too late (it happened to me, with the intrinsics displayed on the screenshot here).

- Since the width of the columns is based on the length of the column title, it sometimes gives strange displays:
Image
It'd probably be best to have a minimum column width that is wider than what we have here, and that may also depend of the component that is displayed in it. 

- It is not visible in the previous screenshot, but the focus highlight is on the whole cell rather than on the element that is being edited. It is quite striking when using the full screen mode (with rows that are higher than expected because they're filling up the empty space). 

- I noticed some inconsistent behaviors with the scrollbars (especially the vertical one) depending on where we are hovering when attempting to scroll. For example it does not work at all when hovering (without clicking or having the focus anywhere on the table view) on an integer attribute, but it does when hovering on a string attribute, on the very same row. When hovering on the header, the scroll starts but is then "stopped" before we get to reach the bottom of the list.

- The horizontal is not always visible, as it is attached to the bottom of the table view, which is itself often hidden: 
Image
It is a problem because scrolling down the table view vertically never brings us to reach it if it is not visible in the first place, as only the outer scrollbar (the Attribute Editor's) allows it. There probably is a way to attach it to the bottom of the view rather than the bottom of the component itself. 

- Scrolling when hovering (without having selected) on an attribute may change its value. I believe this a behavior that we have on several attribute components but it should be disabled here unless the attribute is selected. Otherwise attributes' values are changed unwillingly and unknowingly. At least they are correctly recorded on the undo stack!
 
- When scrolling over int attributes in a large group (and apparently attempting to change their values), I got the following errors:
```python
[2026-07-30 19:16:20.815][root][ERROR] Error while trying command 'Set Attribute 'CameraInit_1.viewpoints[0].viewId'': 
  File "meshroom/core/desc/attribute.py", line 475, in validateValue
    return int(value)
           ^^^^^^^^^^
ValueError: cannot convert float NaN to integer

During handling of the above exception, another exception occurred:

  File "meshroom/ui/commands.py", line 66, in tryAndPush
    res = command.redoImpl()
          ^^^^^^^^^^^^^^^^^^
  File "meshroom/ui/commands.py", line 321, in redoImpl
    attribute.value = self.value
    ^^^^^^^^^^^^^^^
  File "meshroom/core/attribute.py", line 249, in _setValue
    convertedValue = self.validateValue(value)
                     ^^^^^^^^^^^^^^^^^^^^^^^^^
  File "meshroom/core/attribute.py", line 420, in validateValue
    return self._desc.validateValue(value)
           ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "meshroom/core/desc/attribute.py", line 477, in validateValue
    raise ValueError(f"IntParam only supports int value (param: {self.name}, value: "
ValueError: IntParam only supports int value (param: viewId, value: nan, type: <class 'float'>)
```

PR #3146, opened 2026-06-22
WIP interface categories

servantftransperfect:
# Interface-based node category system

Introduces a lightweight interface mechanism for node descriptors, replacing hardcoded node type lists with a declarative, extensible approach. The initial interface implemented is `FeatureProviderInterface`.

See https://github.com/alicevision/AliceVision/pull/2142

---

## Motivation

Previously, the UI determined which node to use as the active "feature provider" by checking against a hardcoded list of node type names (`"featureProvider": ["FeatureExtraction", "FeatureMatching", ...]`). This made it impossible for external plugins to participate without modifying core Meshroom. Similarly, the QML code for resolving features folders contained node-type-specific branching logic.

---

## Changes

**`meshroom/core/desc/interface.py`** *(new file)*
- Introduces `Interface`, `InterfaceMeta`, and `InterfacedClass` — a minimal metaclass-based system that automatically populates a `interfaces` list on any node descriptor class that inherits from an `Interface`.
- Defines `FeatureProviderInterface` as the first concrete interface, with a `getFeaturesFolders(node)` method.

**`meshroom/core/desc/node.py`**
- `BaseNode` now inherits from `InterfacedClass`, enabling all node descriptors to declare interfaces.

**`meshroom/ui/scene.py`**
- `initActiveNodes` now iterates all registered node plugins and registers an `ActiveNode` entry per interface declared on the descriptor.
- `setActiveNode` updates the interface-keyed active node entries when a node becomes active.
- New `callInterfaceMethod(interfaceName, methodName)` slot allows QML to generically call any method defined by an interface on the currently active node for that interface, with error handling.

**`meshroom/ui/qml/Viewer/Viewer2D.qml`**
- Active node lookups for features now use `"FeatureProviderInterface"` as the key instead of `"featureProvider"`.
- Node-type-specific branching for resolving feature folders is replaced with a single call to `_currentScene.callInterfaceMethod("FeatureProviderInterface", "getFeaturesFolders")`.

---

## How to implement a new interface

In an external plugin descriptor:
```python
class MyNode(desc.CommandLineNode, desc.interface.FeatureProviderInterface):

    def getFeaturesFolders(self, node) -> list:
        return [node.attribute("output").value]
```
No changes to core Meshroom are needed for the node to be discovered and used by the UI.

### DIFF
--- a/meshroom/core/desc/interface.py
+++ b/meshroom/core/desc/interface.py
@@ -1,0 +1,69 @@
+class Interface:
+    """Marker base class for node descriptor interfaces.
+
+    Inherit from this to declare a new interface. Classes that inherit from
+    an Interface are automatically discovered by InterfaceMeta and listed in
+    the ``interfaces`` attribute of the concrete node descriptor class.
+    """
+
+    @staticmethod
+    def upstreamNodesWithInterface(node, attribute, interface_name):
+        result = []
+        visited = set()
+        stack = [attribute]
+
+        while stack:
+            attr = stack.pop()
+            if not attr.isLink:
+                continue
+            source_node = attr.inputRootLink.node
+            if source_node in visited:
+                continue
+            visited.add(source_node)
+            if interface_name in source_node.nodeDesc.interfaces:
+                result.append(source_node)
+                continue
+            for _, input_attr in source_node.attributes.items():
+                stack.append(input_attr)
+
+        return result
+
+class InterfaceMeta(type):
+    """Metaclass that populates the ``interfaces`` attribute on any class that uses it.
+
+    When a class is created with this metaclass, ``interfaces`` is set to the list
+    of ``Interface`` subclass names found in the class's MRO (excluding the class
+    itself and the base ``Interface`` marker).
+    """
+    def __new__(mcs, name, bases, namespace):
+        cls = super().__new__(mcs, name, bases, namespace)
+        cls.interfaces = [
+            b.__name__ for b in type.mro(cls)
+            if b is not Interface and b is not cls
+            and isinstance(b, type) and issubclass(b, Interface)
+        ]
+        return cls
+    
+class InterfacedClass(metaclass=InterfaceMeta):
+    """Base class for objects that want automatic interface discovery via InterfaceMeta."""
+    pass
+
+
+class FeatureProviderInterface(Interface):
+
+    def getFeaturesFolders(self, node) -> list:
+        raise NotImplementedError
+    
+    def getDescriberTypes(self, node) -> list:
+        raise NotImplementedError
+
+class MatchProviderInterface(Interface):
+
+    def getMatchesFolders(self, node) -> list:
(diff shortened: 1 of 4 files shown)


meshroom/core/desc/interface.py
```diff
@@ -0,0 +1,35 @@
...
+        cls.interfaces = [
+            b.__name__ for b in type.mro(cls)
+            if b is not Interface and b is not cls
+            and isinstance(b, type) and issubclass(b, Interface)
+        ]
+        return cls
+    
+class InterfacedClass(metaclass=InterfaceMeta):
+    """Base class for objects that want automatic interface discovery via InterfaceMeta."""
+    pass
```
nicolas-lambert-tc, line 49: For clarity I would rename it `HasInterfaces`

meshroom/core/desc/interface.py
```diff
@@ -0,0 +1,35 @@
...
+    of ``Interface`` subclass names found in the class's MRO (excluding the class
+    itself and the base ``Interface`` marker).
+    """
+    def __new__(mcs, name, bases, namespace):
+        cls = super().__new__(mcs, name, bases, namespace)
+        cls.interfaces = [
+            b.__name__ for b in type.mro(cls)
+            if b is not Interface and b is not cls
+            and isinstance(b, type) and issubclass(b, Interface)
+        ]
```
nicolas-lambert-tc, line 44: I would rename it `interfaceNames` to prevent user to try code like `if MyInterface in MyNode.interfaces:`

meshroom/ui/scene.py
```diff
@@ -985,11 +992,38 @@ def setBuildingIntrinsics(self, value):
...
+            return None
+        node = entry.node
+        method = getattr(node.nodeDesc, methodName, None)
+        if callable(method):
+            try:
+                return method(node)
+            except Exception as e:
+                logging.error(f"callInterfaceMethod: {interfaceName}.{methodName} raised: {e}")
+                return None
+        return None
```
nicolas-lambert-tc, line 1011: On failure, it could try to load the next possible activeNode that implement this interface, using a while ?

meshroom/ui/scene.py
```diff
@@ -418,9 +412,16 @@ def initActiveNodes(self):
...
+        for nodeType, nodePlugin in meshroom.core.pluginManager.getRegisteredNodePlugins().items():
+            nodeDesc = nodePlugin.nodeDescriptor 
+            for interface in nodeDesc.interfaces:
+                if not self._activeNodes.get(interface):
+                    self._activeNodes.add(ActiveNode(interface, parent=self))
+
         allUiNodes = set(self.uiNodes) | usedNodeTypes | allLoadedNodeTypes
 
-        for nodeType in allUiNodes:
+        for nodeType in allUiNodes:
```
Alxiice, line 424:
```suggestion
        for nodeType in allUiNodes:
```

PR #3151, opened 2026-06-29
Add new `IONode` and `OutputNode`

nicolas-lambert-tc:
Screenshot_2026-06-29_11-26-25

## Description

This PR introduces generic `IONode` and `OutputNode` descriptor concepts to replace the previous `CopyFiles`-specific handling of batch outputs.

Until now, `meshroom_batch -o` was hardcoded around `CopyFiles` nodes. This change makes output export behavior rely on nodes explicitly declaring themselves as output nodes, allowing other node types to expose output behavior without being tied to `CopyFiles`.

`IONode` is introduced as a shared base descriptor for input/output node categories. It centralizes common attribute helper methods and allows UI/core logic to identify IO nodes consistently.

## Features list

- [X] Add a generic `OutputNode` to allow external overrides on exposed attributes.
- [X] `CopyFiles` is now an `OutputNode`

## Usage examples

Set the same output folder on every output node:

```bash
meshroom_batch -p cameraTracking -i /input/images -o /output/results -s /tmp/project.mg
```

Set a specific output folder for one output node:

```bash
meshroom_batch -p customPipeline -i /input/images -o CopyFiles_1=/output/copyfiles -s /tmp/project.mg
```

Set different folders for several output nodes:

```bash
meshroom_batch -p customPipeline -i /input/images \
  -o ExportMesh_1=/output/mesh ExportTextures_1=/output/textures \
  -s /tmp/project.mg
```

Set an exposed non-file output attribute explicitly:

```bash
meshroom_batch -p customPipeline -i /input/images \
  -o MyOutput_1.exportLabel=preview \
  -s /tmp/project.mg
```

When only a non-file output attribute is provided, existing/default output folders are left unchanged:

```bash
meshroom_batch -p customPipeline -i /input/images \
  -o MyOutput_1.exportLabel=final \
  -s /tmp/project.mg
```

Mix a global output folder with a targeted output attribute:

```bash
meshroom_batch -p customPipeline -i /input/images \
  -o /output/default MyOutput_1.exportLabel=final \
  -s /tmp/project.mg
```

Example `OutputNode` declaration:

```python
from typing import ClassVar

class ExportResult(desc.Node, desc.OutputNode):
    outputAttributes: ClassVar[list[str]] = ["folder", "label", "enabled"]

    inputs = [
        desc.File(
            name="folder",
            label="Output Folder",
            description="Destination folder.",
            value="",
        ),
        desc.StringParam(
            name="label",
            label="Export Label",
            description="Label exposed to meshroom_batch.",
            value="default",
        ),
        desc.BoolParam(
            name="enabled",
            label="Enabled",
            description="Whether this export is enabled.",
            value=True,
        ),
    ]

    outputs = []
```

With this declaration:
- `meshroom_batch -o ExportResult_1=/tmp/export` sets all exposed `File` output attributes, here `folder`.
- `meshroom_batch -o ExportResult_1.folder=/tmp/export` targets `folder` explicitly.
- `meshroom_batch -o ExportResult_1.label=preview` sets the exposed string attribute.
- `meshroom_batch -o ExportResult_1.enabled=false` sets the exposed boolean attribute.
- `meshroom_batch -o ExportResult_1.label=final` does not modify `folder`.

## Implementation remarks

- `OutputNode` is implemented as a descriptor mixin, similarly to the existing input/init node concepts. `CopyFiles` now inherits from this mixin and declares `outputAttributes = ["output"]`, preserving existing behavior while making the mechanism extensible.

- `IONode` acts as the shared base class for IO-oriented nodes. It centralizes common attribute helper methods such as resetting, extending, and setting node attributes. `InputNode` keeps its `initialize(...)` contract, while `OutputNode` keeps its stricter export semantics.

- Output nodes expose attributes through the `outputAttributes` class variable. These attributes may be of any descriptor type. Only exposed attributes whose descriptor is `desc.File` are automatically initialized from `meshroom_batch -o` as output folders.

- `meshroom_batch -o` still supports the previous node-level syntax, for example `CopyFiles_1=/output/path`. It also supports attribute-level targeting with `OutputNode.attribute=value`, allowing non-file exposed attributes to be configured explicitly. If no output folder is specified, folder attributes are left unchanged.

- Graph template loading now keeps or removes output nodes based on the new `OutputNode` concept instead of checking for the `CopyFiles` node type directly. The previous `copyOutputs` argument is preserved as a deprecated alias to avoid breaking existing callers.

- The UI header color logic uses the shared `IONode` base class to distinguish IO nodes from regular processing nodes without duplicating checks for each IO subtype.


meshroom/core/graph.py
```diff
@@ -958,6 +967,63 @@ def findInputNodes(self):
...
+            - "path": set the output folder on all output nodes.
+            - "OutputNodeName=path": set the output folder on one output node.
+            - "OutputNodeName.attribute=value": set one exposed output attribute.
+        """
+        outputNodes = self.findOutputNodes()
+        if len(outputNodes) == 0:
+            raise RuntimeError('meshroom_batch requires a pipeline graph with at least ' +
+                               'one output node, none found.')
+
+        reExtract = re.compile(r'(\w+)(?:\.(\w+))?=(.*)')  # NodeName=value or NodeName.attribute=value
```
Alxiice: You can use this: `(\w+)(([:.])(\w[\w.]*))?=(.*)`
To handle the `NodeInstanceName.attribute=value` / `NodeType:attribute=value` that is in use in meshroom_batch 
- `N.A=V`: set the attribute on the node instance
- `N:A=V`: set the attribute on all nodes of the type

meshroom/ui/qml/GraphEditor/Node.qml
```diff
@@ -29,7 +29,15 @@ Item {
     /// Combined x and y
     property point position: Qt.point(x, y)
     /// Styling
-    readonly property color defaultColor: isCompatibilityNode ? activePalette.mid : !node.isComputableType ? "#BA3D69" : activePalette.base
+    readonly property color defaultColor: {
+        if (isCompatibilityNode)
+            return activePalette.mid
+        if (node.isIONode)
+            return Colors.blue
+        if (!node.isComputableType)
```
Alxiice, line 37: Just an idea: maybe replace this with `isInputNode` ?
- input = red
- output = blue
- other (even InitNode that are non-computable) : activePalette.base

meshroom/core/desc/node.py
```diff
@@ -672,3 +665,84 @@ def setAttributes(self, node, attributesDict):
...
+        Args:
+            node (Node): the node whose output folder must be initialized
+            outputFolder (str): the output folder to set
+        """
+        outputFolderAttributes = self.getOutputFolderAttributes(node)
+        if not outputFolderAttributes:
+            raise AttributeError(f"'{node.name}' has no exposed File output attribute.")
+
+        for attr in outputFolderAttributes:
+            self.setAttributes(node, {attr.name: outputFolder})
```
Alxiice, line 788: As the same value is set on all output folder attributes, what is the use of having a list for `outputFolderAttributes` ?
Can you provide an example on how it can be used ?

bin/meshroom_batch
```diff
@@ -231,46 +232,24 @@ with meshroom.core.graph.GraphModification(graph):
...
-                if n.name == node:  # If found, set the value, and remove it from the list
-                    n.output.value = value
-                    copyNodes.pop(i)
-                    if globalCopyPath == '':  # Fallback in case some nodes would have no path
-                        globalCopyPath = value
-                    break
-
-        for n in copyNodes:  # Set the remaining CopyPath nodes with the global path
-            n.output.value = globalCopyPath
+        graph.configureOutputNodes(args.output)
```
Alxiice, line 262: if you moved this inside `graph` then it would logical to do the same for `args.input`

meshroom/core/graph.py
nicolas-lambert-tc: Thanks, I'll fix it

meshroom/ui/qml/GraphEditor/Node.qml
nicolas-lambert-tc: As we talked, I've changed the color of the initNodes in green to prevent users to think the node is in some error state

meshroom/core/desc/node.py
nicolas-lambert-tc: I think it can be useful for nodes that export multiple files type (ex: albedo, normalMap, depthmap) and you want to control each outputs fodlers from the batch

bin/meshroom_batch
nicolas-lambert-tc: I think it's already the case (l 199)
```
    if args.input:
        # get input nodes
        inputNodes = graph.findInputNodes()
        inputNodesNames = [n.getName() for n in inputNodes]
```

Alxiice (approved):

bin/meshroom_batch
```diff
@@ -88,12 +88,13 @@ general_group.add_argument(
 
 general_group.add_argument(
     '-o', '--output', 
-    metavar='FOLDER COPYFILES_INSTANCE=FOLDER',
+    metavar='FOLDER OUTPUT_NODE_INSTANCE=FOLDER OUTPUT_NODE_INSTANCE.ATTRIBUTE=VALUE',
```
Alxiice: Missing OUTPUT_NODE_TYPE:ATTRIBUTE=VALUE

meshroom/core/graph.py
```diff
@@ -958,6 +967,63 @@ def findInputNodes(self):
...
+            - "path": set the output folder on all output nodes.
+            - "OutputNodeName=path": set the output folder on one output node.
+            - "OutputNodeName.attribute=value": set one exposed output attribute.
+        """
+        outputNodes = self.findOutputNodes()
+        if len(outputNodes) == 0:
+            raise RuntimeError('meshroom_batch requires a pipeline graph with at least ' +
+                               'one output node, none found.')
+
+        reExtract = re.compile(r'(\w+)(?:[:.](\w[\w.]*))?=(.*)')  # NodeName=value or NodeName.attribute=value
```
Alxiice: It doesn't work with NodeType:param=value

servantftransperfect:
graph.configureOutputNodes is not used in app.py.

servantftransperfect:
Potential bug if there is a "=" in the file path set

bin/meshroom_batch
nicolas-lambert-tc: done

cbentejac (approved):

cbentejac (approved):

PR #3153, opened 2026-06-30
Explicit templates file management

nicolas-lambert-tc:
<p align="center">
  image
  image

</p>

## Description

This PR makes Meshroom template files explicit in the project flow. Templates can now be detected from their file extension or serialized header, opened through a dedicated template flow, tracked in their own recent-files list, and clearly identified in the GraphEditor when an opened graph comes from a template.

## Features list

- [X] Add the ".mgt" extension to explicitly represent template files
- [X] Old templates files (".mg" with `template:true`) are still loadable
- [X] Display a visible GraphEditor indicator when the current graph was opened from a template.

## Implementation remarks

Template detection is centralized in `meshroom.core.files` so both core and UI code use the same rules. 
Opening a template still initializes an unsaved graph, but the source template filepath is stored separately on the graph model for UI feedback and recent-template tracking. 
The GraphEditor indicator is intentionally lightweight and disappears once the graph becomes a regular saved project.

### DIFF
--- a/meshroom/ui/qml/Application.qml
+++ b/meshroom/ui/qml/Application.qml
@@ -102,7 +102,7 @@ Page {
          * Otherwise, show a warning dialog and returns false.
          * Closing the warning dialog reopens the specified `sourceSaveDialog`, to allow the user to try again.
          */
-        const emptyFilename = Filepath.basename(filepath).trim() === ".mg";
+        const emptyFilename = Filepath.basename(filepath).trim() === sourceSaveDialog.defaultSuffix;
 
         // Provided filename is not valid
         if (emptyFilename) {
@@ -195,8 +195,8 @@ Page {
         signal closed(var result)
 
         title: "Save Template"
-        nameFilters: ["Meshroom Graphs (*.mg)"]
-        defaultSuffix: ".mg"
+        nameFilters: ["Meshroom Templates (*.mgt)"]
+        defaultSuffix: ".mgt"
         fileMode: Platform.FileDialog.SaveFile
         onAccepted: {
             if (!validateFilepathForSave(currentFile, saveTemplateDialog))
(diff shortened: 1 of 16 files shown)


servantftransperfect:
Why don't we use the header to check  .mgt validity ?

meshroom/ui/qml/Application.qml
```diff
@@ -1509,6 +1509,23 @@ Page {
...
+                                const templatePath = coreGraph ? coreGraph.templateFilepath : ""
+                                return templatePath
+                            }
+
+                            text: MaterialIcons.library_books
+                            visible: Boolean(templateFilePath)
+                            font.pointSize: 11
+                            padding: 2
+                            checked: true
+                            ToolTip.text: `Current graph is a template: ${templateFilePath}`
```
cbentejac, line 1526: This is not updated if the currently opened template is saved as the new one.

meshroom/ui/qml/Application.qml
```diff
@@ -830,6 +828,63 @@ Page {
...
+                        shortcut: Shortcut {
+                            sequence: "Ctrl+Shift+T"
+                            context: Qt.ApplicationShortcut
+                            onActivated: saveAsTemplateAction.triggered()
+                        }
+                        onTriggered: {
+                            initFileDialogFolder(saveTemplateDialog)
+                            saveTemplateDialog.open()
+                        }
+                    }
```
cbentejac, line 886: This should add the new template to the list of recently opened templates, and re-trigger the list of accessible templates (in order for the newly added template to be directly accessible from the "New..." menu if it is located in a folder that's looked in for templates.

meshroom/ui/qml/Application.qml
cbentejac: Since templates now have their own "recent" list, the folder containing the last opened template should be saved for the dialog to open in that folder rather than in the most recent project's folder.

cbentejac (changes requested):

cbentejac (approved):
The requested changes regarding the UI part will be addressed in an ulterior PR.

PR #3154, opened 2026-06-30
Add node to fetch parameter values from another Meshroom scene

Alxiice:
## Description

Add nodes to retrieve infos from a different meshroom scene.
- `GetMeshroomSceneParams`: take a list of parameters to fetch, and a scene, and create a JSON file with the param values.
- to use it easily in scenes, `UnwrapMeshroomSceneParam` will unwrap this JSON file and extract a single item so that we can use it in the scene.

image

## Miscancellous

- Little fix on the AttributeItemdelegate : model is unavailable when collapsed. It creates issues when we try to dynamically add items in a list attribute in the update for example.


meshroom/ui/qml/GraphEditor/AttributeItemDelegate.qml
```diff
@@ -779,9 +779,9 @@ RowLayout {
                 }
                 ListView {
                     id: lv
-                    model: listAttributeLayout.expanded ? attribute.value : undefined
-                    visible: model !== undefined && count > 0
-                    implicitHeight: Math.min(contentHeight, 300)
+                    model: attribute.value
```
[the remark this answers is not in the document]
Alxiice: It's a component instanciated from a Loader which means that if the component exists, atribute is filled too. So this should never happen.

meshroom/nodes/general/scripts/extractMeshroomSceneParams.py
```diff
@@ -0,0 +1,107 @@
...
+requestedParams = []
+for item in args.request.split(REQUEST_SEPARATOR):
+    item = item.strip()
+    if not item:
+        continue
+    if NODE_PARAM_SEPARATOR not in item:
+        raise ValueError(
+            f"Invalid request '{item}': expected 'nodeInstance{NODE_PARAM_SEPARATOR}param'."
+        )
+    nodeName, paramPath = item.split(NODE_PARAM_SEPARATOR)
```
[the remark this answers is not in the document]
Alxiice: If we have multiple times the node separator, then the attributes passed to the script are wrong. Therefore it's expected that the script fails. On the other hand if we add this protection and there is a second node separator in the item string, then there will be another issue raised later

meshroom/nodes/general/GenerateMeshroomScene.py
cbentejac:
```suggestion
__version__ = "2.0"
```
The version needs to be updated since there is a change in the list of attributes.

cbentejac:
```suggestion
            label="Parameter Overrides",
```

meshroom/nodes/general/GetMgSceneParams.py
```diff
@@ -0,0 +1,209 @@
...
+
+    def update(self, node):
+        try:
+            self.updateChoices(node)
+        except Exception as e:
+            logging.warning(f"[UnwrapMeshroomSceneParam] Failed to set the choices: {e}")
+        try:
+            self.setOutput(node)
+        except Exception as e:
+            logging.warning(f"[UnwrapMeshroomSceneParam] Failed to set the node output: {e}")
```
cbentejac: Should be moved to its own file.

meshroom/nodes/general/GetMgSceneParams.py
```diff
@@ -0,0 +1,209 @@
...
+    MeshroomSceneParameter node to fill the selectedParameter choices,
+    so it's better to use it directly after this node.
+    """
+
+    category = "Utils"
+
+    inputs = [
+        desc.File(
+            name="jsonFile",
+            label="JSON file",
```
cbentejac:
```suggestion
            label="JSON File",
```

meshroom/nodes/general/GetMgSceneParams.py
```diff
@@ -0,0 +1,209 @@
...
+            description=(
+                "JSON file generated by a MeshroomSceneParameter node.\n"
```
cbentejac: `MeshroomSceneParameter` does not exist.

```suggestion
                "JSON file generated by a GetMeshroomSceneParams node.\n"
```

meshroom/nodes/general/GetMgSceneParams.py
```diff
@@ -0,0 +1,209 @@
...
+        )
+    ]
+
+
+class UnwrapMeshroomSceneParam(desc.InputNode, desc.InitNode):
+    """Unwrap the JSON file created by a MeshroomSceneParameter node
+    to expose its items to the graph.
+    
+    It uses the jsonFile connection to fetch parameters from the
+    MeshroomSceneParameter node to fill the selectedParameter choices,
```
cbentejac:
```suggestion
    GetMeshroomSceneParams node to fill the selectedParameter choices,
```

meshroom/nodes/general/GetMgSceneParams.py
```diff
@@ -0,0 +1,209 @@
...
+    category = "Utils"
+
+    inputs = [
+        desc.File(
+            name="jsonFile",
+            label="JSON file",
+            description=(
+                "JSON file generated by a MeshroomSceneParameter node.\n"
+                "Note that the input must be a MeshroomSceneParameter node "
+                "because we use the input to fetch the choice parameters."
```
cbentejac:
```suggestion
                "JSON file generated by a GetMeshroomSceneParams node.\n"
                "Note that the input must be a GetMeshroomSceneParams node "
                "because we use the input to fetch the choice parameters."
```

meshroom/nodes/general/GetMgSceneParams.py
```diff
@@ -0,0 +1,209 @@
...
+            return
+        with open(jsonFile, "r") as f:
+            data = json.load(f)
+        logging.debug(f"JSON file : {jsonFile}:\n{data}")
+        for item in data:
+            if item.get("node") == nodeInstance and item.get("parameter") == param:
+                node.outputValue.value = item.get("value")
+                break
+        else:
+            node.outputValue.value = ""
```
cbentejac: The issue with these static methods is that if we disconnect the "JSON File" parameter, "Selected Parameter" and "Output Value" will still retain their previous values without ever being udpated.

meshroom/nodes/general/GetMgSceneParams.py
```diff
@@ -0,0 +1,209 @@
...
+            values=[],
+            exclusive=True,
+            exposed=True,
+        )
+    ]
+
+    outputs = [
+        desc.StringParam(
+            name="outputValue",
+            description="Extracted value from the JSON file",
```
cbentejac:
```suggestion
            description="Extracted value from the JSON file.",
```

meshroom/nodes/general/GetMgSceneParams.py
```diff
@@ -0,0 +1,209 @@
...
+                ),
+                desc.BoolParam(
+                    name="failOnMissingParams",
+                    description="Fail if we don't find one or several params inside the scene.",
+                    value=True,
+                    invalidate=False
+                ),
+            ]
+        )
+    ]
```
cbentejac, line 108: Shouldn't this also be added to the "advanced" group?


```suggestion
        desc.GroupAttribute(
            name="advanced",
            items=[
                desc.BoolParam(
                    name="failOnMissingScene",
                    description="Fail if the scene doesn't exist.",
                    value=True,
                    invalidate=False
                ),
                desc.BoolParam(
                    name="failOnMissingParams",
                    description="Fail if we don't find one or several params inside the scene.",
                    value=True,
                    invalidate=False
                ),
            ],
            advanced=True,
        )
    ]
```

cbentejac (changes requested):

meshroom/nodes/general/GetMgSceneParams.py
Alxiice: This is true howvever this is not a huge issue:
- First of all we have the same issue if we remove the JSON file. And meshroom has no way to notify the node that the cache have been deleted, so there will always have kind of this issue
- Also it will be evaluated at runtime anyways so the output value is not really an issue.
- The Selected Parameter is just a convenient way not to have to open the JSON file but the only important part is that we get the correct key when the node is evaluated

meshroom/nodes/general/scripts/extractMeshroomSceneParams.py
```diff
@@ -0,0 +1,153 @@
...
+    )
+    parser.add_argument("--output", required=True, help="Output JSON file.")
+
+    parser.add_argument(
+        "--failOnMissingScene", action="store_true",
+        help="Fail if the scene doesn't exist."
+    )
+    parser.add_argument(
+        "--failOnMissingParams", action="store_true",
+        help="FaiFail if we don't find one or several params inside the scene."
```
cbentejac:
```suggestion
        help="Fail if we don't find one or several parameters inside the scene."
```

meshroom/nodes/general/GenerateMeshroomScene.py
cbentejac, line 133: Not introduced by this PR, but there's a duplication of the "-p {templateScene}" part of the command. 
```suggestion

```

meshroom/nodes/general/UnwrapMeshroomSceneParam.py
```diff
@@ -0,0 +1,100 @@
...
+__version__ = "1.0"
+
+import json
+import logging
+from pathlib import Path
+from meshroom.core import desc
+
+
+class UnwrapMeshroomSceneParam(desc.CommandLineNode):
+    """Unwrap the JSON file created by a MeshroomSceneParameter node
```
cbentejac:
```suggestion
    """Unwrap the JSON file created by a GetMeshroomSceneParams node
```

cbentejac (changes requested):
Minor change requests, with a highlight on the one in GenerateMeshroomScene.py. 

Otherwise it has been tested and looks good to me .

cbentejac (approved):

PR #3156, opened 2026-07-02
UI: Fix the QML engine instant coding feature

Alxiice:
## Description

Fix the QML engine instant coding feature

Peek 2026-07-03 11-09


Alxiice:
Comment from https://github.com/alicevision/Meshroom/pull/3159
```python
self._debounceTimer = QTimer(parent=self, singleShot=True, interval=100)
```

PR #3158, opened 2026-07-03
UI: Add buttons to unlock the node

Alxiice:
## Description

Add 2 buttons on the Graph Editor node context menu:
- "Unlock Node": unlock current node, only available if the node is locked
- "Unlock Selected Nodes": unlock selected nodes

When the node is locked you can simply unlock this node:
image

Also you can unlock selected nodes:
image

### DIFF
--- a/meshroom/ui/qml/GraphEditor/GraphEditor.qml
+++ b/meshroom/ui/qml/GraphEditor/GraphEditor.qml
@@ -781,6 +781,17 @@ Item {
                         onTriggered: uigraph.restartJobErrorTasks(nodeMenu.currentNode)
                     }
                     MenuItem {
+                        text: "Unlock Node"
+                        enabled: nodeMenu.currentNode.locked
+                        height: visible ? implicitHeight : 0
+                        onTriggered: uigraph.graph.forceUnlockNodes([nodeMenu.currentNode])
+                    }
+                    MenuItem {
+                        text: "Unlock Selected Nodes"
+                        height: visible ? implicitHeight : 0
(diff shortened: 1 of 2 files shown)


meshroom/ui/qml/GraphEditor/GraphEditor.qml
```diff
@@ -780,6 +780,17 @@ Item {
...
+                        text: "Unlock Node"
+                        enabled: nodeMenu.currentNode.locked
+                        height: visible ? implicitHeight : 0
+                        onTriggered: uigraph.graph.forceUnlockNodes([nodeMenu.currentNode])
+                    }
+                    MenuItem {
+                        text: "Unlock Selected Nodes"
+                        height: visible ? implicitHeight : 0
+                        onTriggered: uigraph.graph.forceUnlockNodes(_currentScene.getSelectedNodes())
+                    }
```
cbentejac, line 793: This is a bit counter-intuitive: to enable the "Unlock Node", the selected node NEEDS to be locked (which makes sense). However, even if a single node is selected, the "Unlock Selected Nodes" option will always be enabled because there is no check on whether the selection contains more than one node and whether it contains at least a locked one.

Image

cbentejac, line 793: I would probably also hide (and not just disable) the "Unlock Selected Nodes" option if there's only one selected node.

cbentejac (changes requested):
Minor changes requested regarding the two new options in the Graph Editor. Otherwise, looks good to me.

PR #3159, opened 2026-07-03
[UI] Add an auto-scroll feature to connect edges on huge scenes

Alxiice:
## Description

When scenes are too big and we want to connect nodes that cannot be seen at the same time, it's not convenient to create edges.

This PR aims to add an "auto-scroll" feature that will move the graph when we approach the edges of the graph editor. This way users can easily drag edges, move in the graph and connect the edges.

## Notes for the review

Aside from the code review I would like to know what would be the best settings for
- the border margins (threshold where the scroll is enabled)
- the scroll speed

## Demo

auto scroll

### DIFF
--- a/meshroom/ui/qml/GraphEditor/GraphEditor.qml
+++ b/meshroom/ui/qml/GraphEditor/GraphEditor.qml
@@ -23,6 +23,7 @@ Item {
     // Autoscroll properties
     readonly property int autoscrollMargin: 100  // Border thickness (width in pixels)
     readonly property int autoscrollSpeed: 15    // Speed factor
+    readonly property real maxScrollSpeed: 3.0   // Clamp scroll speed
     property bool isDraggingEdge: false          // Tracks whether we are dragging an AttributePin
     property point dragMousePos: Qt.point(0, 0)  // Position of the mouse during edge dragging
 
@@ -179,6 +180,10 @@ Item {
 
         function onEdgeDraggingChanged(dragging) {
             root.isDraggingEdge = dragging
+            root.highlightedBorderRight=false
+            root.highlightedBorderLeft=false
+            root.highlightedBorderTop=false
+            root.highlightedBorderBottom=false
         }
 
         function onEdgeDragMousePosChanged(windowX, windowY) {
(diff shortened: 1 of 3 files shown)


meshroom/ui/utils.py
```diff
@@ -9,49 +9,46 @@
...
-                return
-            # Restore root item geometry
-            if self._rootItem:
-                root.setGeometry(self._rootItem.geometry())
-                self._rootItem.deleteLater()
-            self._rootItem = root
-
-        self.objectCreated.connect(onObjectCreated)
+        # Add a single shot timer to launch the reload after all events are processed
+        self._debounceTimer = QTimer(singleShot=True, interval=100)
```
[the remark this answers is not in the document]
Alxiice: Issue to fix in https://github.com/alicevision/Meshroom/pull/3156

meshroom/ui/qml/GraphEditor/GraphEditor.qml
```diff
@@ -156,6 +173,102 @@ Item {
...
+    Connections {
+        target: _currentScene 
+        ignoreUnknownSignals: true
+
+        function onEdgeDraggingChanged(dragging) {
+            root.isDraggingEdge = dragging
+            root.highlightedBorderRight=false
+            root.highlightedBorderLeft=false
+            root.highlightedBorderTop=false
+            root.highlightedBorderBottom=false
```
raphaelKoskas, line 186: Since those attributes are always reset together, maybe regroup all those booleans inside a single object ?

meshroom/ui/qml/GraphEditor/GraphEditor.qml
```diff
@@ -156,6 +173,102 @@ Item {
...
+                deltaX = -root.autoscrollSpeed * factorX
+            }
+            // Top
+            if (mouseY < root.autoscrollMargin) {
+                var factorY = Math.min(root.maxScrollSpeed, (root.autoscrollMargin - mouseY) / root.autoscrollMargin)
+                deltaY = root.autoscrollSpeed * factorY
+            }
+            // Bottom
+            else if (mouseY > root.height - root.autoscrollMargin) {
+                var factorY = Math.min(root.maxScrollSpeed, (mouseY - (root.height - root.autoscrollMargin)) / root.autoscrollMargin)
```
raphaelKoskas, line 226: The computations for factorX and factorY are almost identical in each case. Maybe this could be refactored within a function ?

raphaelKoskas (commented):
- Function-wise, this code does what it supposed to do and I'm good with the scrolling speed and margins you defined.
- Keeping the name of the attribute getting pulled visible is a good idea, but brings out an odd behaviour in the context of this fix :
    - the label square follows the view translation within the graph editor, but is clipped to the meshroom window, meaning specifically in the context this fix was made for, I  can easily get the label square to leave the graph editor panel,
    - and I believe a more interesting feature, especially for this context, would be to have the label square to track the mouse position (with a small offset for readability) instead.
labelSquareLeavingGraphEditorPanel

Aside from this, it looks good to me.

Issue #3166, opened 2026-07-13
Create a new node subtype to allow execution directly in Meshroom environment


Alxiice:
https://github.com/alicevision/Meshroom/blob/507e2dc7d36892c8d44b464a79faaa486be8d9a1/meshroom/core/node.py#L705-L707

Because of this when we launch graph execution, any node based on `desc.Node` will execute in a subprocess.

On several occasions this isn't necessary, for example for nodes that will do a simple file lookup and don't require a specific environment. There is a workaround which is using `desc.CommandLineNode` with the `processChunk` method:

```python
import os
from meshroom.core import desc

class GetFileExt(desc.CommandLineNode):
    inputs = [
        desc.File(
            name="path",
            value="",
            exposed=True,
        ),
    ]

    outputs = [
        desc.StringParam(
            name="extension",
            value=None,
        ),
    ]

    def processChunk(self, chunk):
        chunk.node.extension.value = os.path.splitext(chunk.node.path.value)[1]
```

**Comparison:**
| mode                   | time  |
|------------------------|-------|
| `desc.Node`            | 4.8s  |
| `desc.CommandLineNode` | 0.02s |

PR #3167, opened 2026-07-13
Improve performance on path manipulation nodes

Alxiice:
## Description

- Add a node `GetFileExtension`
- Replace `desc.Node` by `desc.CommandLineNode` on
  - `GetParentFolder`
  - `FlattenFiles`
  - `PathJoin`

Using CommandLineNode bypass the `_processInIsolatedEnvironment` call (see https://github.com/alicevision/Meshroom/issues/3166) that launches a subprocess.
Bypassing it saves several seconds: the node compute takes almost 0 sec.

### DIFF
--- a/meshroom/core/desc/__init__.py
+++ b/meshroom/core/desc/__init__.py
@@ -47,6 +47,7 @@ from .node import (
     InternalAttributesFactory,
     MrNodeType,
     Node,
+    InlineNode,
     NodeVersionType,
     NodeVersionTypeEnum,
 )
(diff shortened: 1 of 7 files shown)


fabiencastan (changes requested):
Do not use commandlinenode for that. This is hack. Add a way to disable subprocess in Node.

Alxiice:
@fabiencastan I added a `InlineNode` type (we can discuss about the name). I found that was a simpler way to avoid computing inside the current env (rather than modifying the whole graph compute / process system).

PR #3177, opened 2026-08-06
[GraphEditor] Rewrite AttributeItemDelegate with outside definitions for the components

raphaelKoskas:
## Description

This PR moves the definition of the components inside AttributeItemDelegate to AttributeControls, as externalised components, thus easing their use in a broader context than the GraphEditor.

## Features list

- [X]  Rewrote the components in separated files

## Implementation remarks
- ListAttribute and GroupAttribute have been left as is due to their high dependency on the attribute object.


meshroom/ui/qml/GraphEditor/AttributeControls/ColorHue.qml
```diff
@@ -0,0 +1,53 @@
...
+        text: displayValue
+        selectByMouse: true
+        validator: DoubleValidator {
+            locale: 'C'  // Use '.' decimal separator disregarding the system locale
+        }
+        onEditingFinished: setTextFieldAttribute(root.attribute, text)
+        onAccepted: setTextFieldAttribute(root.attribute, text)
+        Component.onDestruction: {
+            if (activeFocus)
+                setTextFieldAttribute(root.attribute, text)
```
cbentejac: Not sure this works, I think `setTextFieldAttribute` is defined in AttributeItemDelegate and is definitely connected to attributes.

meshroom/ui/qml/GraphEditor/AttributeControls/ColorHue.qml
```diff
@@ -0,0 +1,53 @@
...
+        readonly property real formattedValue: value.toFixed(stepDecimalCount)
+        enabled: root.editable
+        value: attribute.value
+        from: 0
+        to: 1
+        stepSize: 0.01
+        snapMode: Slider.SnapAlways
+        onPressedChanged: {
+            if (!pressed)
+                _currentScene.setAttribute(attribute, formattedValue)
```
cbentejac: This should be defined in AttributeItemDelegate when the component is declared.

meshroom/ui/qml/GraphEditor/AttributeItemDelegate.qml
```diff
@@ -478,25 +478,13 @@ RowLayout {
...
-                verticalAlignment: Text.AlignVCenter
-                padding: 4
-                background: Rectangle {
-                    anchors.fill: parent
-                    border.width: 0
-                    radius: 20
-                    color: Qt.darker(palette.window, 1.1)
-                }
-            }
+            AttributeControls.NotComputed{}
```
cbentejac:
```suggestion
            AttributeControls.NotComputed {}
```
Generally speaking, it's preferred to have a whitespace between the braces and the identifier.

meshroom/ui/qml/GraphEditor/AttributeControls/PushButton.qml
cbentejac, line 1: Might be overkill for this one, as it absolutely does not do anything within the component, and there is no customisation of the standard `Button` at all.

meshroom/ui/qml/GraphEditor/AttributeItemDelegate.qml
```diff
@@ -811,333 +585,74 @@ RowLayout {
...
-                        }
-                        else
-                        {
-                            _currentScene.setAttribute(attribute, !attribute.value)
-                        }
+            AttributeControls.CheckBoxRow{
+                editable: root.editable
+                keyable: root.attribute.keyable
+                keyedValue: attribute.keyValues.getValueAtKeyOrDefault(_currentScene.selectedViewId)
+                plainValue: attribute.value
```
cbentejac:
```
An error occurred executing the property metacall 1 on property "keyValues" of Attribute(0x55b103ec37b0)
  File "/repo/meshroom/core/attribute.py", line 299, in _getKeyValues
    raise RuntimeError(f"Cannot get keyValues of {self._getFullName()}, the attribute is not keyable.")
RuntimeError: Cannot get keyValues of ImageMasking_1.invert, the attribute is not keyable.
/repo/meshroom/ui/qml/GraphEditor/AttributeItemDelegate.qml:599: Error: Cannot get keyValues of ImageMasking_1.invert, the attribute is not keyable.
```
This occurs for all the `BoolParam` in any node, and it is enough to select the node in the Graph Editor to have it appear (no further interaction is required). Here it's for the `ImageMasking` node but a `CameraInit` will do too (with more errors as there are more boolean attributes).

meshroom/ui/qml/GraphEditor/AttributeControls/Color.qml
```diff
@@ -0,0 +1,66 @@
+import QtQuick
+import QtQuick.Controls
+import QtQuick.Layouts
+import QtQuick.Dialogs
+
+RowLayout {
+    id: root
+    required property var attribute
```
cbentejac: This is not making the component attribute-agnostic. You explicitly expect an attribute (of the type "Attribute", since you later access "attribute.value"). 

What is great though is that you did remove all the calls to the attribute-specific function, and emit the proper signal to adapt the behaviour depending on where the component will be instantiated.

```suggestion
    required property var value
```

meshroom/ui/qml/GraphEditor/AttributeControls/Color.qml
```diff
@@ -0,0 +1,66 @@
...
+    property bool editable
+    property string previousColor: ""
+    signal clicked(var checked, var previousColor, var colorTextValue)
+    signal editingFinished(var text)
+    signal accepted(var text)
+    signal destruction(bool activeFocus, var text)
+    CheckBox {
+        id: colorCheckbox
+        Layout.alignment: Qt.AlignLeft
+        checked: attribute.value === "" ? false : true
```
cbentejac:
```suggestion
        checked: value === "" ? false : true
```

meshroom/ui/qml/GraphEditor/AttributeControls/Color.qml
```diff
@@ -0,0 +1,66 @@
...
+
+        }
+    }
+    TextField {
+        id: colorText
+        Layout.alignment: Qt.AlignLeft
+        implicitWidth: 100
+        enabled: colorCheckbox.checked && root.editable
+        visible: colorCheckbox.checked
+        text: colorCheckbox.checked ? attribute.value : ""
```
cbentejac:
```suggestion
        text: colorCheckbox.checked ? value : ""
```

meshroom/ui/qml/GraphEditor/AttributeControls/Color.qml
```diff
@@ -0,0 +1,66 @@
...
+            onClicked: colorDialog.open()
+        }
+    }
+    ColorDialog {
+        id: colorDialog
+        title: "Please choose a color"
+        selectedColor: colorText.text
+        onAccepted: {
+            colorText.text = colorDialog.selectedColor
+            // Artificially trigger change of attribute value
```
cbentejac:
```suggestion
            // Artificially trigger change of value
```

meshroom/ui/qml/GraphEditor/AttributeControls/ColorHue.qml
```diff
@@ -0,0 +1,51 @@
+import QtQuick
+import QtQuick.Controls
+import QtQuick.Layouts
+
+RowLayout {
+    id: root
+    required property var attribute
```
cbentejac: Similarly to the comments I made on the "Color" component:

```suggestion
    required property var value
```

meshroom/ui/qml/GraphEditor/AttributeControls/ColorHue.qml
```diff
@@ -0,0 +1,51 @@
...
+    property bool editable
+    signal editingFinished(var text)
+    signal accepted(var text)
+    signal destruction(bool activeFocus, var text)
+    signal pressedChanged(bool pressed, var formattedValue)
+    TextField {
+        implicitWidth: 100
+        enabled: root.editable
+        // Cast value to string to avoid intrusive scientific notations on numbers
+        property string displayValue: String(slider.pressed ? slider.formattedValue : attribute.value)
```
cbentejac:
```suggestion
        property string displayValue: String(slider.pressed ? slider.formattedValue : root.value)
```

meshroom/ui/qml/GraphEditor/AttributeControls/ColorHue.qml
```diff
@@ -0,0 +1,51 @@
...
+            locale: 'C'  // Use '.' decimal separator disregarding the system locale
+        }
+        onEditingFinished: root.editingFinished(text)
+        onAccepted: root.accepted(text)        
+        Component.onDestruction: root.destruction(activeFocus, text)
+    }
+    Rectangle {
+        height: slider.height
+        width: height
+        color: Qt.hsla(slider.pressed ? slider.formattedValue : attribute.value, 1, 0.5, 1)
```
cbentejac:
```suggestion
        color: Qt.hsla(slider.pressed ? slider.formattedValue : root.value, 1, 0.5, 1)
```

meshroom/ui/qml/GraphEditor/AttributeControls/ColorHue.qml
```diff
@@ -0,0 +1,51 @@
...
+    }
+    Slider {
+        id: slider
+        Layout.fillWidth: true
+        readonly property int stepDecimalCount: 2
+        readonly property real formattedValue: value.toFixed(stepDecimalCount)
+        enabled: root.editable
+        value: root.attribute.value
```
cbentejac:
```suggestion
        value: root.value
```

meshroom/ui/qml/GraphEditor/AttributeControls/SliderField.qml
```diff
@@ -0,0 +1,84 @@
...
+	property bool editable
+    ExpressionTextField {
+        id: expressionTextField
+        implicitWidth: 100
+        Layout.fillWidth: !slider.active
+        enabled: root.editable
+        // Cast value to string to avoid intrusive scientific notations on numbers
+        property string displayValue: String(slider.active && slider.item.pressed ? slider.item.formattedValue :
+                                            attribute.keyable ? attribute.keyValues.getValueAtKeyOrDefault(_currentScene.selectedViewId) :
+                                            attribute.value)
```
cbentejac: If the notion of keyable exists directly in the component definition, then again, it is not attribute-agnostic.

`attribute` should not be used as such, and the properties of the Attribute object should definitely not be called from within the component (otherwise it means that we cannot use this slider for anything else).

meshroom/ui/qml/GraphEditor/AttributeControls/SliderField.qml
```diff
@@ -0,0 +1,84 @@
...
+        text: displayValue
+        selectByMouse: true
+        // Note: Use autoScroll as a workaround for alignment
+        // When the value change keep the text align to the left to be able to read the most important part
+        // of the number. When we are editing (item is in focus), the content should follow the editing.
+        autoScroll: activeFocus
+        isInt: attribute.type === "FloatParam" ? false : true
+        onEditingFinished: {
+            if (!hasExprError) {
+                setTextFieldAttribute(root.attribute, expressionTextField.evaluatedValue)
```
cbentejac: Same here: we refer directly to the type of attribute and we even call directly a function whose goal is to set the attribute's value through the undo stack. 

In AttributeItemDelegate, you use `setTextFieldAttribute` correctly for the ColorHue component (the call is external to the component, it is made in the context of the AttributeItemDelegate only, and it is what we expect).

meshroom/ui/qml/GraphEditor/AttributeItemDelegate.qml
```diff
@@ -413,7 +413,7 @@ RowLayout {
         }
     }
 
-    function setTextFieldAttribute(value) {
+    function setTextFieldAttribute(attr, value) {
```
cbentejac: I don't think it is necessary to change the prototype of the function. Since no call to it should be made outside of this file, the current attribute can always be accessed through `root.attribute` as it was before.

This also simplifies the calls that are made to it in slots (for example, if we look again at the ColorHue component, there's no need to call `root.attribute` from the slot rather than having it internalized here).

cbentejac (changes requested):
Components are not, for the most part, fully attribute-agnostic: they either depend on an "attribute" property which is expected to be "shaped" like an Attribute object (which we want to avoid) or they call directly a method that makes attribute-related operations only.

Some components like ColorHue are perfectly instantiated in AttributeItemDelegate, with the slots correctly defined upon instantiation (rather than directly within the component) and this is what we should be doing with the other components as well. (there's just a minor hiccup regarding the direct reference to the "attribute" property, we should only declare and forward the value directly)

The ideal would be to have no mention whatsoever to attributes within the components, as is done in the `Choice.qml` and `ChoiceMulti.qml` files.

However the split between components is exactly what we expect (except maybe for the `PushButton.qml` which does really do anything beside instantiating a standard button) and this heavily improves the readability of `AttributeItemDelegate.qml`. Most of the hard work has already been done, so kudos on that part!

meshroom/ui/qml/GraphEditor/AttributeControls/CheckBoxRow.qml
```diff
@@ -0,0 +1,16 @@
+import QtQuick
+import QtQuick.Controls
+
+Row {
+    id: root
+    property bool editable
+    property var keyedValue
```
cbentejac:
```suggestion
    property var checked
```
It does not make sense to refer to the keyed attributes here, it is a specificity of the implementation of some attributes.

meshroom/ui/qml/GraphEditor/AttributeControls/CheckBoxRow.qml
```diff
@@ -0,0 +1,16 @@
...
+    signal toggled()
+    CheckBox {
+        enabled: root.editable
+        checked: root.keyedValue
```
cbentejac:
```suggestion
        checked: root.checked
```

cbentejac (changes requested):

cbentejac (approved):

PR #3188, opened 2026-09-18
[nodes] Add new `StringJoin` node to concatenate strings

cbentejac:
## Description

This PR adds a utility node that allows to concatenate strings with a set separator. Whitespaces and new lines can be managed.

Whitespaces management will always take precedence over new lines management.


meshroom/nodes/general/StringJoin.py
```diff
@@ -0,0 +1,95 @@
...
+            value="",
+        ),
+        desc.GroupAttribute(
+            name="spacesManagement",
+            description="Option to manage spaces and new lines in input strings.",
+            items=[
+                desc.ChoiceParam(
+                    name="newLinesManagement",
+                    description="The new line characters will be processed before the whitespaces. "
+                                "If new lines are replaced with whitespaces, these whitespaces will be replaced with the whitespace-related options."
```
CedricThebault:
```suggestion
                                "If new lines are replaced with whitespaces, these whitespaces will be replaced with the whitespace-related options.\n"
```

demoulinv (approved):

